PR #2, opened 2025-08-14
Implement OAuth login flow

nathanjcochran:
Updates the `tiger auth login` command to support logging in with OAuth:

- If any of the `--public-key`, `--secret-key`, or `--project-id` flags (or corresponding environment variables) are provided, the command still works like it previously did - i.e. it uses the provided values, and prompts for any that are missing.
- If none of those flags/env vars are given, the command initiates an OAuth flow (specifically, using the Authorization Code Grant):
   1. First, it opens the user's browser to a console page that prompts them to authorize the CLI app (if they are not already logged in, they must first log into the console).
   2. Upon clicking "Authorize", they are redirected back to a localhost callback endpoint running as part of the CLI, with an authorization code in the `?code=` query string parameter
   3. The CLI then exchanges the authorization code for an OAuth access token and redirect the user to a success page that tells them to return to their terminal.
   4. The CLI uses the access token to fetch a list of projects that the user belongs to. If they only belong to one, it uses that as the default project. If they belong to multiple, they are prompted to choose one at the terminal.
   5. The CLI fetches the user's profile information (name and email)
   6. The CLI then uses the access token and project ID to create a new set of client credentials. The credentials are given a name that includes the user's name/email.
   7. The credentials and project ID are stored like usual.
   
Note that there are still some things I'd like to improve as fast-follows (but I'd like to get this merged soon because the conflicts are killing me 😅). For example, some of the logic involves making calls to the GraphQL endpoints directly. To fix that, we will need to add a few new endpoints to the public API that accept OAuth access tokens instead of the API Key.
   
### Depends on:

- https://github.com/timescale/savannah-gateway/pull/1312
- https://github.com/timescale/web-cloud/pull/5920

### How to test

1. Make sure https://github.com/timescale/savannah-gateway/pull/1312 is running in the dev environment.
2. Check out and run https://github.com/timescale/web-cloud/pull/5920 locally.
3. Set the new `console_url` and `gateway_url` config values appropriately:
   ```
   console_url: https://local.dev.timescale.com:8080
   gateway_url: https://console.dev.timescale.com/api
   ```
4. Run `tiger auth login`.

### DIFF
--- a/internal/tiger/cmd/auth_test.go
+++ b/internal/tiger/cmd/auth_test.go
@@ -301,10 +301,8 @@ func startMockOAuthServer(t *testing.T, projects []Project) *httptest.Server {
 
 		// Handle different GraphQL queries
 		if strings.Contains(query, "getAllProjects") {
-			response := GetAllProjectsResponse{
-				Data: struct {
-					GetAllProjects []Project `json:"getAllProjects"`
-				}{
+			response := GraphQLResponse[GetAllProjectsData]{
+				Data: &GetAllProjectsData{
 					GetAllProjects: projects,
 				},
 			}
@@ -313,10 +311,8 @@ func startMockOAuthServer(t *testing.T, projects []Project) *httptest.Server {
 				http.Error(w, "Failed to encode response", http.StatusInternalServerError)
 			}
 		} else if strings.Contains(query, "getUser") {
-			response := GetUserResponse{
-				Data: struct {
-					GetUser User `json:"getUser"`
-				}{
+			response := GraphQLResponse[GetUserData]{
+				Data: &GetUserData{
 					GetUser: User{
 						ID:    "user-456",
 						Name:  "Test User",
@@ -329,10 +325,8 @@ func startMockOAuthServer(t *testing.T, projects []Project) *httptest.Server {
 				http.Error(w, "Failed to encode response", http.StatusInternalServerError)
 			}
 		} else if strings.Contains(query, "createPATRecord") {
-			response := CreatePATRecordResponse{
-				Data: struct {
-					CreatePATRecord PATRecordResponse `json:"createPATRecord"`
-				}{
+			response := GraphQLResponse[CreatePATRecordData]{
+				Data: &CreatePATRecordData{
 					CreatePATRecord: PATRecordResponse{
 						ClientCredentials: struct {
 							AccessKey string `json:"accessKey"`
--- a/internal/tiger/cmd/graphql.go
+++ b/internal/tiger/cmd/graphql.go
@@ -16,11 +16,20 @@ type GraphQLClient struct {
 	URL string
 }
 
-// GetAllProjectsResponse represents the response from the getAllProjects query
-type GetAllProjectsResponse struct {
-	Data struct {
-		GetAllProjects []Project `json:"getAllProjects"`
-	} `json:"data"`
+// GraphQLResponse represents a generic GraphQL response wrapper
+type GraphQLResponse[T any] struct {
+	Data   *T             `json:"data"`
+	Errors []GraphQLError `json:"errors,omitempty"`
+}
+
+// GraphQLError represents an error returned in a GraphQL response
+type GraphQLError struct {
+	Message string `json:"message"`
+}
+
+// GetAllProjectsData represents the data from the getAllProjects query
+type GetAllProjectsData struct {
+	GetAllProjects []Project `json:"getAllProjects"`
 }
 
 // Project represents a project from the GraphQL API
@@ -40,19 +49,17 @@ func (c *GraphQLClient) getUserProjects(accessToken string) ([]Project, error) {
 		}
 	`
 
-	response, err := makeGraphQLRequest[GetAllProjectsResponse](c.URL, accessToken, query, nil)
+	response, err := makeGraphQLRequest[GetAllProjectsData](c.URL, accessToken, query, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	return response.Data.GetAllProjects, nil
+	return response.GetAllProjects, nil
 }
 
-// GetUserResponse represents the response from the getUser query
-type GetUserResponse struct {
-	Data struct {
-		GetUser User `json:"getUser"`
-	} `json:"data"`
+// GetUserData represents the data from the getUser query
+type GetUserData struct {
+	GetUser User `json:"getUser"`
 }
 
 // User represents a user from the GraphQL API
@@ -74,19 +81,17 @@ func (c *GraphQLClient) getUser(accessToken string) (*User, error) {
 		}
 	`
 
-	response, err := makeGraphQLRequest[GetUserResponse](c.URL, accessToken, query, nil)
+	response, err := makeGraphQLRequest[GetUserData](c.URL, accessToken, query, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	return &response.Data.GetUser, nil
+	return &response.GetUser, nil
 }
 
-// CreatePATRecordResponse represents the response from the createPATRecord mutation
-type CreatePATRecordResponse struct {
-	Data struct {
-		CreatePATRecord PATRecordResponse `json:"createPATRecord"`
-	} `json:"data"`
+// CreatePATRecordData represents the data from the createPATRecord mutation
+type CreatePATRecordData struct {
+	CreatePATRecord PATRecordResponse `json:"createPATRecord"`
 }
 
 // PATRecordResponse represents the response from creating a PAT record
@@ -117,12 +122,12 @@ func (c *GraphQLClient) createPATRecord(accessToken, projectID, patName string) 
 		},
 	}
 
-	response, err := makeGraphQLRequest[CreatePATRecordResponse](c.URL, accessToken, query, variables)
+	response, err := makeGraphQLRequest[CreatePATRecordData](c.URL, accessToken, query, variables)
 	if err != nil {
 		return nil, err
 	}
 
-	return &response.Data.CreatePATRecord, nil
+	return &response.CreatePATRecord, nil
 }
 
 // makeGraphQLRequest makes a GraphQL request to the API
@@ -162,10 +167,23 @@ func makeGraphQLRequest[T any](queryURL, accessToken, query string, variables ma
 		return nil, fmt.Errorf("failed to read GraphQL response: %w", err)
 	}
 
-	var response T
+	var response GraphQLResponse[T]
 	if err := json.Unmarshal(body, &response); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal GraphQL response: %w", err)
 	}
 
-	return &response, nil
+	// Check for GraphQL errors
+	if len(response.Errors) > 0 {
+		var errorMessages []string
+		for _, gqlErr := range response.Errors {
+			errorMessages = append(errorMessages, gqlErr.Message)
+		}
+		return nil, fmt.Errorf("GraphQL errors: %s", strings.Join(errorMessages, "; "))
+	}
+
+	if response.Data == nil {
+		return nil, fmt.Errorf("GraphQL response contains no data")
+	}
+
+	return response.Data, nil
 }


internal/tiger/cmd/auth.go
```diff
@@ -50,72 +72,96 @@ func buildLoginCmd() *cobra.Command {
...
+				}
 
-			cmd.SilenceUsage = true
+				// Combine the keys in the format "public:secret" for storage
+				apiKey = fmt.Sprintf("%s:%s", publicKey, secretKey)
+			}
 
 			// Validate the API key by making a test API call
-			fmt.Fprintln(cmd.OutOrStdout(), "Validating API key...")
+			fmt.Fprintln(c.out, "Validating API key...")
```
cevian: is there a reason for this change? cmd.OutOrStdout() still seems correct

cevian: also below

internal/tiger/cmd/auth.go
```diff
@@ -50,72 +72,96 @@ func buildLoginCmd() *cobra.Command {
...
 			}
 
-			// If any credentials are missing, prompt for them all at once
-			if publicKey == "" || secretKey == "" || projectID == "" {
-				// Set SilenceUsage before attempting interactive prompts
-				// This prevents showing usage help for TTY/environmental errors
-				cmd.SilenceUsage = true
-
-				var err error
+			apiKey := fmt.Sprintf("%s:%s", publicKey, secretKey)
```
cevian: nit: seems a bit weird to set here and on line 160

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
 	return cmd
 }
 
+type loginCmd struct {
```
cevian: nit: I don't think this is a cobra cmd so I think a better name is `OathLogin` or something

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+	fmt.Fprintln(c.out, "Opening browser for authentication...")
+	if err := openBrowser(authURL); err != nil {
+		fmt.Fprintf(c.out, "Failed to open browser: %s\nPlease manually navigate to: %s\n", err, authURL)
+	}
+
+	// Wait for callback with timeout
+	select {
+	case result := <-server.resultChan:
+		return result.accessToken, result.err
+	case <-time.After(2 * time.Minute):
```
cevian: 2 min seems short if the need to login (find password etc)? maybe 5-10?

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+	patRecord, err := c.createPATRecord(accessToken, projectID, c.buildPATName(user))
+	if err != nil {
+		return "", fmt.Errorf("failed to create PAT record: %w", err)
+	}
+
+	// Combine access key and secret key with colon
+	return fmt.Sprintf("%s:%s", patRecord.ClientCredentials.AccessKey, patRecord.ClientCredentials.SecretKey), nil
+}
+
+// Use user name, fallback to email if name is empty, fallback to hostname as last resort
```
cevian: can we add a comment that this is the name displayed in the dashboard?

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+	// Wait for callback with timeout
+	select {
+	case result := <-server.resultChan:
+		return result.accessToken, result.err
+	case <-time.After(2 * time.Minute):
+		return "", fmt.Errorf("authentication timeout - no callback received within 2 minutes")
+	}
+}
+
+func generateCodeVerifier() (string, error) {
```
cevian: can we use https://pkg.go.dev/golang.org/x/oauth2#GenerateVerifier here? I'd prefer using the official oath2 code whenever possible?

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+	data := make([]byte, 32)
+	if _, err := rand.Read(data); err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.WithPadding(base64.NoPadding).EncodeToString(data), nil
+}
+
+func generateCodeChallenge(verifier string) string {
```
cevian: we can use https://pkg.go.dev/golang.org/x/oauth2#S256ChallengeFromVerifier 

or the option.

internal/tiger/cmd/auth.go
cevian: nit: I'd prefer this in a separate go file. Its a lot of code with different dependencies.

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+	response, err := makeGraphQLRequest[CreatePATRecordResponse](c.cfg.GatewayURL, accessToken, query, variables)
+	if err != nil {
+		return nil, err
+	}
+
+	return &response.Data.CreatePATRecord, nil
+}
+
+// makeGraphQLRequest makes a GraphQL request to the API using generics
+func makeGraphQLRequest[T any](gatewayURL, accessToken, query string, variables map[string]any) (*T, error) {
```
cevian: nit would prefer all the graphql stuff isolated into a separate file with a note that we are reusing the graphql endpoints because they are already available.

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+
+	for i, project := range m.projects {
+		cursor := " "
+		if m.cursor == i {
+			cursor = ">"
+		}
+		s += fmt.Sprintf("%s %s (%s)\n", cursor, project.Name, project.ID)
+	}
+
+	s += "\nUse ↑/↓ arrows to navigate, Enter to select, q to quit"
```
cevian: can we also allow selecting by number? (probably easier for ai bots)

specs/oauth.md
```diff
@@ -0,0 +1,198 @@
...
+   - Fallback to encrypted file storage if keychain unavailable
+   - CLI shuts down local HTTP server
+
+#### Error Handling:
+- Network errors: Retry with exponential backoff
+- User cancellation: Clean exit with helpful message
+- Invalid credentials: Clear error message with retry option
+- PKCE validation failures: Security error, force re-authentication
+
+### 2. Client Credentials Flow (CI/CD)
```
cevian: is this relevant? seems we are just doing the regular pat token for cicd. If its not relevant please delete the section

specs/oauth.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Token Storage
+
+#### Secure Keychain (Preferred)
+- **macOS**: Keychain Services API
+- **Windows**: Windows Credential Manager
+- **Linux**: Secret Service (libsecret/gnome-keyring)
+
+Storage keys:
+- `tiger-cli-access-token`: Current access token
```
cevian: we don't store the access or refresh tokens right?

specs/oauth.md
```diff
@@ -0,0 +1,198 @@
...
+file:
+- Location: `~/.config/tiger/credentials.json`
+- Permissions: 0600 (owner read/write only)
+
+### Token Usage
+
+All API requests include bearer token in Authorization header: ```
+Authorization: Bearer <access_token> ```
+
+### Token Refresh Flow
```
cevian: we are using a PAT so no refresh token right?

specs/oauth.md
```diff
@@ -0,0 +1,198 @@
+# Tiger CLI OAuth Integration Specification
```
cevian, line 1: I think we should document design decisions near the top of this doc. 2 major ones I see:

1) we exchange auth/refresh tokens for a PAT. We do this so you can easily revoke the token through the dashboard. We only store the PAT
2. We reuse the graphql endpoints that already exist. For DRY.

In general I think this need to be updated.

internal/tiger/cmd/auth.go
```diff
@@ -36,11 +50,19 @@ func getServiceName() string {
 	return serviceName
 }
 
+// OAuth parameters
+// TODO: Currently unused, but should probably support
+const clientID = "65183398-ece9-40c4-84e7-84974c51255b"
```
cevian: if this is unused, lets remove from this PR

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+		return nil
+	}
+
+	return errors.Join(
+		doClose(s.server),
+		doClose(s.listener),
+	)
+}
+
+func (c *loginCmd) startOAuthServer(state, codeVerifier string) (*oauthServer, error) {
```
cevian: nit: can't this be a constructor for oathServer?

internal/tiger/cmd/auth.go
```diff
@@ -205,6 +242,537 @@ func buildAuthCmd() *cobra.Command {
...
+
+	return &oauthServer{
+		server:      server,
+		listener:    listener,
+		redirectURI: redirectUri,
+		resultChan:  resultChan,
+	}, nil
+}
+
+func (c *loginCmd) createOAuthCallbackHandler(expectedState, codeVerifier string, resultChan chan<- oauthResult) http.HandlerFunc {
```
cevian: nit: why is this not a method of oauthServer?

internal/tiger/cmd/auth.go
nathanjcochran: I just found `fmt.Fprintln(cmd.OutOrStdout(), ...)` to be an unnecessarily verbose way to print something to the console. And since we're calling `cmd.OutOrStdout()` many times throughout the file, I thought I'd consolidate/condense it a bit. But yeah, doesn't matter much either way (both are correct).

internal/tiger/cmd/auth.go
nathanjcochran: Yeah, this happened when rebasing onto your most recent changes. I wanted to improve it (along with a bunch of other things in here) but was also just trying to get this PR merged before there were more conflicts.

internal/tiger/cmd/auth.go
nathanjcochran: Increased to 5 minutes here: https://github.com/timescale/tiger-cli/pull/2/commits/ff1d60d8bf9b88b02c1387708da3d734560a5687

internal/tiger/cmd/auth.go
nathanjcochran: Done here: https://github.com/timescale/tiger-cli/pull/2/commits/aee80669ce455cd7a798d6b00effb86b123300be

internal/tiger/cmd/auth.go
nathanjcochran: Ahh good call! I didn't even know that package existed. That's what I get for asking Claude to write this for me 😅. Interesting that the [implementation of the official function](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.30.0:pkce.go;l=20-39) is identical to the one Claude spit out for me, but Claude decided to copy it rather than just use the oauth2 package directly 🤦‍♂️.

I updated everything I could to use that package here: https://github.com/timescale/tiger-cli/pull/2/commits/a03af0943cec5ca024599d2c8bf42d8c12befd44. It allowed me to delete some other custom code too, which is nice 💯.

internal/tiger/cmd/auth.go
nathanjcochran: See comment above. Done here: https://github.com/timescale/tiger-cli/commit/a03af0943cec5ca024599d2c8bf42d8c12befd44

specs/oauth.md
nathanjcochran: Sorry, this spec is totally outdated. It was a starting point, but the actual code ended up diverging from it quite a lot. I'm inclined to just delete this spec entirely, unless you feel strongly about keeping it around (in which case I need to more-or-less rewrite it from scratch).

internal/tiger/cmd/auth.go
nathanjcochran: It is currently not being verified by the back-end, but it is being sent. And I think that ideally, we would be verifying it in some capacity (I am planning to look into that when I look into PKCE). The Client ID is a fairly important part of the OAuth spec, and while we don't technically need it right now, I think it would be good to adhere to the spec as much as possible. In particular, if we ever want to re-use the authorization pages I added to `web-cloud` for the sake of authorizing different applications, we would need to support the Client ID as a way to distinguish between them. Better to send it along now than encounter difficulties down the road, imo.

internal/tiger/cmd/auth.go
nathanjcochran: Sorry, what exactly do you mean? Go doesn't have "constructors" per se like other languages do. This method essentially _is_ the constructor for `oauthServer`.

internal/tiger/cmd/auth.go
nathanjcochran: Because it accesses a field on `loginCmd`, and `oauthServer` doesn't exist yet at the time this function is called. This function is called as part of the logic that creates the `oauthServer`.

internal/tiger/cmd/auth.go
nathanjcochran: I was going to encapsulate the entire login command (e.g. the `RunE` function) within this type, but that got messed up in the last rebase. I agree it's not great though. I'll look into restructuring/renaming it.

internal/tiger/cmd/auth.go
nathanjcochran: Moved into another file here: https://github.com/timescale/tiger-cli/pull/2/commits/a3d18fef4b8a0dfedbb96a65c2e576a2ecc2ee89

internal/tiger/cmd/auth.go
nathanjcochran: Reverted as part of this refactor: https://github.com/timescale/tiger-cli/pull/2/commits/94782a1e501153f234c776fe2112da50c8be7cd0

internal/tiger/cmd/auth.go
nathanjcochran: I improved this as part of a larger refactor/reorganization of the code: https://github.com/timescale/tiger-cli/pull/2/commits/94782a1e501153f234c776fe2112da50c8be7cd0

internal/tiger/cmd/auth.go
nathanjcochran: Renamed the struct as part of a larger reorganization/refactor of the code: https://github.com/timescale/tiger-cli/pull/2/commits/94782a1e501153f234c776fe2112da50c8be7cd0

nathanjcochran: Moved all of the OAuth logic into a separate file: https://github.com/timescale/tiger-cli/pull/2/commits/94782a1e501153f234c776fe2112da50c8be7cd0

internal/tiger/cmd/auth.go
nathanjcochran: I refactored this a bit here: https://github.com/timescale/tiger-cli/pull/2/commits/94782a1e501153f234c776fe2112da50c8be7cd0. There is now a separate `oauthCallback` struct that implements `http.Handler`.

internal/tiger/cmd/auth.go
nathanjcochran: Not sure that ai bots will be going through this exact flow (given that it involves interacting with the browser), but done here: https://github.com/timescale/tiger-cli/pull/2/commits/399753c45d4140ce27bc59cae3f3bab19d2a0acf

specs/oauth.md
nathanjcochran: Updated the spec to match reality: https://github.com/timescale/tiger-cli/pull/2/commits/2c678e05220e90a9f715fe69f80f9356c90eba13

specs/oauth.md
nathanjcochran: Updated the spec to match reality: https://github.com/timescale/tiger-cli/pull/2/commits/2c678e05220e90a9f715fe69f80f9356c90eba13

specs/oauth.md
nathanjcochran: Updated the spec to match reality: https://github.com/timescale/tiger-cli/pull/2/commits/2c678e05220e90a9f715fe69f80f9356c90eba13

specs/oauth.md
nathanjcochran: Done here, along with a couple other design decisions I thought were worth documenting explicitly: https://github.com/timescale/tiger-cli/pull/2/commits/2c678e05220e90a9f715fe69f80f9356c90eba13

specs/oauth.md
```diff
@@ -0,0 +1,152 @@
...
+
+### OAuth flow uses new `web-cloud` authorization page
+
+- Not the built-in FusionAuth login page
+- Ensures user does not have to log in again if they're already logged into the console
+- Page styling and functionality are fully in our control
+- No need to deal with creating different FusionAuth applications, etc.
+  (FusionAuth is not really well-suited to this use case).
+
+### Client secret is not verified
```
cevian, line 20: could we verify the same is true in similar CLI projects?

PR #7, opened 2025-09-10
Add service fork command

Askir:
This PR implements the `tiger service fork` command to create forks of existing services.

This PR now relies on the new fork api with different forking strategies: NOW, LATEST_SNAPSHOT, and PITR
there are 5 open PRs for that:

- https://github.com/timescale/savannah-public/pull/41
- https://github.com/timescale/timescale-graphql-client/pull/4
- https://github.com/timescale/savannah-gateway/pull/1351
- https://github.com/timescale/savannah-projects/pull/1283
- https://github.com/timescale/savannah-common/pull/2471

The openapi spec from the REST api PR is the on that this one is developed against. However this code has not yet been tested against that (since the other PRs have not yet been merged).

*Examples*
```bash
# Fork a service at the current db state (either via new snapshot or wal replay)
tiger service fork svc-12345 --now

# Fork a service at last snapshot
tiger service fork svc-12345 --last-snapshot

# For a service at a specific point in time
tiger service fork svc-12345 --to-timestamp 2025-01-15T10:30:00Z

# Fork a service with custom name
tiger service fork svc-12345 --now --name my-forked-db

# Fork a service with custom ressources
tiger service fork svc-12345 --now --cpu 2000 --memory 8

# Fork without waiting for completion
tiger service fork svc-12345 --now --no-wait

# No default behavior - user must specify one of the three timing options
tiger service fork svc-12345
Error: Must specify --now, --last-snapshot or --to-timestamp
```

### DIFF
--- a/internal/tiger/api/types.go
+++ b/internal/tiger/api/types.go
@@ -93,7 +93,7 @@ type ForkServiceCreate struct {
 	Free *bool `json:"free,omitempty"`
 
 	// MemoryGbs The initial memory allocation in gigabytes. If not provided, will inherit from parent service.
-	MemoryGbs interface{} `json:"memory_gbs,omitempty"`
+	MemoryGbs *int `json:"memory_gbs,omitempty"`
 
 	// Name A human-readable name for the forked service. If not provided, will use parent service name with "-fork" suffix.
 	Name *string `json:"name,omitempty"`
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -975,7 +975,7 @@ components:
           description: The initial CPU allocation in milli-cores. If not provided, will inherit from parent service.
           example: 1000
         memory_gbs:
-          format: integer
+          type: integer
           description: The initial memory allocation in gigabytes. If not provided, will inherit from parent service.
           example: 4
         fork_strategy:


internal/tiger/cmd/service.go
```diff
@@ -1055,6 +1056,215 @@ func formatAllowedCPUValues(configs []CPUMemoryConfig) string {
...
+				return fmt.Errorf("invalid request parameters")
+			default:
+				return fmt.Errorf("API request failed with status %d", forkResp.StatusCode())
+			}
+		},
+	}
+
+	// Add flags
+	cmd.Flags().StringVar(&forkServiceName, "name", "", "Name for the forked service (auto-generated if not provided)")
+	cmd.Flags().BoolVar(&forkNoWait, "no-wait", false, "Don't wait for fork operation to complete")
```
nathanjcochran: Should we support any of [these other flags](https://github.com/timescale/tiger-cli/blob/dedc4cb51952817079358608d672115cfdb4fbdc/internal/tiger/cmd/service.go#L422-L430) from the create service command?

In particular, I think we always want to have `--wait-timeout` as an option when `--no-wait` is an option. And we probably also want to support `--cpu` and `--memory`, so you can make a fork with different resources. And I'm not sure if you can create a fork in a different region, but if so, we should probably support `--region` too.

Also, do we want the behavior where the new fork is set as your default service for future Tiger CLI commands (which is how the create service command [currently works](https://github.com/timescale/tiger-cli/blob/9e5cb755d3eee2073a5452710a9259ea91c61865/internal/tiger/cmd/service.go#L390-L396))? If so, we would also want to add the `--no-set-default` flag.

But we can do that stuff in follow-up PRs too, if you'd rather just get this one merged.

internal/tiger/cmd/service.go
```diff
@@ -1055,6 +1056,215 @@ func formatAllowedCPUValues(configs []CPUMemoryConfig) string {
...
+
+				// Handle wait behavior
+				if forkNoWait {
+					fmt.Fprintf(cmd.OutOrStdout(), "⏳ Service is being forked. Use 'tiger service list' to check status.\n")
+					return nil
+				}
+
+				// Wait for service to be ready
+				fmt.Fprintf(cmd.OutOrStdout(), "⏳ Waiting for fork to complete (timeout: 30m)...\n")
+				if err := waitForServiceReady(client, projectID, forkedServiceID, 30*time.Minute, cmd); err != nil {
```
nathanjcochran: I definitely think we should support the `--wait-timeout` flag, rather than hard-coding it to 30 minutes here.

openapi.yaml
```diff
@@ -395,7 +395,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/ForkInput'
+              $ref: '#/components/schemas/ServiceCreate'
```
nathanjcochran: This entire `openapi.yaml` file seems to be copied from `savannah-public` (see [here](https://github.com/timescale/savannah-public/blob/7fe03d064acccf650f7919ba832725fcefa84e0f/openapi.yaml#L384-L407)). It would probably be better if we were somehow sharing it instead of duplicating it, but given that's where we're currently at, we should probably update that file with these same changes, to keep them in sync. Can you open a PR against that repo? 🙏

internal/tiger/cmd/service.go
Askir: I have added support for most of the flag mentioned there now. I don't think `--region` will work. Well at least for tiger storage it would cause issues since we have the storage deployed separately for each cluster afaik.

openapi.yaml
Askir: I created a PR to update the REST api with the new fork options and created a new openapi spec, that one will then be synced: https://github.com/timescale/savannah-public/pull/41

internal/tiger/api/types.go
```diff
@@ -70,10 +77,29 @@ type Error struct {
...
+	// - LAST_SNAPSHOT: Use existing snapshot for fast fork
+	// - NOW: Create new snapshot for up-to-date fork
+	// - PITR: Point-in-time recovery using target_time
+	ForkStrategy ForkStrategy `json:"fork_strategy"`
+
+	// Free Whether to create a free forked service (if true, cpu_millis and memory_gbs must not be provided)
+	Free *bool `json:"free,omitempty"`
+
+	// MemoryGbs The initial memory allocation in gigabytes. If not provided, will inherit from parent service.
+	MemoryGbs interface{} `json:"memory_gbs,omitempty"`
```
nathanjcochran: Why is this being generated as an `interface{}` type? I'd expect it to be a `*int`, like `CpuMillis`.

EDIT: It's a problem in the `openapi.yaml` spec. See my other comment below.

openapi.yaml
```diff
@@ -950,6 +961,38 @@ components:
...
+        name:
+          type: string
+          description: A human-readable name for the forked service. If not provided, will use parent service name with "-fork" suffix.
+          example: "my-production-db-fork"
+        cpu_millis:
+          type: integer
+          description: The initial CPU allocation in milli-cores. If not provided, will inherit from parent service.
+          example: 1000
+        memory_gbs:
+          format: integer
```
nathanjcochran: I think this should be `type: integer` instead (see my other comment about the generated type being `interface{}`). Also left a comment [here](https://github.com/timescale/savannah-public/pull/41#discussion_r2394789873) about it.

internal/tiger/cmd/service.go
```diff
@@ -984,3 +985,276 @@ func waitForServiceDeletion(client *api.ClientWithResponses, projectID string, s
...
+  tiger service fork svc-12345 --last-snapshot
+
+  # Fork a service at a specific point in time
+  tiger service fork svc-12345 --to-timestamp 2025-01-15T10:30:00Z
+
+  # Fork with custom name
+  tiger service fork svc-12345 --now --name my-forked-db
+
+  # Fork with custom resources
+  tiger service fork svc-12345 --now --cpu 2000 --memory 8
```
nathanjcochran, line 1032: Might be good to include a `--free` example too

internal/tiger/cmd/service.go
```diff
@@ -984,3 +985,276 @@ func waitForServiceDeletion(client *api.ClientWithResponses, projectID string, s
...
+
+  # Fork without setting as default service
+  tiger service fork svc-12345 --now --no-set-default
+
+  # Fork without waiting for completion
+  tiger service fork svc-12345 --now --no-wait
+
+  # Fork with custom wait timeout
+  tiger service fork svc-12345 --now --wait-timeout 45m`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
nathanjcochran, line 1042: I think we should use the `Args` field here to add validation on the number of arguments that can be provided (it only accepts a single optional service ID argument, right?)

```suggestion
        Args:  cobra.MaximumNArgs(1),
		RunE: func(cmd *cobra.Command, args []string) error {
```

internal/tiger/cmd/service.go
```diff
@@ -984,3 +985,276 @@ func waitForServiceDeletion(client *api.ClientWithResponses, projectID string, s
...
+	// Add flags
+	cmd.Flags().StringVar(&forkServiceName, "name", "", "Name for the forked service (defaults to '{source-name}-fork')")
+	cmd.Flags().BoolVar(&forkNoWait, "no-wait", false, "Don't wait for fork operation to complete")
+	cmd.Flags().BoolVar(&forkNoSetDefault, "no-set-default", false, "Don't set this service as the default service")
+	cmd.Flags().DurationVar(&forkWaitTimeout, "wait-timeout", 30*time.Minute, "Wait timeout duration (e.g., 30m, 1h30m, 90s)")
+
+	// Timing strategy flags
+	cmd.Flags().BoolVar(&forkNow, "now", false, "Fork at the current database state (creates new snapshot or uses WAL replay)")
+	cmd.Flags().BoolVar(&forkLastSnapshot, "last-snapshot", false, "Fork at the last existing snapshot (faster)")
+	cmd.Flags().StringVar(&forkToTimestamp, "to-timestamp", "", "Fork at a specific point in time (RFC3339 format, e.g., 2025-01-15T10:30:00Z)")
```
nathanjcochran, line 1253: I think you could use `cmd.Flags().TimeVar` here instead, which would allow you to push more of the validating/parsing into the `cobra` library. See the [TimeVar](https://pkg.go.dev/github.com/spf13/pflag#TimeVar) docs.

internal/tiger/cmd/service.go
```diff
@@ -984,3 +985,276 @@ func waitForServiceDeletion(client *api.ClientWithResponses, projectID string, s
...
+
+	// Resource customization flags
+	cmd.Flags().IntVar(&forkCPU, "cpu", 0, "CPU allocation in millicores (inherits from source if not specified)")
+	cmd.Flags().IntVar(&forkMemory, "memory", 0, "Memory allocation in gigabytes (inherits from source if not specified)")
```
nathanjcochran, line 1257: Do we also want to add a `-free` flag to it's possible to fork a paid service to a free service (with `--free` or a free service to a paid service (with `--free=false`)? Currently, I don't think this CLI command exposes that ability, even though the API endpoint does. Not sure where we landed on the requirements for that for launch.

PR #9, opened 2025-09-10
Release Pipeline

nathanjcochran:
Adds [GoReleaser](https://goreleaser.com/customization/) configuration for building, packaging, and releasing Tiger CLI, as well as a GitHub Actions workflow for triggering GoReleaser on version tag pushes.

Builds are created for the following:
- Operating Systems:
   - Linux
   - Mac
   - Windows
- Architectures:
   - i386
   - amd64
   - arm64

Additionally, Linux packages are created for:
- Ubuntu/Debian systems (`.deb`)
- Redhat/Fedora/CentOS (`.rpm`)
- Alpine Linux (`.apk`)

Releases are created in GitHub, and all artifacts are also pushed to a new [tiger-cli-releases](https://us-east-1.console.aws.amazon.com/s3/buckets/tiger-cli-releases) S3 bucket in our production AWS account, which has public `GetObject` access (note that I created this bucket manually because I wasn't seeing similar buckets in our Terraform config, but I'm happy to do it via Terraform if someone can point me to a good example).

One of the files pushed to the S3 bucket is an `install.sh` script, which can downloaded via `curl` and piped directly into `sh` to download and install the latest Tiger CLI:

```bash
curl -fsSL https://tiger-cli-releases.s3.amazonaws.com/install/install.sh | sh
```

The pipeline also pushes a commit to our https://github.com/timescale/homebrew-tap repository to update the tiger-cli cask, which makes the package available via homebrew:

```bash
brew install timescale/tap/tiger-cli
```

There are several other installation methods as well (such as direct download of an archive or Linux package). See the README.md for more info.

Note that while we could have relied on GitHub alone for downloading the `install.sh` script and release artifacts, I figured uploading to S3 might provide a more reliable source, and would make it easy to put everything behind a custom URL in the future if we want to.  It also made it possible to test this while the GitHub repo is still private. That being said, it would be very easy to update everything to use GitHub URLs directly instead of S3 URLs after we make the repo public, if people feel that's a better approach.

Finally, note that there are several remaining TODOs that plan to tackle as follow-ups to this PR:
1. Signing and notarizing packages. This mostly just requires tracking down the correct keys to use and wiring it up to GoReleaser.
2. Verifying that the installation works for all operating systems and architectures. This is time-consuming, and I'd rather make small fixes in later PRs, rather than hold this PR up.
3. Potentially implementing some kind of automatic self-update behavior in the CLI, so that users don't get stuck with an old version without realizing a newer version is available (not an immediate priority).

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -123,19 +123,21 @@ fetch_with_retry() {
     local description="${2:-content}"
     local max_retries=3
     local retry_count=0
+    local backoff_seconds=1
 
-    while [ ${retry_count} -lt "${max_retries}" ]; do
+    while [ ${retry_count} -le "${max_retries}" ]; do
         local content
         if content=$(curl -fsSL "${url}" 2>/dev/null); then
             echo "${content}"
             return 0
         else
             retry_count=$((retry_count + 1))
-            if [ "${retry_count}" -lt "${max_retries}" ]; then
+            if [ "${retry_count}" -le "${max_retries}" ]; then
                 log_warn "${description} fetch failed, retrying (${retry_count}/${max_retries})..."
-                sleep 1
+                sleep ${backoff_seconds}
+                backoff_seconds=$((backoff_seconds * 2))
             else
-                log_error "Failed to fetch ${description} after ${max_retries} attempts"
+                log_error "Failed to fetch ${description} after $((max_retries + 1)) attempts"
                 log_error "URL: ${url}"
                 exit 1
             fi
@@ -150,20 +152,22 @@ download_with_retry() {
     local description="${3:-file}"
     local max_retries=3
     local retry_count=0
+    local backoff_seconds=1
 
     log_info "Downloading ${description}..."
     log_info "URL: ${url}"
 
-    while [ ${retry_count} -lt "${max_retries}" ]; do
+    while [ ${retry_count} -le "${max_retries}" ]; do
         if curl -fsSL "${url}" -o "${output_file}"; then
             return 0
         else
             retry_count=$((retry_count + 1))
-            if [ "${retry_count}" -lt "${max_retries}" ]; then
+            if [ "${retry_count}" -le "${max_retries}" ]; then
                 log_warn "${description} download failed, retrying (${retry_count}/${max_retries})..."
-                sleep 1
+                sleep ${backoff_seconds}
+                backoff_seconds=$((backoff_seconds * 2))
             else
-                log_error "Failed to download ${description} after ${max_retries} attempts"
+                log_error "Failed to download ${description} after $((max_retries + 1)) attempts"
                 log_error "URL: ${url}"
                 exit 1
             fi


scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+    while [ ${retry_count} -lt "${max_retries}" ]; do
+        local content
+        if content=$(curl -fsSL "${url}" 2>/dev/null); then
+            echo "${content}"
+            return 0
+        else
+            retry_count=$((retry_count + 1))
+            if [ "${retry_count}" -lt "${max_retries}" ]; then
+                log_warn "${description} fetch failed, retrying (${retry_count}/${max_retries})..."
+                sleep 1
```
cevian: nit: probably better as exponential in # of retries

scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+    log_info "URL: ${url}"
+
+    while [ ${retry_count} -lt "${max_retries}" ]; do
+        if curl -fsSL "${url}" -o "${output_file}"; then
+            return 0
+        else
+            retry_count=$((retry_count + 1))
+            if [ "${retry_count}" -lt "${max_retries}" ]; then
+                log_warn "${description} download failed, retrying (${retry_count}/${max_retries})..."
+                sleep 1
```
cevian: same here re: exponential

scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+ensure_install_dir() {
+    local install_dir
+    install_dir="$(detect_install_dir)"
+
+    log_info "Selected install directory: ${install_dir}"
+
+    # Create install directory if it doesn't exist
+    if [ ! -d "${install_dir}" ]; then
+        if [ "${install_dir}" = "/usr/local/bin" ]; then
+            sudo mkdir -p "${install_dir}"
```
cevian: I don't like going sudo in this script for any reason. If /usr/local/bin doesn't exists lets just exit with an error

scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+            exit 1
+        fi
+    elif command -v shasum >/dev/null 2>&1; then
+        if ! echo "${formatted_checksum}" | shasum -a 256 -c >/dev/null 2>&1; then
+            log_error "Checksum validation failed using shasum"
+            log_error "For security reasons, installation has been aborted"
+            exit 1
+        fi
+    else
+        log_warn "No SHA256 utility available (tried sha256sum, shasum)"
```
cevian: lets make this an error. Everybody should have shasum for goodness sake

scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+    local binary_path="$1"
+    local install_dir="$2"
+
+    log_info "Installing to ${install_dir}..."
+
+    # Copy binary (using sudo if needed)
+    if [ -w "${install_dir}" ]; then
+        cp "${binary_path}" "${install_dir}/${BINARY_NAME}"
+    else
+        sudo cp "${binary_path}" "${install_dir}/${BINARY_NAME}"
```
cevian: should we really sudo here? I am very sceptical

scripts/install.sh
```diff
@@ -0,0 +1,456 @@
...
+    binary_path="$(extract_archive "${archive_name}" "${tmp_dir}" "${platform}")"
+
+    # Copy binary to install directory
+    copy_binary_to_install_dir "${binary_path}" "${install_dir}"
+
+    # Verify installation
+    verify_installation "${install_dir}"
+
+    # Show usage information
+    log_success "Get started with:"
```
cevian, line 452: I think the get started with should have `auth login` as the first step. Then "get help with etc." But the next step should clearly be auth login.

scripts/install.sh
nathanjcochran: This was present because you can't put something in `/usr/local/bin` without sudo.

I decided to just remove `/usr/local/bin` as a installation directory candidate (it was already the last in the list, and probably wouldn't be used 99% of the time). That allowed me to get rid of the `sudo` calls. I also improved the logic for detecting the installation directory, so it now fails early if it can't find or create a writable directory (which should be extremely unlikely, unless the user has done some really weird things with their `$HOME` directory permissions): https://github.com/timescale/tiger-cli/pull/9/commits/3f9d3c6c1cffb282110eeee534ad85a1a70456db

scripts/install.sh
nathanjcochran: See comment here: https://github.com/timescale/tiger-cli/pull/9#discussion_r2353014709

scripts/install.sh
nathanjcochran: Done here: https://github.com/timescale/tiger-cli/pull/9/commits/b26f1944eabff6be72aaba7759a0c47108d63622

scripts/install.sh
nathanjcochran: Done here: https://github.com/timescale/tiger-cli/pull/9/commits/ee9357e792545fe521cf5144af9746fdc80ec59d

scripts/install.sh
nathanjcochran: Done here: https://github.com/timescale/tiger-cli/pull/9/commits/8a7f67430fcda0c08f99255b9ad17bc4fc7dfa69

scripts/install.sh
nathanjcochran: Done here: https://github.com/timescale/tiger-cli/pull/9/commits/8a7f67430fcda0c08f99255b9ad17bc4fc7dfa69

PR #12, opened 2025-09-15
MCP Server

nathanjcochran:
This PR adds an initial implementation of the MCP server. It is intended to be a basis that we can continue to build on and refine (there is definitely still more to improve). I am hoping to get it merged in sooner rather than later, to avoid merge conflicts and make it possible for other people to start building on top of this foundation (since it's a pretty large PR).

Includes the following changes:
- Adds a new `tiger mcp start` command for starting the mcp server. Includes `http` and `stdio` subcommands (stdio is the default).
- Adds an MCP server with the [v0 service tools](https://github.com/timescale/tiger-cli/blob/ce12b442ed1a2af2a266375d174d8c20d41db544/specs/spec_mcp.md?plain=1#L9-L18) from the spec (minus `tiger_service_delete`, since there has been some debate about whether we want to include that, due to its potentially destructive nature).
- Moves around and refactors a bunch of code so it can be shared between the CLI and MCP server. This is probably the source of the majority of the changes in this PR, but I didn't see a good way around it besides duplicating all of that code. Apologies!

Note that we intend to add easy-to-use `tiger mcp install` commands in a future iteration.

Note that there are some changes from the original spec (which I also updated to reflect the changes):
- The commands to start the MCP server are `tiger mcp start [stdio|http]`, rather than `tiger mcp [stdio|http]`. See [here](https://iobeam.slack.com/archives/C099S47CSRX/p1758144743940239) for my reasoning.
- I made more of the `tiger_service_create` tool call parameters optional, with the same defaults that are used by the corresponding CLI commands. When they were required, the LLM was often just choosing random values anyways, when it had no reason to make a more specific choice. Better to provide opinionated defaults that the LLM can override if it has good reason to, imo.
- I removed the `vpc_id` parameter from the `tiger_service_create` tool call, as it isn't yet implemented in the `tiger service create` CLI command or the `savannah-public` API endpoint for creating services.
- I combined the `cpu` and `memory` parameters to the `tiger_service_create` tool call into a single `cpu_memory` parameter that accepts an enum of the valid CPU/memory combinations. See [here](https://iobeam.slack.com/archives/C099S47CSRX/p1758297671934699?thread_ts=1758296649.921299&cid=C099S47CSRX) for my reasoning.
- I made the `tiger_service_create` and `tiger_service_update_password` tool calls store the password in the same way the corresponding CLI commands do (Note, however, that I did not yet include the logic for making a new service the default service - we can add that in a follow-up if we want it).

### How to test

To install the MCP server into Claude Code using the stdio transport:

```bash
claude mcp add tiger-mcp -- tiger mcp start
```

Alternatively, to use the HTTP transport:

```bash
claude mcp add --transport http tiger-mcp -- http://localhost:8080
tiger mcp start http
```

Then start up `claude` and very that the MCP server is connected with the `/mcp` command. If it's connected, try telling Claude you need a new database 😄

### DIFF
--- a/specs/spec_mcp.md
+++ b/specs/spec_mcp.md
@@ -6,6 +6,12 @@ The Tiger MCP (Model Context Protocol) Server provides programmatic access to Ti
 
 The MCP server is written in Go and launched via the Tiger CLI, sharing the same authentication, configuration, and API client.
 
+## Design Decisions
+
+### Dynamic Configuration Loading
+
+Each MCP tool call dynamically creates the API client and loads configuration at execution time, rather than initializing these once when the MCP server starts. This design ensures that configuration changes (API keys, project IDs, etc.) take effect immediately for subsequent tool calls without requiring users to restart the MCP server or reconnect their AI assistant. Users can run `tiger auth login` to update authentication or `tiger config set` to modify other configuration values and see changes reflected instantly in their AI interactions.
+
 ## v0 Tool Priority
 
 For the initial v0 release, implement these essential tools first:


internal/tiger/cmd/auth.go
```diff
@@ -222,103 +201,6 @@ func flagOrEnvVar(flagVal, envVarName string) string {
...
-	apiKeyFile := fmt.Sprintf("%s/api-key", configDir)
-
-	err := os.Remove(apiKeyFile)
-	if err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("failed to remove API key file: %w", err)
-	}
-
-	return nil
-}
-
```
nathanjcochran, line 321: Moved these functions to the `config` package so they can be used by the new `mcp` package. Also refactored it a bit to expose methods for interacting with the keychain, so the unit tests can rely on those methods instead of interacting with the keychain directly.

internal/tiger/cmd/auth_test.go
```diff
@@ -437,13 +439,24 @@ func TestAuthLogin_OAuth_SingleProject(t *testing.T) {
...
+		`Validating API key\.\.\.\n`+
+		`Successfully logged in and stored API key\n`+
+		`Set default project ID to: project-123\n$`, regexp.QuoteMeta(mockServerURL))
+
+	matched, err := regexp.MatchString(expectedPattern, output)
+	if err != nil {
+		t.Fatalf("Regex compilation failed: %v", err)
+	}
+	if !matched {
+		t.Errorf("Output doesn't match expected pattern.\nPattern: %s\nActual output: '%s'", expectedPattern, output)
```
nathanjcochran, line 455: Fixes a failing test caused by #6. Same thing for the code below.

internal/tiger/cmd/db_test.go
```diff
@@ -927,15 +928,6 @@ func TestBuildConnectionString(t *testing.T) {
...
 }
 
-// Helper functions for creating pointers
-func stringPtr(s string) *string {
-	return &s
-}
-
-func intPtr(i int) *int {
-	return &i
-}
```
nathanjcochran, line 937: Created a generic `Ptr` function in the new `util` package that replaces these type-specific versions.

internal/tiger/cmd/service.go
```diff
@@ -709,14 +699,6 @@ func sanitizeServicesForOutput(services []api.Service) []map[string]interface{}
 	return sanitized
 }
 
-// derefString safely dereferences a string pointer, returning empty string if nil
-func derefString(s *string) string {
-	if s == nil {
-		return ""
-	}
-	return *s
-}
```
nathanjcochran, line 718: Create a new generic `Deref` function in the `util` package which replaces this type-specific version.

internal/tiger/cmd/service.go
```diff
@@ -725,22 +707,6 @@ func formatTimePtr(t *time.Time) string {
...
-	return string(*status)
-}
-
-// formatServiceType formats a ServiceType pointer, returning empty string if nil
-func formatServiceType(serviceType *api.ServiceType) string {
-	if serviceType == nil {
-		return ""
-	}
-	return string(*serviceType)
-}
```
nathanjcochran, line 742: These functions were really just dereferencing a value and casting it to a string. I create a new generic `derefStr` function in the `util` package that replaces these type-specific versions.

internal/tiger/cmd/service.go
```diff
@@ -783,10 +749,24 @@ func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID s
...
+
+	// Output the message with appropriate emoji
+	if result.Success {
+		fmt.Fprintf(cmd.OutOrStdout(), "🔐 %s\n", result.Message)
+	} else {
+		if result.Method == "none" {
+			fmt.Fprintf(cmd.OutOrStdout(), "💡 %s\n", result.Message)
+		} else {
+			fmt.Fprintf(cmd.OutOrStdout(), "⚠️  %s\n", result.Message)
+		}
```
nathanjcochran, line 769: Moved the output logic out of `SavePasswordWithMessages` and renamed it to `SavePasswordWithResult` (in the `util` package) so it can be shared with the `mcp` package. The logic for printing the output is now here in the `cmd` package.

internal/tiger/cmd/service.go
```diff
@@ -806,81 +786,6 @@ func setDefaultService(serviceID string, cmd *cobra.Command) error {
...
-		cpuCores := float64(config.CPUMillis) / 1000
-		if cpuCores == float64(int(cpuCores)) {
-			combinations = append(combinations, fmt.Sprintf("%.0f CPU/%.0fGB", cpuCores, config.MemoryGbs))
-		} else {
-			combinations = append(combinations, fmt.Sprintf("%.1f CPU/%.0fGB", cpuCores, config.MemoryGbs))
-		}
-	}
-	return strings.Join(combinations, ", ")
-}
-
```
nathanjcochran, line 883: Moved to the new `util` package so it can be shared with the `mcp` package.

internal/tiger/cmd/version.go
```diff
@@ -5,23 +5,18 @@ import (
...
 
 	"github.com/spf13/cobra"
+	"github.com/timescale/tiger-cli/internal/tiger/config"
 )
 
-// These variables are set at build time via ldflags in the GoReleaser pipeline
-// for production releases. Default values are used for local development builds.
-var Version = "dev"
-var BuildTime = "unknown"
-var GitCommit = "unknown"
```
nathanjcochran, line 14: Moved to the `config` package so they can be referenced by the `mcp` package.

internal/tiger/mcp/server.go
```diff
@@ -0,0 +1,190 @@
...
+		Title: "List Database Services",
+		Description: `List all database services in your current TigerData project.
+
+This tool retrieves a complete list of database services with their basic information including status, type, region, and resource allocation. Use this to get an overview of all your services before performing operations on specific services.
+
+Perfect for:
+- Getting an overview of your database infrastructure
+- Finding service IDs for other operations
+- Checking service status and resource allocation
+- Discovering services across different regions`,
```
nathanjcochran: These tool definitions can probably be improved. But I don't want to get too hung up on that now - it will be easy to iterate and refine them later, in follow-up PRs.

internal/tiger/mcp/server.go
```diff
@@ -0,0 +1,190 @@
...
+	cfg, err := config.Load()
+	if err != nil {
+		return "", fmt.Errorf("failed to load config: %w", err)
+	}
+
+	if cfg.ProjectID == "" {
+		return "", fmt.Errorf("project ID is required. Please run 'tiger auth login' with --project-id")
+	}
+	return cfg.ProjectID, nil
+}
```
nathanjcochran, line 96: I decided that each tool call should invoke these methods to create the API client and load the config, rather than invoking them once at the start when creating the MCP server. That way, if a user updates their API Key (i.e. logs in) or changes a config value (e.g. their project ID) after initializing the MCP server, the changes will take effect immediately for the next tool call, without requiring the user to re-connect the MCP or restart the agent.

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+}
+
+func (ServiceShowInput) Schema() *jsonschema.Schema {
+	schema := util.Must(jsonschema.For[ServiceShowInput](nil))
+
+	schema.Properties["service_id"].Description = "The unique identifier of the service to show details for. Use tiger_service_list to find service IDs."
+	schema.Properties["service_id"].Examples = []any{"fgg3zcsxw4"}
+
+	return schema
+}
```
nathanjcochran, line 60: The `github.com/modelcontextprotocol/go-sdk/mcp` package would internally call `jsonschema.For` to generate the input JSON schema for us when calling `mcp.AddTool` if we didn't provide an explicit `InputSchema`. But the JSON schemas that are automatically inferred from structs don't capture some important info (such as descriptions, examples, and enum value). I therefore decided on a middle-ground route, where we infer the schema via an explicit call to `jsonschema.For` (which ensures the schema matches our structs), and then amend it with additional information as necessary. Open to thoughts/feedback on this pattern though!

internal/tiger/cmd/mcp.go
```diff
@@ -0,0 +1,257 @@
...
+
+The MCP server provides programmatic access to TigerData Cloud Platform resources
+through Claude and other AI assistants. It exposes Tiger CLI functionality as MCP
+tools that can be called by AI agents.
+
+Configuration:
+The server automatically uses the CLI's stored authentication and configuration.
+No additional setup is required beyond running 'tiger auth login'.
+
+Available Tools:
```
cevian: nit not sure this is a great idea as this list is going to bitrot quickly

internal/tiger/cmd/mcp.go
```diff
@@ -0,0 +1,257 @@
...
+			logging.Error("HTTP server error", zap.Error(err))
+		}
+	}()
+
+	// Wait for context cancellation. Once canceled, stop handling signals and
+	// revert to default signal handling behavior. This allows a second
+	// SIGINT/SIGTERM to forcibly kill the server (useful if there's currently
+	// an active MCP session but you want to kill it anyways).
+	<-ctx.Done()
+	stop()
```
cevian, line 211: don't we have a defer stop(); above. Either delete the defer or this.

internal/tiger/cmd/mcp.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// Shutdown server gracefully
+	logging.Info("Shutting down HTTP server...")
+	return httpServer.Shutdown(context.Background())
+}
+
+// getListener finds an available port starting from the specified port and returns the listener
+func getListener(host string, startPort int) (net.Listener, int, error) {
+	for port := startPort; port < startPort+100; port++ {
```
cevian, line 220: We should probably add a `--force-exact-port` option to disable this auto-finding behavior

internal/tiger/cmd/service.go
cevian: consider moving the emojis in there. Right now very disconnected

internal/tiger/mcp/server.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/timescale/tiger-cli/internal/tiger/api"
+	"github.com/timescale/tiger-cli/internal/tiger/config"
+	"github.com/timescale/tiger-cli/internal/tiger/logging"
+)
+
+const (
+	// TODO: Is this right?
+	serverName  = "tiger-mcp"
+	serverTitle = "Tiger MCP"
```
cevian, line 18: probably want something more descriptive here

internal/tiger/mcp/server.go
cevian: can you please make a linear ticket for improving these

internal/tiger/mcp/server.go
cevian: sound good. add this decision to the spec please.

internal/tiger/mcp/service_tools.go
cevian: can't we use the `description` and `examples` field tag instead? https://github.com/swaggest/jsonschema-go?tab=readme-ov-file

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+
+// ServiceShowOutput represents output for tiger_service_show
+type ServiceShowOutput struct {
```
cevian, line 63: why do we need this if it just contains a serviceDetail?

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+	schema.Properties["region"].Default = util.Must(json.Marshal("us-east-1"))
+	schema.Properties["region"].Examples = []any{"us-east-1", "us-west-2", "eu-west-1", "eu-central-1", "ap-southeast-1"}
+
+	cpuMemoryCombinations := util.GetAllowedCPUMemoryConfigs().Strings()
+	schema.Properties["cpu_memory"].Description = "CPU and memory allocation combination. Choose from the available configurations."
+	schema.Properties["cpu_memory"].Enum = util.AnySlice(cpuMemoryCombinations)
+	schema.Properties["cpu_memory"].Default = util.Must(json.Marshal(cpuMemoryCombinations[0])) // Default to smallest config
+
+	schema.Properties["replicas"].Description = "Number of high-availability replicas for fault tolerance. Higher replica counts increase cost but improve availability."
+	schema.Properties["replicas"].Minimum = util.Ptr(0.0)
```
cevian, line 114: let's put in a max of 5?

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+	// TODO: schema.Properties["replicas"].Maximum =
+	schema.Properties["replicas"].Default = util.Must(json.Marshal(0))
+	schema.Properties["replicas"].Examples = []any{0, 1, 2}
+
+	schema.Properties["wait"].Description = "Whether to wait for the service to be fully ready before returning. Recommended for scripting."
+	schema.Properties["wait"].Default = util.Must(json.Marshal(true))
+	schema.Properties["wait"].Examples = []any{true, false}
+
+	schema.Properties["timeout"].Description = "Timeout in minutes when waiting for service to be ready. Only used when 'wait' is true."
+	schema.Properties["timeout"].Minimum = util.Ptr(0.0)
```
cevian, line 124: let's add a maximim

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+			result, err := util.SavePasswordWithResult(api.Service(service), initialPassword)
+			output.PasswordStorage = &result
+			if err != nil {
+				logging.Debug("MCP: Password storage failed", zap.Error(err))
+			} else {
+				logging.Debug("MCP: Password saved successfully", zap.String("method", result.Method))
+			}
+		}
+
+		// If wait is requested (the default), wait for service to be ready
```
cevian: not sure this should be the default in cmp (can address in followup pr)

internal/tiger/mcp/service_tools.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+
+	// Load fresh project ID from current config
+	projectID, err := s.loadProjectID()
+	if err != nil {
+		return nil, ServiceCreateOutput{}, err
+	}
+
+	// Auto-generate service name if not provided
+	if input.Name == "" {
```
cevian, line 372: can all of this default setting be shared with cli (can be followup pr)

internal/tiger/mcp/server.go
nathanjcochran: [AGE-160](https://linear.app/tigerdata/issue/AGE-160/tiger-cli-improve-service-tool-definitions)

internal/tiger/mcp/service_tools.go
nathanjcochran: That is a different JSON schema package.

The [github.com/modelcontextprotocol/go-sdk/mcp](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp) package depends specifically on [github.com/google/jsonschema-go/jsonschema](https://pkg.go.dev/github.com/google/jsonschema-go@v0.2.3/jsonschema), which does not accept the same struct tags. It _does_ accept a single struct tag containing the property description, per the [docs](https://pkg.go.dev/github.com/google/jsonschema-go@v0.2.3/jsonschema#For):

> This function recognizes struct field tags named "jsonschema". A jsonschema tag on a field is used as the description for the corresponding property. 

However, it does not accept examples, enums, minimums/maximums, or any other information via struct tags. I did try putting the descriptions in `jsonschema:` struct tags at one point, but they were harder to read when they were crammed in there, imo. And given that we need to add the examples, enums, etc. outside of the struct definition anyways, I decided to just move the descriptions out too. But I'm totally flexible on that if you feel strongly one way or the other.

internal/tiger/cmd/mcp.go
nathanjcochran: Yeah, that's fair. Removed the list of available tools here: https://github.com/timescale/tiger-cli/pull/12/commits/bde3f121d85dbb27e58307d608e2b57f588cc2f5

internal/tiger/cmd/mcp.go
nathanjcochran: The `stop()` function is idempotent and safe to call multiple times. The deferred call ensures it's called on error or panic. The call on line 215 implements the behavior explained in the comment above:

>	// Wait for context cancellation. Once canceled, stop handling signals and
>	// revert to default signal handling behavior. This allows a second
>	// SIGINT/SIGTERM to forcibly kill the server (useful if there's currently
>	// an active MCP session but you want to kill it anyways).

By calling `stop()` explicitly here (before triggering the HTTP server to gracefully shutdown, and therefore before the function returns), the custom signal handling logic is removed, and it therefore reverts to the [default Go behavior](https://pkg.go.dev/os/signal#hdr-Default_behavior_of_signals_in_Go_programs), which is to exit the program immediately. This means that the first `ctrl-c` triggers graceful shutdown (i.e. waiting for all outstanding HTTP requests to complete), and the second `ctrl-c` immediately exits the program (a pattern I've seen elsewhere).

I added this because it was otherwise impossible to forcibly kill the MCP server while Claude Code was connected to it (without running `kill -9` in other terminal), because the Streamable HTTP transport apparently involves opening long-lived HTTP requests from the MCP client to the server.

I'm open to changing how this works, but as it's currently written, but removing either of the `stop` calls would cause issues.

nathanjcochran: Left a comment about the idempotency of `stop()`, which will hopefully help make this more clear: https://github.com/timescale/tiger-cli/pull/12/commits/8f1b446c9a5ee24461212774da3597ea8276b319

internal/tiger/cmd/mcp.go
nathanjcochran: Created [AGE-161](https://linear.app/tigerdata/issue/AGE-161/tiger-cli-add-force-exact-port-to-prevent-port-finding) about adding a `--force-exact-port` option.

Although imo, we might also want to consider getting rid of the port-finding logic altogether and just fail with an error if the given port isn't available, given that clients must be configured with a specific port, and I can therefore see very few situations where it would be useful for it to start up a different port than expected.

internal/tiger/mcp/service_tools.go
nathanjcochran: I don't have a strong opinion either way. But it is kind of nice that each tool has an `Input` struct and an `Output` struct. `ServiceDetail` is shared/reused across multiple tool calls, so it couldn't be renamed to `ServiceShowOutput` if we made it the top-level response from the `tiger_service_show` tool call (which would therefore break the `Input`/`Output` struct naming pattern).

The current pattern does introduce an extra level of nesting, but it also makes `ServiceShowOutput` better match the structure of [ServiceCreateOutput](https://github.com/timescale/tiger-cli/blob/8f1b446c9a5ee24461212774da3597ea8276b319/internal/tiger/mcp/service_tools.go#L132-L137) (where `ServiceDetail` is a nested field), and provides room to add other top-level fields in the future, if need be (like in `ServiceCreateOutput`).

internal/tiger/mcp/service_tools.go
nathanjcochran: Any suggestion for what it should be? If we default to 30 minutes (which is already a fairly long time), should the max be an hour? Two hours? Where do we draw the line (or, what's the maximum amount of time that a service can take to spin up)?

internal/tiger/mcp/service_tools.go
nathanjcochran: Set max number of replicas to 5 here: https://github.com/timescale/tiger-cli/pull/12/commits/154750efaed3ee113dcb31fcc1ae8475a2d38a35

internal/tiger/mcp/service_tools.go
nathanjcochran: I had the same thought tbh, but I was just following [the spec](https://github.com/timescale/tiger-cli/blob/956dacbfba4d4a9bf005cfb2fff428c46cc797e5/specs/spec_mcp.md?plain=1#L109) on this one 😅.

I'll update the code to make it not wait by default, because I do think the several-minute wait time really interrupts the flow of working with the agent.

nathanjcochran: Made `wait: false` the default here: https://github.com/timescale/tiger-cli/pull/12/commits/3ab3f0412009bca50b26290d53f8e59ab10b892f

internal/tiger/mcp/service_tools.go
nathanjcochran: Yup, we can definitely continue to consolidate some of this duplicate code in follow-ups 👍

internal/tiger/mcp/server.go
nathanjcochran: Added the decision to the spec here: https://github.com/timescale/tiger-cli/pull/12/commits/3c927bab62a149104f6c5afe13876480953ff947

PR #13, opened 2025-09-16
Publish Linux packages to Packagecloud

nathanjcochran:
This is a follow-up to #9. It wires up the release pipeline to push Linux packages to a new [`tiger-cli` repo](https://packagecloud.io/timescale/tiger-cli) in Packagecloud. I decided to make a new repo, rather than reuse the existing [`timescaledb` repo](https://packagecloud.io/timescale/timescaledb), because Tiger CLI seemed pretty distinct from TimescaleDB and the other related packages that are currently in there (and because I didn't want to start polluting that repo with my testing 😅). But I'm totally open to reusing the timescaledb repo if that's what people think is best (should be an easy swap) - just let me know.

See the installation instructions in the README for information on how to configure the repository and install packages. Note that I have not yet tested this myself (I plan to test Ubuntu soon, but I don't have easy access to the other Linux distros at the moment).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ curl -s https://packagecloud.io/install/repositories/timescale/tiger-cli/script.
 sudo apt-get install tiger-cli
 ```
 
+For manual repository installation instructions, see [here](https://packagecloud.io/timescale/tiger-cli/install#manual-deb).
+
 ### Red Hat/CentOS/Fedora
 ```bash
 # Add repository
@@ -63,6 +65,8 @@ curl -s https://packagecloud.io/install/repositories/timescale/tiger-cli/script.
 # Install tiger-cli
 sudo yum install tiger-cli
 ```
+
+For manual repository installation instructions, see [here](https://packagecloud.io/timescale/tiger-cli/install#manual-rpm).
 
 ## Usage
 


.goreleaser.yaml
```diff
@@ -65,7 +65,6 @@ nfpms:
     formats:
       - deb
       - rpm
-      - apk
```
nathanjcochran, line 68: Packagecloud does not appear to accept Alpine packages. I therefore decided to just stop building them for now, as we have no good story for how to install them. Also, Alpine is not a very common distro outside of docker images (to my knowledge).

README.md
```diff
@@ -46,18 +46,22 @@ brew install timescale/tap/tiger-cli
 go install github.com/timescale/tiger-cli/cmd/tiger@latest
 ```
 
-### Direct binary download
```
nathanjcochran, line 49: I decided to remove the "Direct binary download" section, as macOS binaries installed via this method will currently be quarantined and prevented from opening via Gatekeeper.

If/when we have code signing and notarization sorted out (see #11), we can consider bringing this option back. But I'd rather not encourage users to do it if it won't work on all operating systems.

README.md
```diff
@@ -46,18 +46,22 @@ brew install timescale/tap/tiger-cli
...
+### Debian/Ubuntu
 ```bash
-curl -LO https://github.com/timescale/tiger-cli/releases/latest/download/tiger-cli_$(uname -s)_$(uname -m).tar.gz
-tar -xzf tiger-cli_$(uname -s)_$(uname -m).tar.gz
-chmod +x tiger
-mkdir -p ~/.local/bin && mv tiger ~/.local/bin/
+# Add PackageCloud repository
+curl -s https://packagecloud.io/install/repositories/timescale/tiger-cli/script.deb.sh | sudo bash
```
nathanjcochran, line 52: These installation scripts provided by Packagecloud make configuring the package repository very easy. It handles installing necessary dependencies, importing gpg keys, creating the `/etc/apt/sources.list.d/timescale_tiger-cli.list` file, etc.

We could consider including the manual instructions here instead (i.e. if we don't like asking users to run a script from a third-party), but it would complicate these instructions a bit (it would look more like [this](https://github.com/timescale/tiger-cli/compare/nathan/packagecloud...nathan/packagecloud-manual-repo)). I'm open to whatever though.

cevian: Yeah I think this is right. I would add "For manual installation instructions see here" and link them to https://packagecloud.io/timescale/tiger-cli/install#manual

nathanjcochran: Good call, done here: https://github.com/timescale/tiger-cli/pull/13/commits/baf20a87bb3f37682e92ed1e7fe0b10a526596fa

PR #14, opened 2025-09-19
Fix console URL

nathanjcochran:
See [Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1758308042133889).

Fixes a couple instances of an incorrect console URL. Updates it from `https://console.timescale.com` to `https://console.cloud.timescale.com`

### DIFF
--- a/internal/tiger/cmd/auth.go
+++ b/internal/tiger/cmd/auth.go
@@ -64,7 +64,7 @@ The project ID will be stored in the configuration file.
 You may also provide API keys via flags or environment variables, in which case
 they will be used directly. The CLI will prompt for any missing information.
 
-You can find your API credentials and project ID at: https://console.timescale.com/dashboard/settings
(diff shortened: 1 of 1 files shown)


internal/tiger/cmd/auth.go
```diff
@@ -64,7 +64,7 @@ The project ID will be stored in the configuration file.
 You may also provide API keys via flags or environment variables, in which case
 they will be used directly. The CLI will prompt for any missing information.
 
-You can find your API credentials and project ID at: https://console.timescale.com/dashboard/settings
+You can find your API credentials and project ID at: https://console.cloud.timescale.com/dashboard/settings
```
nathanjcochran, line 67: I decided to leave this URL hard-coded to the prod URL, since we haven't loaded the config at this point in the process, and I didn't want to mess with that (right now, at least). For the URL used in `promptForCredentials` below, I'm using the `ConsoleURL` from the config (which was easy to do there), so it should be accurate it dev too.

PR #16, opened 2025-09-20
Add service start/stop commands

mfreed:
Adds tiger service start and tiger service stop commands to manage database service lifecycle.

  New Commands

  - tiger service start [service-id] - Start a stopped service
  - tiger service stop [service-id] - Stop a running service

  Both commands support --no-wait and --wait-timeout flags for async operation control.

  Features

  - Status monitoring: Real-time progress with completion indicators
  - Proper error handling: New ExitConflict exit code for state conflicts (409 responses)
  - Complete error mapping: 401/403/404/409 responses map to appropriate exit codes
  - Integration testing: Enhanced lifecycle and error scenario test coverage

  Usage

  // Start/stop with completion wait (default)
  tiger service start svc-12345
  tiger service stop svc-12345

  // Return immediately after request accepted
  tiger service start svc-12345 --no-wait

  Technical Changes

  - Updated OpenAPI spec and regenerated API client
  - Added waitForServicePaused() function following existing patterns
  - Enhanced integration tests with start/stop validation
  - Added ExitConflict = 7 exit code for HTTP 409 responses

  Compatibility

  Fully backward compatible. No breaking changes to existing functionality.

  ---
  Co-Authored-By: Claude noreply@anthropic.com

### DIFF
--- a/internal/tiger/cmd/errors.go
+++ b/internal/tiger/cmd/errors.go
@@ -9,6 +9,7 @@ const (
 	ExitAuthenticationError = 4 // Authentication error
 	ExitPermissionDenied    = 5 // Permission denied
 	ExitServiceNotFound     = 6 // Service not found
+	ExitConflict            = 7 // Resource conflict (e.g., service cannot be modified in current state)
 )
 
 // exitCodeError creates an error that will cause the program to exit with the specified code
--- a/internal/tiger/cmd/integration_test.go
+++ b/internal/tiger/cmd/integration_test.go
@@ -358,6 +358,130 @@ func TestServiceLifecycleIntegration(t *testing.T) {
 		t.Logf("✅ psql command with updated password succeeded")
 	})
 
+	t.Run("StopService", func(t *testing.T) {
+		if serviceID == "" {
+			t.Skip("No service ID available from create test")
+		}
+
+		t.Logf("Stopping service: %s", serviceID)
+
+		output, err := executeIntegrationCommand(
+			"service", "stop", serviceID,
+			"--wait-timeout", "10m", // Longer timeout for integration tests
+		)
+
+		if err != nil {
+			// Check if it's a conflict error (service already stopped)
+			if exitErr, ok := err.(interface{ ExitCode() int }); ok && exitErr.ExitCode() == ExitConflict {
+				t.Logf("Service is already stopped (conflict error): %v", err)
+			} else {
+				t.Fatalf("Service stop failed: %v\nOutput: %s", err, output)
+			}
+		} else {
+			// Verify stop success message
+			if !strings.Contains(output, "Stop request accepted") &&
+			   !strings.Contains(output, "stopped successfully") {
+				t.Errorf("Expected stop success message, got: %s", output)
+			}
+			t.Logf("Service stop completed successfully")
+		}
+	})
+
+	t.Run("VerifyServiceStopped", func(t *testing.T) {
+		if serviceID == "" {
+			t.Skip("No service ID available from create test")
+		}
+
+		t.Logf("Verifying service is stopped")
+
+		output, err := executeIntegrationCommand("service", "describe", serviceID, "--output", "json")
+		if err != nil {
+			t.Fatalf("Failed to describe service after stop: %v\nOutput: %s", err, output)
+		}
+
+		// Parse JSON to check status
+		var service api.Service
+		if err := json.Unmarshal([]byte(output), &service); err != nil {
+			t.Fatalf("Failed to parse service JSON: %v", err)
+		}
+
+		var status string
+		if service.Status != nil {
+			status = string(*service.Status)
+		}
+
+		t.Logf("Service status after stop: %s", status)
+
+		// The status should be PAUSED for stopped services
+		if status != "PAUSED" {
+			t.Logf("Warning: Expected service status to be PAUSED, got %s. This might be expected depending on timing or service state.", status)
+		} else {
+			t.Logf("✅ Service is correctly in PAUSED state")
+		}
+	})
+
+	t.Run("StartService", func(t *testing.T) {
+		if serviceID == "" {
+			t.Skip("No service ID available from create test")
+		}
+
+		t.Logf("Starting service: %s", serviceID)
+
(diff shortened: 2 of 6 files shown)


internal/tiger/cmd/service.go
```diff
@@ -781,6 +783,46 @@ func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID s
...
+
+			service := *resp.JSON200
+			status := formatDeployStatus(service.Status)
+
+			switch status {
+			case "PAUSED":
+				fmt.Fprintf(cmd.OutOrStdout(), "⏹️  Service has been stopped successfully!\n")
+				return nil
+			case "FAILED", "ERROR":
+				return fmt.Errorf("service stop operation failed with status: %s", status)
```
cevian, line 818: nit: hmm can we make this error message more precise? We know the services is failed, we don't actually know for sure it's because of the pause. If you could also fix this in the wait function above this would be great.
```suggestion
				return fmt.Errorf("service failed with status: %s", status)
```

internal/tiger/cmd/integration_test.go
```diff
@@ -358,6 +358,130 @@ func TestServiceLifecycleIntegration(t *testing.T) {
...
+		t.Logf("Service status after stop: %s", status)
+
+		// The status should be PAUSED for stopped services
+		if status != "PAUSED" {
+			t.Logf("Warning: Expected service status to be PAUSED, got %s. This might be expected depending on timing or service state.", status)
+		} else {
+			t.Logf("✅ Service is correctly in PAUSED state")
+		}
+	})
+
```
cevian, line 422: Here I would re-run the stop command to make sure that there are no errors (or there are errors depending on what you think should happen) when you stop an already stopped service.

internal/tiger/cmd/integration_test.go
```diff
@@ -358,6 +358,130 @@ func TestServiceLifecycleIntegration(t *testing.T) {
...
+		t.Logf("Service status after start: %s", status)
+
+		// The status should be READY for started services
+		if status != "READY" {
+			t.Logf("Warning: Expected service status to be READY, got %s. This might be expected depending on timing or service state.", status)
+		} else {
+			t.Logf("✅ Service is correctly in READY state")
+		}
+	})
+
```
cevian, line 484: similarly, lets test what happens when you start an already-running service.

internal/tiger/cmd/service.go
```diff
@@ -1063,3 +1105,195 @@ func formatAllowedMemoryValues(configs []CPUMemoryConfig) string {
...
+
+				// Wait for service to become ready
+				return waitForServiceReady(client, cfg.ProjectID, serviceID, startWaitTimeout, cmd)
+			case 404:
+				return exitWithCode(ExitServiceNotFound, fmt.Errorf("service '%s' not found in project '%s'", serviceID, cfg.ProjectID))
+			case 401:
+				return exitWithCode(ExitAuthenticationError, fmt.Errorf("authentication failed: invalid API key"))
+			case 403:
+				return exitWithCode(ExitPermissionDenied, fmt.Errorf("permission denied: insufficient access to start service"))
+			case 409:
```
cevian, line 1190: is this what happens when you start an already running service?

internal/tiger/cmd/errors.go
```diff
@@ -9,6 +9,7 @@ const (
 	ExitAuthenticationError = 4 // Authentication error
 	ExitPermissionDenied    = 5 // Permission denied
 	ExitServiceNotFound     = 6 // Service not found
+	ExitConflict            = 7 // Resource conflict (e.g., service cannot be modified in current state)
```
cevian, line 12: do we want a separate code here? why not ExitGeneralError? I am just not sure myself when this error happens and I am not sure users will know either

nathanjcochran:
I created a new branch off of this one, and opened a new PR: https://github.com/timescale/tiger-cli/pull/103

It contains the same changes, but addresses the code review feedback and updates a few things to adhere to our current patterns.

internal/tiger/cmd/service.go
nathanjcochran: Done in https://github.com/timescale/tiger-cli/pull/103

internal/tiger/cmd/integration_test.go
nathanjcochran: Done in https://github.com/timescale/tiger-cli/pull/103

internal/tiger/cmd/integration_test.go
nathanjcochran: Done in https://github.com/timescale/tiger-cli/pull/103

internal/tiger/cmd/service.go
nathanjcochran: Trying to start stop a service that is already stopped (or start a service that is already started) fails with a `400 Bad Request` error in the GraphQL layer, which is passed through the REST API layer. So this logic was actually incorrect - `409 Conflict` is never currently returned.

Fixed in https://github.com/timescale/tiger-cli/pull/103

internal/tiger/cmd/errors.go
nathanjcochran: I removed the separate exit code in https://github.com/timescale/tiger-cli/pull/103, largely because it wasn't actually wired up correctly (see my comment above about `409 Conflict` never actually being returned from the API).

The `400 Bad Request` returned from the REST API is now being translated into `ExitInvalidParameters` (exit code 3) by [exitWithErrorFromStatusCode](https://github.com/timescale/tiger-cli/blob/ee37221b2fb88bd22c123734d509b37f767774ce/internal/tiger/cmd/errors.go#L45-L47). I'm not sure if that's the _best_ exit code, but I think it's probably fine to now, given that our current pattern is to derive exit codes from HTTP status codes. If we want some other exit code, I think we should fix it upstream by returning a different status code from the REST API (though I think that can be a follow-up).

nathanjcochran: Created [AGE-249](https://linear.app/tigerdata/issue/AGE-249/tiger-cli-service-startstop-api-endpoints-should-return-409-if-service) about following up to change the status code upstream.

PR #19, opened 2025-09-22
Add password_storage config file support for persistent configuration

cevian:
## Summary

Fixes AGE-157

The global `--password-storage` flag was available but not persisted in config files, making it impossible to configure password storage globally and persistently. This was particularly problematic for MCP tools that needed consistent configuration.

## Changes

- ✅ Add `PasswordStorage` field to Config struct with proper mapstructure/yaml tags
- ✅ Add `DefaultPasswordStorage` constant and use it consistently for CLI flag and config defaults  
- ✅ Add `password_storage` support to all config operations (`set`, `unset`, `reset`, `show`)
- ✅ Add validation for `password_storage` values (`keyring`, `pgpass`, `none`)
- ✅ Add `password_storage` to config show output (table, JSON, and YAML formats)
- ✅ Add comprehensive test coverage for all `password_storage` config operations

## Result

Users can now persistently configure password storage:

```bash
# Set password storage method
tiger config set password_storage pgpass

# Verify it's saved
tiger config show
# Shows: Password Storage: pgpass

# All precedence levels work correctly:
# CLI flag > Environment variable > Config file > Default
```

This enables MCP tools to configure password storage once and have it apply globally and persistently across all CLI operations.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/internal/tiger/cmd/config_test.go
+++ b/internal/tiger/cmd/config_test.go
@@ -330,6 +330,12 @@ func TestConfigSet_ValidValues(t *testing.T) {
 				if cfg.Analytics != expected {
 					t.Errorf("Expected Analytics %t, got %t", expected, cfg.Analytics)
 				}
+			case "password_storage":
+				if cfg.PasswordStorage != tt.value {
+					t.Errorf("Expected PasswordStorage %s, got %s", tt.value, cfg.PasswordStorage)
+				}
+			default:
+				t.Fatalf("Unhandled test case for key: %s", tt.key)
 			}
 		})
 	}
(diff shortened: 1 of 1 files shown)


internal/tiger/cmd/config_test.go
```diff
@@ -216,6 +228,9 @@ func TestConfigSet_ValidValues(t *testing.T) {
 		{"service_id", "new-service", "Set service_id = new-service"},
 		{"output", "json", "Set output = json"},
 		{"analytics", "false", "Set analytics = false"},
+		{"password_storage", "pgpass", "Set password_storage = pgpass"},
+		{"password_storage", "none", "Set password_storage = none"},
+		{"password_storage", "keyring", "Set password_storage = keyring"},
```
nathanjcochran, line 290: I think you also need to update the switch statement below (line 254) to contain a `password_storage` case (might also want to add a `default` case that fails the test, so we don't miss it again).

internal/tiger/cmd/config_test.go
```diff
@@ -332,6 +350,7 @@ func TestConfigUnset_ValidKeys(t *testing.T) {
 		{"project_id", "Unset project_id"},
 		{"service_id", "Unset service_id"},
 		{"output", "Unset output"},
+		{"password_storage", "Unset password_storage"},
```
nathanjcochran, line 453: Similarly, need to update the switch statement below (line 374) to include a `password_storage` case.

PR #20, opened 2025-09-23
Fix issues with config file/directory logic

nathanjcochran:
This PR fixes several issues with the config file/directory logic:

1. The config file path passed as a [flag](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/cmd/root.go#L80-L81) was being used when initializing viper (see [here](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/cmd/root.go#L65)  and [here](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/config/config.go#L36)), but the value [loaded into the Config struct](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/config/config.go#L66) was using the result of [GetConfigDir()](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/config/config.go#L180-L191), which does not take the flag into consideration. As a result, when the config file was [saved](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/config/config.go#L79), it would not be saved back to the location it was loaded from (i.e. the location provided via the `--config` flag).
2. The default value for the `--config` flag was being provided by `GetConfigDir()` (see [here](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/cmd/root.go#L80-L81)). That function returned either the value provided via the `TIGER_CONFIG_DIR` env var or the default value. As a result, the flag's default value (i.e. in the help text) would change if the `TIGER_CONFIG_DIR` env var was present, which is not how the defaults for any of the other flags work if their corresponding env vars are set (they always show the "true" default value).
3. The `--config` flag did not match the naming convention of the `TIGER_CONFIG_DIR` env var. Furthermore, the CLI flag expected a full file path, whereas the env var just expected the path to the directory. This was inconsistent with the other flags, which all have a 1:1 correspondence with their env vars.
4. The functions for writing and reading the API key to/from a fall back file used the `GetConfigDir()` function to get the path to the config dir where the API key file is stored (e.g. see [here](https://github.com/timescale/tiger-cli/blob/4df1cb315b3c2bb4b4d7853c64c9259e3403e277/internal/tiger/config/api_key.go#L48)). However, that meant there was no way to control this via a flag, since that function did not take the flag into account (only the `TIGER_CONFIG_DIR` env var).

I also fixed and cleaned up a bunch of tests, and added some new tests to cover some of these issues and prevent them from happening again.

### DIFF
--- a/internal/tiger/config/config.go
+++ b/internal/tiger/config/config.go
@@ -37,7 +37,7 @@ func SetupViper(configDir string) error {
 	configFile := GetConfigFile(configDir)
 	viper.SetConfigFile(configFile)
 
-	// Configure vipe to read from env vars
+	// Configure viper to read from env vars
 	viper.SetEnvPrefix("TIGER")
 	viper.AutomaticEnv()
 
@@ -188,6 +188,9 @@ func GetConfigFile(dir string) string {
 	return filepath.Join(dir, ConfigFileName)
 }
 
+// TODO: This function is currently used to get the directory that the API
+// key fallback file should be stored in (see api_key.go). But ideally, those
+// functions would take a Config struct and use the ConfigDir field instead.
 func GetConfigDir() string {
 	return filepath.Dir(viper.ConfigFileUsed())
 }


internal/tiger/cmd/root.go
```diff
@@ -77,8 +76,7 @@ tiger auth login
 	cobra.OnInitialize(initConfigFunc)
 
 	// Add persistent flags
-	defaultConfigFile := filepath.Join(config.GetConfigDir(), config.ConfigFileName)
-	cmd.PersistentFlags().StringVar(&cfgFile, "config", defaultConfigFile, "config file")
```
nathanjcochran, line 81: I changed the flag from `--config` (which accepted the full file path to the config file) to `--config-dir` (which accepts the path to the directory containing the config file) to match the `TIGER_CONFIG_DIR` env var, and to make it clear that this config directory is used for both the `config.yaml` file as well as the API key fallback file. This is also in-line with the [original spec](https://github.com/timescale/tiger-cli/blob/6c9fe8453dbc3f6176c4ae121a99a216f5d0a935/specs/spec.md?plain=1#L50).

I also updated it to use `GetDefaultConfigDir` as the default value (which is shown in the help output), which returns the "true" default value whether or not `TIGER_CONFIG_DIR` is set.

internal/tiger/config/config.go
```diff
@@ -53,6 +59,11 @@ func SetupViper(configFile string) error {
 		if err := viper.ReadInConfig(); err != nil {
 			return fmt.Errorf("error reading config file: %w", err)
 		}
+
+		// Configure viper to watch for file changes and update its in-memory
+		// representation of the config. Note that this won't automatically
+		// update [Config] structs already returned from [Load].
+		viper.WatchConfig()
```
nathanjcochran, line 64: This ensures that config file changes made while the MCP server is running take effect immediately (i.e. for the next tool call). This depends on the fact that the config is re-loaded via a call to `Load()` for each MCP tool call.

internal/tiger/config/config.go
```diff
@@ -66,7 +77,7 @@ func Load() (*Config, error) {
 		ConfigDir: GetConfigDir(),
 	}
 
-	// Use the global Viper instance that's already configured by initConfig() and bindFlags()
```
nathanjcochran, line 69: Neither of these function (`initConfig()` or `bindFlags()`) exist anywhere in the code base. Not sure where this comment came from, but I updated it to refer to the `SetupViper` function above, which seems most accurate.

internal/tiger/config/config.go
```diff
@@ -178,10 +186,10 @@ func (c *Config) Reset() error {
 }
 
 func GetConfigDir() string {
-	if dir := os.Getenv("TIGER_CONFIG_DIR"); dir != "" {
-		return expandPath(dir)
-	}
+	return filepath.Dir(viper.ConfigFileUsed())
+}
```
nathanjcochran, line 196: This function was being called to construct the file path that the API key is stored in when the keychain is not available (e.g. [here](https://github.com/timescale/tiger-cli/blob/a91b34949266fe5fcb7c8715f76469ca886cc089/internal/tiger/config/api_key.go#L48)), but was not taking the flag into consideration. I redefined it to accurately reflect the value of the flag (if provided), or env var, or default. It would probably be better to pass the config struct into those functions and use the `ConfigDir` field, but that's a bit of a larger refactor I don't want to undertake at the present moment.

internal/tiger/config/config.go
```diff
@@ -190,6 +196,18 @@ func GetConfigDir() string {
...
+	if configDirFlag.Changed {
+		return expandPath(configDirFlag.Value.String())
+	}
+
+	if dir := os.Getenv("TIGER_CONFIG_DIR"); dir != "" {
+		return expandPath(dir)
+	}
+
+	return GetDefaultConfigDir()
+}
```
nathanjcochran, line 217: This new function is responsible for returning the `--config-dir` flag value (if set), or the `TIGER_CONFIG_DIR` env var value (if set), or the default value.

internal/tiger/cmd/auth_test.go
```diff
@@ -729,101 +729,3 @@ func TestAuthLogout_Success(t *testing.T) {
 		t.Fatal("API key should be removed after logout")
 	}
 }
-
-func TestStoreAPIKeyToFile(t *testing.T) {
```
nathanjcochran, line 733: Moved into `./config/api_key_test.go`. Should have moved these tests when I moved the corresponding functions in #12.

internal/tiger/cmd/root.go
```diff
@@ -61,24 +60,17 @@ tiger auth login
...
+		if err := config.SetupViper(config.GetEffectiveConfigDir(configDirFlag)); err != nil {
 			fmt.Fprintf(os.Stderr, "Error setting up config: %v\n", err)
 			os.Exit(1)
 		}
-
-		if debug {
-			if configFile := viper.ConfigFileUsed(); configFile != "" {
-				fmt.Fprintln(os.Stderr, "Using config file:", configFile)
-			}
-		}
```
nathanjcochran, line 74: This was duplicative with the information logged [here](https://github.com/timescale/tiger-cli/blob/6c9fe8453dbc3f6176c4ae121a99a216f5d0a935/internal/tiger/cmd/root.go#L49), and it was generating a lot of noise in the `go test` output.

PR #21, opened 2025-09-23
Add proxy to docs mcp server

nathanjcochran:
Closes: AGE-145

Adds a proxy to the hosted docs MCP server (https://github.com/timescale/tiger-docs-mcp-server).

By default, all tools, resources, resource templates, and prompts from the docs MCP server are proxied and exposed on the Tiger CLI MCP server (although as of right now, I think the docs MCP only exposes tools).

The proxied tools can be disabled via the new `docs_mcp_enabled` config: `tiger config set docs_mcp_enabled false`.

The URL of the docs MCP can be set via the new `docs_mcp_url` config. For example, to connect to the dev docs MCP (which is the only one that exists at the moment, since it hasn't been deployed to prod yet), use: `tiger config set docs_mcp_url https://mcp.dev-us-east-1.ops.dev.timescale.com/docs`

### How to Test

1. Pull down this branch and build it with `go install ./...`
2. Log in with `tiger auth login`
3. Set the `docs_mcp_url` config to point to the dev docs server: `tiger config set docs_mcp_url https://mcp.dev-us-east-1.ops.dev.timescale.com/docs`
4. Configure Claude Code to use the MCP server: `claude mcp add tiger-mcp -- tiger mcp start`
5. Run Claude Code: `claude`
6. Run `/mcp` to view the installed MCP servers. Select `tiger-mcp`, then `View Tools`. It should show a total of 7 tools: 4 defined via the MCP server itself, and 3 that are proxied from the remove docs MCP.
7. Prompt Claude and try to get it to use one of the docs tools. For example, ask it how to set up a hypertable.

### DIFF
--- a/internal/tiger/cmd/config.go
+++ b/internal/tiger/cmd/config.go
@@ -133,33 +133,35 @@ func buildConfigCmd() *cobra.Command {
 func outputTable(cfg *config.Config, cmd *cobra.Command) error {
 	out := cmd.OutOrStdout()
 	fmt.Fprintln(out, "Current Configuration:")
-	fmt.Fprintf(out, "  API URL:        %s\n", cfg.APIURL)
-	fmt.Fprintf(out, "  Console URL:    %s\n", cfg.ConsoleURL)
-	fmt.Fprintf(out, "  Gateway URL:    %s\n", cfg.GatewayURL)
+	fmt.Fprintf(out, "  API URL:     %s\n", cfg.APIURL)
+	fmt.Fprintf(out, "  Console URL: %s\n", cfg.ConsoleURL)
+	fmt.Fprintf(out, "  Gateway URL: %s\n", cfg.GatewayURL)
 	fmt.Fprintf(out, "  Docs MCP:       %t\n", cfg.DocsMCP)
 	fmt.Fprintf(out, "  Docs MCP URL:   %s\n", cfg.DocsMCPURL)
-	fmt.Fprintf(out, "  Project ID:     %s\n", valueOrEmpty(cfg.ProjectID))
-	fmt.Fprintf(out, "  Service ID:     %s\n", valueOrEmpty(cfg.ServiceID))
-	fmt.Fprintf(out, "  Output:         %s\n", cfg.Output)
-	fmt.Fprintf(out, "  Analytics:      %t\n", cfg.Analytics)
-	fmt.Fprintf(out, "  Debug:          %t\n", cfg.Debug)
-	fmt.Fprintf(out, "  Config Dir:     %s\n", cfg.ConfigDir)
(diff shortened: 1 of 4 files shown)


internal/tiger/mcp/proxy.go
```diff
@@ -0,0 +1,410 @@
...
+			zap.String("resource_uri", remoteURI),
+		)
+
+		if p.session == nil {
+			return nil, fmt.Errorf("not connected to remote MCP server")
+		}
+
+		// Forward the request to remote server with original URI
+		readParams := &mcp.ReadResourceParams{
+			URI: remoteURI, // Use original URI for remote server
```
cevian: I don't think this will work for resource templates since this would hold the template (e.g. `http://my-resource/{name}`) not the uri (e.g. `http://my-resource/cool-resource`). why not get the uri from mcp.ReadResourceRequest

internal/tiger/mcp/proxy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if resourcesResp == nil || len(resourcesResp.Resources) == 0 {
+		logging.Debug("No resources found on remote server")
+		return nil
+	}
+
+	// Register each remote resource as a proxy resource
+	for _, resource := range resourcesResp.Resources {
+		if resource.URI == "" {
+			logging.Warn("Skipping resource with empty URI")
```
cevian, line 209: why would this ever happen and why wouldn't we want to error if it did? (leaving one comment but there are similar things all over the place). Is this an mcp protocol thing?

internal/tiger/mcp/proxy.go
nathanjcochran: Good catch! Fixed here: https://github.com/timescale/tiger-cli/pull/21/commits/8ee0c5d0e0eff440a13b78ee67ff9946f92b81c3

I also realized I could simply the proxy logic to use the passed-in params structs (in most cases) rather than creating new ones, which should also help prevent issues like this.

internal/tiger/mcp/proxy.go
nathanjcochran: > why would this ever happen

Frankly, it's just some extra-defensive code that Claude wrote, and it seemed harmless to me to keep around. I don't think it should ever happen under normal circumstances, but it's true that we wouldn't want to register the resource/tool/prompt if it somehow did (I think that would probably result in other errors or broken tools).

> why wouldn't we want to error if it did?

I don't think we'd want to return an error, as that would prevent other valid resources from being registered/proxied. I think it's best for the proxy stuff to be "best effort", in the sense that it tries its best to register all of the tools/resources/prompts that it can, and simply logs warnings for any it fails to register.

PR #25, opened 2025-09-25
add --with-password to `db connection-string` command

jgpruitt:
By default, `tiger db connection-string` omits the password. Now, `tiger db connection-string --with-password` will include it.

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -60,17 +60,9 @@ Examples:
 				return err
 			}
 
-			connectionString, err := buildConnectionString(service, dbConnectionStringPooled, dbConnectionStringRole, cmd)
+			connectionString, err := buildConnectionString(service, dbConnectionStringPooled, dbConnectionStringRole, dbConnectionStringWithPassword, cmd)
 			if err != nil {
 				return fmt.Errorf("failed to build connection string: %w", err)
-			}
-
-			// Add password to connection string if requested
-			if dbConnectionStringWithPassword {
-				connectionString, err = addPasswordToConnectionString(connectionString, service)
-				if err != nil {
-					return fmt.Errorf("failed to add password to connection string: %w", err)
-				}
 			}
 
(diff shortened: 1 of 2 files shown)


internal/tiger/cmd/db.go
```diff
@@ -58,6 +65,14 @@ Examples:
 				return fmt.Errorf("failed to build connection string: %w", err)
 			}
 
+			// Add password to connection string if requested
+			if dbConnectionStringWithPassword {
+				connectionString, err = addPasswordToConnectionString(connectionString, service)
+				if err != nil {
+					return fmt.Errorf("failed to add password to connection string: %w", err)
+				}
+			}
```
nathanjcochran: I don't love how we're generating a connection string without a password first via `buildConnectionString`, and then parsing it and re-constructing it with a password afterwards if `dbConnectionStringWithPassword` is set to `true`. That seems pretty roundabout to me.

Can't we just pass `dbConnectionStringWithPassword` to `buildConnectionString` and have it build the connection string correctly the first time around?

internal/tiger/cmd/db.go
```diff
@@ -499,3 +515,40 @@ func isConnectionRejected(err error) bool {
...
 	return false
 }
+
+// addPasswordToConnectionString adds the password to a connection string if available
+func addPasswordToConnectionString(connectionString string, service api.Service) (string, error) {
+	// Parse the existing connection string
+	config, err := pgx.ParseConfig(connectionString)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse connection string: %w", err)
+	}
```
nathanjcochran: The fact that we have to parse the connection string we just built, in order to build it again with a password, seems kind of silly to me.

internal/tiger/cmd/db.go
jgpruitt: You're absolutely right! That approach is much cleaner and more efficient. ;)

PR #26, opened 2025-09-26
Add MCP install command and next steps after login

cevian:
## Summary
- Implemented `tiger mcp install` command for configuring MCP servers in AI coding tools
- Added interactive editor selection using Bubble Tea when no editor is specified
- Displays helpful next steps message after successful login

### MCP Install Command
- Support for 6 AI coding tools/editors:
  - Claude Code
  - Cursor
  - Windsurf  
  - Codex
  - Gemini CLI
  - VS Code
- Interactive editor chooser with alphabetical sorting when no editor specified
- Automatic detection of config file locations
- Config file backup before modifications
- Merges with existing MCP server configurations (non-destructive)
- CLI-based installation for Gemini and VS Code

### Login Next Steps
After successful `tiger auth login`, users now see:
```
🎉 Next steps:
• Install MCP server for your favorite AI coding tool: tiger mcp install
• List existing services: tiger service list
• Create a new service: tiger service create
```

### MCP Installation
- JSON-based configuration for Claude Code, Cursor, Windsurf, and Codex
- CLI command execution for Gemini CLI and VS Code
- Robust error handling and config validation
- Uses full Tiger executable path (not relying on PATH)
- Special handling for test environments to ensure consistent test results

### Interactive Editor Selection
- Uses Bubble Tea framework (same as auth project picker)
- Alphabetically sorted list of supported editors
- Clear descriptions for each editor
- Keyboard navigation (arrow keys/j/k, Enter to select, q/Esc to quit)

## Test Coverage
- Comprehensive unit tests for all MCP install functionality
- Tests for config file manipulation and backup
- Tests for CLI-only installations
- Updated auth tests to verify next steps message
- All tests passing ✅

## Examples

```bash
# Interactive editor selection
tiger mcp install

# Install for specific editor
tiger mcp install claude-code
tiger mcp install cursor
tiger mcp install gemini

# With custom config path
tiger mcp install claude-code --config-path ~/custom/mcp.json
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/internal/tiger/cmd/mcp_install.go
+++ b/internal/tiger/cmd/mcp_install.go
@@ -19,6 +19,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/timescale/tiger-cli/internal/tiger/logging"
+	"github.com/timescale/tiger-cli/internal/tiger/mcp"
 	"github.com/timescale/tiger-cli/internal/tiger/util"
 )
 
@@ -44,10 +45,18 @@ type TigerMCPServer struct {
 type clientConfig struct {
 	ClientType           MCPClient // Our internal client type
 	Name                 string
-	EditorNames          []string // Supported client names for this client
-	MCPServersPathPrefix string   // JSON path prefix for MCP servers config (only for JSON config manipulation clients like Cursor/Windsurf)
-	ConfigPaths          []string // Config file locations - used for backup on all clients, and for JSON manipulation on JSON-config clients
-	InstallCommand       []string // CLI command to install MCP server (for CLI-based installation clients like Claude Code/Codex/Gemini/VSCode)
+	EditorNames          []string                        // Supported client names for this client
+	MCPServersPathPrefix string                          // JSON path prefix for MCP servers config (only for JSON config manipulation clients like Cursor/Windsurf)
+	ConfigPaths          []string                        // Config file locations - used for backup on all clients, and for JSON manipulation on JSON-config clients
+	buildInstallCommand  func(tigerPath string) []string // Function to build CLI install command (for CLI-based installation clients)
+}
+
+// BuildInstallCommand constructs the install command with the given Tiger binary path
+func (c *clientConfig) BuildInstallCommand(tigerPath string) []string {
+	if c.buildInstallCommand == nil {
+		return nil
+	}
+	return c.buildInstallCommand(tigerPath)
 }
 
 // supportedClients defines the clients we support for Tiger MCP installation
@@ -62,7 +71,9 @@ var supportedClients = []clientConfig{
 		ConfigPaths: []string{
 			"~/.claude.json",
 		},
-		InstallCommand: []string{"claude", "mcp", "add", "-s", "user", "tigerdata", "tiger", "mcp", "start"},
+		buildInstallCommand: func(tigerPath string) []string {
+			return []string{"claude", "mcp", "add", "-s", "user", mcp.ServerName, tigerPath, "mcp", "start"}
+		},
 	},
 	{
 		ClientType:           Cursor,
@@ -90,7 +101,9 @@ var supportedClients = []clientConfig{
 			"~/.codex/config.toml",
 			"$CODEX_HOME/config.toml",
 		},
-		InstallCommand: []string{"codex", "mcp", "add", "tigerdata", "tiger", "mcp", "start"},
+		buildInstallCommand: func(tigerPath string) []string {
+			return []string{"codex", "mcp", "add", mcp.ServerName, tigerPath, "mcp", "start"}
+		},
 	},
 	{
 		ClientType:  Gemini,
@@ -99,7 +112,9 @@ var supportedClients = []clientConfig{
 		ConfigPaths: []string{
 			"~/.gemini/settings.json",
 		},
-		InstallCommand: []string{"gemini", "mcp", "add", "-s", "user", "tigerdata", "tiger", "mcp", "start"},
+		buildInstallCommand: func(tigerPath string) []string {
+			return []string{"gemini", "mcp", "add", "-s", "user", mcp.ServerName, tigerPath, "mcp", "start"}
+		},
 	},
 	{
 		ClientType:  VSCode,
@@ -110,7 +125,9 @@ var supportedClients = []clientConfig{
 			"~/Library/Application Support/Code/User/mcp.json",
 			"~/AppData/Roaming/Code/User/mcp.json",
 		},
-		InstallCommand: []string{"code", "--add-mcp", `{"name":"tigerdata","command":"tiger","args":["mcp","start"]}`},
+		buildInstallCommand: func(tigerPath string) []string {
+			return []string{"code", "--add-mcp", fmt.Sprintf(`{"name":"%s","command":"%s","args":["mcp","start"]}`, mcp.ServerName, tigerPath)}
+		},
 	},
 }
 
@@ -143,11 +160,11 @@ func installMCPForClient(clientName string, createBackup bool, customConfigPath 
 		if err != nil {
 			return fmt.Errorf("failed to find configuration for %s: %w", clientName, err)
 		}
-	} else if len(clientCfg.InstallCommand) == 0 {
-		// Client has neither ConfigPaths nor InstallCommand
-		return fmt.Errorf("client %s has no ConfigPaths or InstallCommand defined", clientName)
+	} else if clientCfg.buildInstallCommand == nil {
+		// Client has neither ConfigPaths nor buildInstallCommand
+		return fmt.Errorf("client %s has no ConfigPaths or buildInstallCommand defined", clientName)
 	}
-	// else: CLI-only client - configPath remains empty, will use InstallCommand
+	// else: CLI-only client - configPath remains empty, will use buildInstallCommand
 
 	logging.Info("Installing Tiger MCP server configuration",
 		zap.String("client", clientName),
@@ -167,8 +184,8 @@ func installMCPForClient(clientName string, createBackup bool, customConfigPath 
 	}
 
 	// Add Tiger MCP server to configuration
-	if len(clientCfg.InstallCommand) > 0 {
-		// Use CLI approach when install command is configured
+	if clientCfg.buildInstallCommand != nil {
+		// Use CLI approach when install command builder is configured
 		if err := addTigerMCPServerViaCLI(clientCfg); err != nil {
 			return fmt.Errorf("failed to add Tiger MCP server configuration: %w", err)
 		}
@@ -192,7 +209,7 @@ func installMCPForClient(clientName string, createBackup bool, customConfigPath 
 
 	fmt.Printf("\n💡 Next steps:\n")
 	fmt.Printf("   1. Restart %s to load the new configuration\n", clientName)
-	fmt.Printf("   2. The TigerData MCP server will be available as 'tigerdata'\n")
+	fmt.Printf("   2. The TigerData MCP server will be available as '%s'\n", mcp.ServerName)
 	fmt.Printf("\n🤖 Try asking your AI assistant:\n")
 	fmt.Printf("\n   📊 List and manage your TigerData services:\n")
 	fmt.Printf("   • \"List my TigerData services\"\n")
@@ -428,7 +445,7 @@ func (m clientSelectModel) View() string {
 
 // addTigerMCPServerViaCLI adds Tiger MCP server using a CLI command configured in clientConfig
 func addTigerMCPServerViaCLI(clientCfg *clientConfig) error {
-	if len(clientCfg.InstallCommand) == 0 {
+	if clientCfg.buildInstallCommand == nil {
 		return fmt.Errorf("no install command configured for client %s", clientCfg.Name)
 	}
 
@@ -438,17 +455,8 @@ func addTigerMCPServerViaCLI(clientCfg *clientConfig) error {
 		return fmt.Errorf("failed to get Tiger executable path: %w", err)
 	}
 
-	// Build command with full Tiger path replacing "tiger" placeholder
-	installCommand := make([]string, len(clientCfg.InstallCommand))
-	copy(installCommand, clientCfg.InstallCommand)
-	for i, arg := range installCommand {
-		if arg == "tiger" {
-			installCommand[i] = tigerPath
-		} else if strings.Contains(arg, `"command":"tiger"`) {
-			// Handle JSON format for VS Code: replace "tiger" in JSON string
-			installCommand[i] = strings.Replace(arg, `"command":"tiger"`, fmt.Sprintf(`"command":"%s"`, tigerPath), 1)
-		}
-	}
+	// Build the install command with the full Tiger path
+	installCommand := clientCfg.BuildInstallCommand(tigerPath)
 
 	logging.Info("Adding Tiger MCP server using CLI",
 		zap.String("client", clientCfg.Name),
@@ -566,7 +574,7 @@ func addTigerMCPServerViaJSON(configPath string, mcpServersPathPrefix string) er
 	}
 
 	// Create JSON patch to add the Tiger MCP server
-	patch := fmt.Sprintf(`[{ "op": "add", "path": "%s/tigerdata", "value": %s }]`, mcpServersPathPrefix, dataJSON)
+	patch := fmt.Sprintf(`[{ "op": "add", "path": "%s/%s", "value": %s }]`, mcpServersPathPrefix, mcp.ServerName, dataJSON)
 
 	// Apply the patch
 	if err := value.Patch([]byte(patch)); err != nil {
@@ -585,7 +593,7 @@ func addTigerMCPServerViaJSON(configPath string, mcpServersPathPrefix string) er
 	}
 
 	logging.Info("Added Tiger MCP server to configuration",
-		zap.String("server_name", "tigerdata"),
+		zap.String("server_name", mcp.ServerName),
 		zap.String("command", tigerServer.Command),
 		zap.Strings("args", tigerServer.Args),
 	)
--- a/specs/spec_mcp.md
+++ b/specs/spec_mcp.md
@@ -45,7 +45,7 @@ Alternatively, for manual configuration, the Tiger MCP server can be added to yo
 ```json
 {
   "mcpServers": {
-    "tigerdata": {
+    "tiger": {
       "command": "tiger",
       "args": ["mcp", "start"]
     }
@@ -109,7 +109,7 @@ The command adds the Tiger MCP server configuration using the appropriate format
 ```json
 {
   "mcpServers": {
-    "tigerdata": {
+    "tiger": {
       "command": "tiger",
       "args": ["mcp", "start"]
     }
--- a/internal/tiger/cmd/mcp_install_test.go
+++ b/internal/tiger/cmd/mcp_install_test.go
@@ -149,7 +149,7 @@ func TestAddTigerMCPServer(t *testing.T) {
 			mcpServersPathPrefix: "/mcpServers",
 			expectedResult: map[string]interface{}{
 				"mcpServers": map[string]interface{}{
-					"tigerdata": map[string]interface{}{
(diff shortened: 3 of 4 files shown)


internal/tiger/cmd/mcp.go
```diff
@@ -41,11 +41,81 @@ Use 'tiger mcp start' to launch the MCP server.`,
...
+
+  # Install for Cursor IDE
+  tiger mcp install cursor
+
+  # Install without creating backup
+  tiger mcp install claude-code --no-backup
+
+  # Use custom configuration file path
+  tiger mcp install claude-code --config-path ~/custom/config.json`, generateSupportedEditorsHelp()),
+		Args: cobra.MaximumNArgs(1),
```
nathanjcochran: There is also a `ValidArgs` field on the `cobra.Command` type (see [docs](https://pkg.go.dev/github.com/spf13/cobra#Command)), which lets you list the valid arguments. I think it improves shell completions, and can also be combined with [OnlyValidArgs](https://pkg.go.dev/github.com/spf13/cobra#OnlyValidArgs) to make cobra validate the arguments itself.

Just a suggestion.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+}
+
+// supportedClients defines the clients we support for Tiger MCP installation
+// Note: MCPServersPathPrefix can be found in the supportedClientIntegrations map found in:
+// https://github.com/stacklok/toolhive/blob/main/pkg/client/config.go
+var supportedClients = []clientConfig{
+	{
+		ClientType:           ClaudeCode,
+		Name:                 "Claude Code",
+		EditorNames:          []string{"claude-code", "claude"},
```
nathanjcochran: Do Claude Code and Claude Desktop use the same configuration file under the hood? If not, I'd personally expect `claude` to refer to Claude Desktop, and `claude-code` to refer to  Claude Code.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+		MCPServersPathPrefix: "", // Not used for CLI-based installation
+		ConfigPaths: []string{
+			"~/.claude.json", // Default Claude Code config location - needed for backup
+		},
+		InstallCommand: []string{"claude", "mcp", "add", "tigerdata", "tiger", "mcp", "start"},
```
nathanjcochran: I think this installs the MCP server in [Local scope](https://docs.claude.com/en/docs/claude-code/mcp#mcp-installation-scopes) by default. That's _probably_ what we want, but worth noting in any case.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+		},
+		InstallCommand: []string{"gemini", "mcp", "add", "-s", "user", "tigerdata", "tiger", "mcp", "start"},
+	},
+	{
+		ClientType:           VSCode,
+		Name:                 "VS Code",
+		EditorNames:          []string{"vscode", "code", "vs-code"},
+		MCPServersPathPrefix: "", // Not used for VS Code - uses CLI
+		ConfigPaths:          []string{
+			// VS Code doesn't need config paths for backup - it manages its own config
```
nathanjcochran: Is this true? These [docs](https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_add-an-mcp-server) seem to suggest that there's an `mcp.json` file involved (though I only scanned them briefly, so not entirely sure).

internal/tiger/cmd/mcp.go
```diff
@@ -41,11 +41,81 @@ Use 'tiger mcp start' to launch the MCP server.`,
...
 	return cmd
 }
 
+// buildMCPInstallCmd creates the install subcommand for configuring editors
+func buildMCPInstallCmd() *cobra.Command {
+	var noBackup bool
+	var configPath string
+
+	cmd := &cobra.Command{
+		Use:   "install [editor]",
```
nathanjcochran: I think this should be `[client]`, not `[editor]`, since not all MCP clients are code editors. For example, I could see us supporting Claude Desktop in the future.

Same for other copy that currently uses the word "editor". Just my 2 cents.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	tigerClientType, err := mapEditorToTigerClientType(editorName)
+	if err != nil {
+		return err
+	}
+
+	// Get the MCP servers path prefix from our own configuration
+	clientCfg, err := findOurClientConfig(tigerClientType)
+	if err != nil {
+		return err
+	}
```
nathanjcochran: Both of these functions loop over the same list of `supportedClients`. I think this could probably be condensed into a single function.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	"github.com/tidwall/gjson"
+	"go.uber.org/zap"
+
+	"github.com/timescale/tiger-cli/internal/tiger/logging"
+)
+
+// lockTimeout is the maximum time to wait for a file lock
+const lockTimeout = 1 * time.Second
+
+// MCPClient represents our internal client types (extends beyond toolhive support)
```
nathanjcochran: What is `toolhive`? There are a lot of references to it in the comments, but it doesn't look like the dependency is actually used, except in the tests (might be nice to get rid of it entirely, if we can?)

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+		configPath = customConfigPath
+	} else if len(clientCfg.ConfigPaths) > 0 {
+		// Use manual config path discovery for clients with configured paths
+		configPath, err = findClientConfigFile(clientCfg.ConfigPaths)
+		if err != nil {
+			return fmt.Errorf("failed to find configuration for %s: %w", editorName, err)
+		}
+	} else if len(clientCfg.InstallCommand) > 0 {
+		// CLI-only client - no config path needed, will use InstallCommand
+		configPath = "" // Will be set appropriately in success message
```
nathanjcochran: I think this is a no-op? Can probably re-arrange the conditionals to get rid of this case entirely (i.e. convert the `else` clause below to `else if len(clientCfg.InstallCommand) == 0)`.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+
+// generateSupportedEditorsHelp generates the supported editors section for help text
+func generateSupportedEditorsHelp() string {
+	result := "Supported Editors:\n"
+	for _, cfg := range supportedClients {
+		// Show all editor names for this client
+		editorNames := strings.Join(cfg.EditorNames, " (or ")
+		if len(cfg.EditorNames) > 1 {
+			editorNames = fmt.Sprintf("%s)", editorNames)
+		}
```
nathanjcochran: This code is not right. It produces output like this if there's more than one alias for an editor:

```
vscode (or code (or vs-code)
```

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+			return expandedPath, nil
+		}
+	}
+
+	// If no existing config found, use the last path as default
+	if len(configPaths) == 0 {
+		return "", fmt.Errorf("no config paths provided")
+	}
+
+	defaultPath := expandPath(configPaths[len(configPaths)-1]) // Use last path as fallback
```
nathanjcochran: Why last path and not first? I'd expect the first in the list to be the default/normal config path.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	// Expand home directory
+	if strings.HasPrefix(expanded, "~/") {
+		homeDir, err := os.UserHomeDir()
+		if err == nil {
+			expanded = filepath.Join(homeDir, expanded[2:])
+		}
+	}
+
+	return expanded
+}
```
nathanjcochran: We already have a similar function [here](https://github.com/timescale/tiger-cli/blob/73901d3a1643bc3163632f00fa843b647ddcc23e/internal/tiger/config/config.go#L258-L270). Might be worth consolidating into the `utils` package.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+		zap.String("client", clientCfg.Name),
+		zap.String("output", string(output)))
+
+	return nil
+}
+
+// createConfigBackup creates a backup of the existing configuration file and returns the backup path
+func createConfigBackup(configPath string) (string, error) {
+	// Check if config file exists
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
```
nathanjcochran: Docs for [os.IsNotExist](https://pkg.go.dev/os#IsNotExist) say:

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use `errors.Is(err, fs.ErrNotExist)`.

Might be worth switching to `errors.Is(err, fs.ErrNotExist)` here.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+			logging.Error("Failed to release file lock", zap.Error(err))
+		}
+	}()
+
+	// Read existing configuration or create empty one
+	content, err := os.ReadFile(configPath)
+	if err != nil {
+		logging.Info("Config file not found, creating new one")
+		content = []byte("{}")
+	}
```
nathanjcochran, line 547: Might be good to check for [fs.ErrNotExist](https://pkg.go.dev/fs#ErrNotExist) here before continuing with `{}` as the file content. If the error was due to something else besides the file not existing, I don't think we'd want to overwrite it (even if we made a backup that the user could theoretically revert to).

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+
+	if len(content) == 0 {
+		// If the file is empty, initialize with empty JSON object
+		content = []byte("{}")
+	}
+
+	// Parse the JSON with hujson
+	value, err := hujson.Parse(content)
```
nathanjcochran, line 555: Do we really need the `hujson` library? Seems like all it's really doing is inserting a JSON object into an array at a specified JSON path. Could probably done pretty easily just by unmarshalling to `any` and traversing down to the correct place in the object. Especially since we seem to only support single-level paths anyways (per the `ensurePathExists` function below).

Just trying to keep our list of dependencies as small as possible.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	}
+
+	// Format the result
+	formatted, err := hujson.Format(value.Pack())
+	if err != nil {
+		return fmt.Errorf("failed to format patched JSON: %w", err)
+	}
+
+	// Write back to file atomically
+	if err := renameio.WriteFile(configPath, formatted, 0644); err != nil {
```
nathanjcochran: Do we really need `renameio`? We aren't going out of our way to atomically write files in other places (e.g. the config file or the fallback API key file). Why do we need to here?

Also, the [README](https://github.com/google/renameio?tab=readme-ov-file#windows-support) says it doesn't work on Windows, so I think this will probably break the Windows build:

> It is https://github.com/golang/go/issues/22397#issuecomment-498856679, and https://github.com/google/renameio/issues/17.
>
> As it is not possible to provide a correct implementation, this package does not export any functions on Windows.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	// Validate that this is a single-level path
+	key := strings.TrimPrefix(path, "/")
+	if strings.Contains(key, "/") {
+		return fmt.Errorf("nested paths are not supported, got: %s", path)
+	}
+
+	// Check if the key already exists using gjson
+	// For keys with dots (like "amp.mcpServers"), we need to escape the dots for gjson
+	escapedKey := strings.ReplaceAll(key, ".", `\.`)
+	if gjson.GetBytes(content, escapedKey).Exists() {
```
nathanjcochran: `gjson` is another new dependency it would be nice to eliminate imo, if possible.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+
+	// Tiger MCP server configuration
+	tigerServer := TigerMCPServer{
+		Command: tigerPath,
+		Args:    []string{"mcp", "start"},
+	}
+
+	// Create a lock file
+	lockPath := configPath + ".lock"
+	fileLock := flock.New(lockPath)
```
nathanjcochran: Similarly, do we really need `flock` here? If I'm understanding correctly, this kind of file locking would only prevent concurrent instances of Tiger CLI from modifying the file at the same time - it doesn't prevent races with other programs, since they don't know to use this lock file.

Are we really worried about race conditions arising from concurrent `tiger mcp install` commands? Seems like something that users just shouldn't (and almost certainly won't) do. Simplifying the code and eliminating a dependency is probably more beneficial, imo.

I also worry a bit about the lock file not being cleaned up properly via the `defer` call if the program exits unexpectedly, which could cause future invocations of the `tiger mcp install` command to fail. The [docs](https://pkg.go.dev/github.com/gofrs/flock#pkg-overview) also have a warning about operating-system specified behaviors:

> While using this library, remember that the locking behaviors are not guaranteed to be the same on each platform.

Feels like this might just cause more potential issues/complexities than it's solving.

specs/spec_mcp.md
```diff
@@ -57,11 +57,59 @@ The MCP server will automatically use the CLI's stored authentication and config
 
 ### CLI MCP Commands
 
+#### `tiger mcp install <editor>`
+Install and configure the Tiger MCP server for a specific editor or AI assistant. This command automates the configuration process by modifying the appropriate configuration files.
+
+**Supported Editors:**
+- `claude-code` (or `claude`): Configure for Claude Code
+- `cursor`: Configure for Cursor IDE
```
nathanjcochran, line 65: Should we list all of the supported editors here? The code supports more than these two.

Not sure how important it is to you to keep these specs completely accurate and up-to-date.

internal/tiger/cmd/mcp_install.go
nathanjcochran: Is `tigerdata` really the best name for the MCP server?

I have been using `tiger-mcp` locally, fwiw. That's also the MCP server's current "name" in the code (see [here](https://github.com/timescale/tiger-cli/blob/73901d3a1643bc3163632f00fa843b647ddcc23e/internal/tiger/mcp/server.go#L17) and [here](https://github.com/timescale/tiger-cli/blob/73901d3a1643bc3163632f00fa843b647ddcc23e/internal/tiger/mcp/server.go#L31)). I personally think we should reuse that `serverName` constant in any case (but I'm totally open to changing it to something else, if you have a better idea for a name).

internal/tiger/util/password_storage.go
```diff
@@ -1,4 +1,4 @@
-package util
```
cevian, line 1: we needed to move this to resolve circular dependency issues in util. Util should never really import other packages and it was importing config and api before this change

internal/tiger/util/convert_test.go
```diff
@@ -2,10 +2,12 @@ package util
 
 import (
 	"testing"
-
-	"github.com/timescale/tiger-cli/internal/tiger/api"
 )
 
+// Define local types for testing to avoid import cycles
```
cevian, line 7: also done to resolve circular dependencies on api

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,658 @@
...
+	// Get the full path of the currently executing Tiger binary
+	tigerPath, err := getTigerExecutablePath()
+	if err != nil {
+		return fmt.Errorf("failed to get Tiger executable path: %w", err)
+	}
+
+	// Tiger MCP server configuration
+	tigerServer := TigerMCPServer{
+		Command: tigerPath,
+		Args:    []string{"mcp", "start"},
```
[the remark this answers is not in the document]
cevian: not implemented

internal/tiger/cmd/mcp_install.go
cevian: changed it to claude-code only

internal/tiger/cmd/mcp_install.go
cevian: good catch changed to user scope which I think is more what we want.

internal/tiger/cmd/mcp_install.go
cevian: removed a bunch of the comments

internal/tiger/cmd/mcp_install.go
cevian: yup did this. This also made me have to resolve a bunch of circular dependencies around utils but I think thats for the best

internal/tiger/cmd/mcp_install.go
cevian: I prefer to keep husjson as it minimizes changes to the file (for instance it preserves comments and is better at keeping formatting)

internal/tiger/cmd/mcp_install.go
cevian: left a question in slack

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,584 @@
...
+		return fmt.Errorf("failed to marshal Tiger MCP server data: %w", err)
+	}
+
+	// Create JSON patch to add the Tiger MCP server
+	patch := fmt.Sprintf(`[{ "op": "add", "path": "%s/tigerdata", "value": %s }]`, mcpServersPathPrefix, dataJSON)
+
+	// Apply the patch
+	if err := value.Patch([]byte(patch)); err != nil {
+		return fmt.Errorf("failed to apply JSON patch: %w", err)
+	}
```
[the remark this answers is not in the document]
cevian: tested and works without changes

internal/tiger/cmd/mcp_install.go
```diff
@@ -219,48 +214,36 @@ func installMCPForEditor(editorName string, createBackup bool, customConfigPath
...
 			}
 		}
 	}
 
-	// Build list of supported editors from our config
+	// Build list of supported clients from our config for error message
 	var supportedNames []string
 	for _, cfg := range supportedClients {
 		supportedNames = append(supportedNames, cfg.EditorNames...)
 	}
```
nathanjcochran: I think you could use the new `getValidEditorNames()` function here. Seems to do the same thing as this code.

internal/tiger/util/path.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+		return homeDir
+	}
+
+	if strings.HasPrefix(expanded, "~/") {
+		homeDir, err := os.UserHomeDir()
+		if err == nil {
+			expanded = filepath.Join(homeDir, expanded[2:])
+		}
+	}
```
nathanjcochran, line 38: The behavior for what happens when `os.UserHomeDir()` returns an error is different between these two cases. In the first, we return an empty string instead of `~`. In the latter, it looks like we return the original path, including the `~/` prefix. I think we should unify the behavior when `os.UserHomeDir()` returns an error - e.g. return the original path, including `~`, in both cases. Alternatively, this function could just return the error.

internal/tiger/cmd/mcp_install.go
```diff
@@ -0,0 +1,584 @@
...
+	backupPath := fmt.Sprintf("%s.backup.%d", configPath, time.Now().Unix())
+
+	// Read original file
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read original config file: %w", err)
+	}
+
+	// Write backup file
+	if err := os.WriteFile(backupPath, data, 0644); err != nil {
```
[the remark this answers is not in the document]
nathanjcochran: This seems like it might be a good improvement (trying to use the existing file's permissions).

PR #30, opened 2025-09-30
feat: comprehensive MCP service tool improvements

cevian:
## Summary
Comprehensive improvements to MCP service tools including renamed tools, optimized descriptions, improved parameter clarity, and extensive documentation.

### Tool Naming Improvements
- Removed redundant `tiger_` prefix from all MCP tools (e.g., `tiger_service_list` → `service_list`)
- Simplified tool names for better usability

### Description Optimization
- Streamlined tool descriptions for LLM token efficiency
- Removed redundant information while maintaining clarity
- Focused descriptions on essential functionality

### Parameter Improvements
- Renamed `timeout` to `timeout_minutes` in `service_create` for explicit unit clarity
- Improved parameter documentation with better examples and constraints
- Enhanced schema generation with proper type definitions

### Documentation
- Added comprehensive MCP feedback document (`mcp_feedback.md`)
- Analyzed tool naming conventions, parameter documentation, and API design
- Documented implementation status and future improvements

## Commits Included
- `d60341f` fix: rename MCP tools to remove redundant tiger_ prefix
- `3d22a25` feat: implement comprehensive MCP tool improvements  
- `c58b29c` docs: finalize MCP feedback implementation status
- `62e3543` docs: add MCP tool description optimization analysis
- `880f2e8` docs: update MCP feedback with implementation status
- `17f087c` refactor: streamline MCP tool descriptions for LLM optimization
- `a28eba9` refactor: rename timeout to timeout_minutes in service_create MCP tool

## Test Plan
- [x] Code compiles successfully
- [x] MCP tools function correctly with new names
- [x] Schema generation produces valid JSON schemas
- [x] Parameter documentation is clear and accurate


internal/tiger/mcp/service_tools.go
```diff
@@ -16,78 +16,87 @@ import (
...
 
-// ServiceListOutput represents output for tiger_service_list
+// ServiceListOutput represents output for service_list
 type ServiceListOutput struct {
 	Services []ServiceInfo `json:"services"`
 }
 
+func (ServiceListOutput) Schema() *jsonschema.Schema {
+	return util.Must(jsonschema.For[ServiceListOutput](nil))
+}
```
nathanjcochran, line 50: Fwiw, the [mcp.AddTool](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#AddTool) function automatically generates schema for both input types and output types using this exact same logic:

> If the tool's input schema is nil, it is set to the schema inferred from the In type parameter. Types are inferred from Go types, and property descriptions are read from the 'jsonschema' struct tag. Internally, the SDK uses the github.com/google/jsonschema-go package for inference and validation. The In type argument must be a map or a struct, so that its inferred JSON Schema has type "object", as required by the spec. As a special case, if the In type is 'any', the tool's input schema is set to an empty object schema value.
> 
> If the tool's output schema is nil, and the Out type is not 'any', the output schema is set to the schema inferred from the Out type argument, which must also be a map or struct. If the Out type is 'any', the output schema is omitted. 

The only reason I'm doing it manually for the input types (by explicitly setting the `InputSchema` field rather than leaving it `nil`) was so that I could add descriptions, examples, enums, etc. If we aren't adding extra metadata for the output types, I don't think we really _need_ this code (unless we just want to match the pattern for the input types for the sake of code clarity, which I'm also fine with).

internal/tiger/mcp/service_tools.go
```diff
@@ -16,78 +16,87 @@ import (
...
+
 // ServiceInfo represents simplified service information for MCP output
 type ServiceInfo struct {
-	ServiceID string        `json:"id"`
+	ServiceID string        `json:"id" jsonschema:"Service identifier (10-character alphanumeric string)"`
 	Name      string        `json:"name"`
-	Status    string        `json:"status"`
-	Type      string        `json:"type"`
+	Status    string        `json:"status" jsonschema:"Service status (e.g., READY, PAUSED, CONFIGURING, UPGRADING)"`
+	Type      string        `json:"type" jsonschema:"One of: TIMESCALEDB, POSTGRES, VECTOR"`
```
nathanjcochran: It probably doesn't matter as much for output types, but I think using an jsonschema `enum` is generally a better approach for enumerating the possible values of a field (e.g. like we're doing [here](https://github.com/timescale/tiger-cli/blob/a28eba95547e7fa9b6ed6a92c7ce0a13ed1b1856/internal/tiger/mcp/service_tools.go#L110)), rather than putting it in the description.

internal/tiger/mcp/service_tools.go
```diff
@@ -16,78 +16,87 @@ import (
...
+	Type           string        `json:"type" jsonschema:"One of: TIMESCALEDB, POSTGRES, VECTOR"`
 	Region         string        `json:"region"`
 	Created        string        `json:"created,omitempty"`
 	Resources      *ResourceInfo `json:"resources,omitempty"`
-	Replicas       int           `json:"replicas,omitempty"`
-	DirectEndpoint string        `json:"direct_endpoint,omitempty"`
-	PoolerEndpoint string        `json:"pooler_endpoint,omitempty"`
+	Replicas       int           `json:"replicas,omitempty" jsonschema:"Number of HA replicas (0=single node/no HA, 1+=HA enabled)"`
+	DirectEndpoint string        `json:"direct_endpoint,omitempty" jsonschema:"Direct database connection endpoint"`
+	PoolerEndpoint string        `json:"pooler_endpoint,omitempty" jsonschema:"Connection pooler endpoint"`
```
nathanjcochran, line 113: Is there a reason we're providing descriptions for some fields, but not others? Is the assumption that some of the fields are self-explanatory and don't need a description? I would kind of expect all of the fields to have a description if any of them do.

internal/tiger/mcp/service_tools.go
```diff
@@ -144,104 +157,84 @@ type ServiceUpdatePasswordInput struct {
...
+	// service_create
 	mcp.AddTool(s.mcpServer, &mcp.Tool{
-		Name:  "tiger_service_create",
+		Name:  "service_create",
 		Title: "Create Database Service",
-		Description: `Create a new database service in TigerData Cloud.
-
-This tool provisions a new database service with specified configuration including service type, compute resources, region, and high availability options. By default, the tool returns immediately after the creation request is accepted, but the service may still be provisioning and not ready for connections yet.
-
-Only set 'wait: true' if you need the service to be fully ready immediately after the tool call returns. In most cases, leave wait as false (default) for faster responses.
```
nathanjcochran, line 221: Fwiw, I had previously included this because Claude was almost always setting `wait: true`, even when not prompted to, and despite the default being set to `false`. It was just kind of assuming that it would be better to wait for the service to be ready, and didn't seem to fully understand the downsides of pausing the conversation and forcing the user to wait. This message helped it understand that it's often okay to not wait, unless there's a specific reason/need to wait. I didn't try it with your updated descriptions, so it might be fine, but figured I'd mention it since it's something I noticed before.

internal/tiger/mcp/service_tools.go
```diff
@@ -144,104 +157,84 @@ type ServiceUpdatePasswordInput struct {
...
-- Provisioning databases with specific resource requirements
-- Establishing services in different geographical regions`,
-		InputSchema: ServiceCreateInput{}.Schema(),
+		Description: "Create a new database service in TigerData Cloud with specified type, compute resources, region, and HA options.\n" +
+			"\n" +
+			"Default: Returns immediately (service provisions in background).\n" +
+			"wait=true: Blocks until service ready.\n" +
+			"timeout_minutes: Wait duration in minutes (with wait=true).\n" +
+			"\n" +
+			"WARNING: Creates billable resources.",
```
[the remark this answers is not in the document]
nathanjcochran: I agree with this feedback.

docs/mcp_feedback.md
```diff
@@ -0,0 +1,128 @@
...
+**WHY THESE MATTER**: Complete lifecycle management requires all CRUD operations
+
+## Schema Design Observations
+
+### Resource Format Inconsistency
+- **Input**: `"cpu_memory": "0.5 CPU/2GB"` (combined string)
+- **Output**: `{"cpu": "0.5 cores", "memory": "2 GB"}` (separate fields)
+- **Impact**: Requires different parsing logic for input vs output
+- **Recommendation**: Use consistent format or document the rationale
+- **WHY**: Reduces implementation complexity
```
nathanjcochran, line 95: For the record, the rationale for the combined input strings is to prevent the LLM from selecting invalid combinations of CPU/memory in the request. Making it a single field with enum values makes it impossible for the LLM to select invalid combinations.

We could do the same for the output, but we don't have the same problem on the output side (and I'm not actually convinced that consistency between input and output really matters to the LLM)

internal/tiger/mcp/service_tools.go
cevian: We only add descriptions to fields that are non-obvious. Fields like `name`, `region`, `created`, and `paused` are self-explanatory from their names, so adding descriptions would waste tokens without adding value.

internal/tiger/mcp/service_tools.go
cevian: made the language a bit stronger

docs/mcp_feedback.md
cevian: yep agreed

internal/tiger/mcp/service_tools.go
cevian: lets match the pattern

PR #32, opened 2025-10-01
AGE-181: support output formats for `service create`

murrayju:
This adds support for the `--output` flag to the `tiger service create` command, as well as adding a `--with-password` flag, to include the initial password in the output.

image

This also changes the default behavior, to output a table of service information after the status messages. For this, I reused the existing render code used by `tiger service describe`. I modified this to also include the `connection_string` in the output (while respecting the `--with-password` flag).

image

I also changed it so that the status messages are written to `stderr` (instead of `stdout`), and only the important/structured output (json, yaml, or table) is written to `stdout`. This makes it so that the command can be easily piped into other tooling (like `jq`).

image

If you'd rather not see the status messages, you can simply redirect `stderr`

image

### DIFF
--- a/internal/tiger/cmd/db_test.go
+++ b/internal/tiger/cmd/db_test.go
@@ -1056,7 +1056,7 @@ func TestBuildConnectionString_WithPassword_KeyringStorage(t *testing.T) {
 
 	// Store a test password in keyring
 	testPassword := "test-password-keyring-123"
-	storage := util.GetPasswordStorage()
+	storage := password.GetPasswordStorage()
 	err := storage.Save(service, testPassword)
 	if err != nil {
 		t.Fatalf("Failed to save test password: %v", err)
@@ -1107,7 +1107,7 @@ func TestBuildConnectionString_WithPassword_PgpassStorage(t *testing.T) {
 
 	// Store a test password in pgpass
 	testPassword := "test-password-pgpass-456"
-	storage := util.GetPasswordStorage()
+	storage := password.GetPasswordStorage()
 	err := storage.Save(service, testPassword)
 	if err != nil {
 		t.Fatalf("Failed to save test password: %v", err)
@@ -1235,7 +1235,7 @@ func TestDBConnectionString_WithPassword(t *testing.T) {
 
 	// Store a test password
 	testPassword := "test-e2e-password-789"
-	storage := util.GetPasswordStorage()
+	storage := password.GetPasswordStorage()
 	err := storage.Save(service, testPassword)
 	if err != nil {
 		t.Fatalf("Failed to save test password: %v", err)


internal/tiger/cmd/service.go
```diff
@@ -379,10 +385,17 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 			defer cancel()
 
+			// Get output format early to decide where to write messages
+			outputFormat := strings.ToLower(cfg.Output)
+			outputWriter := cmd.OutOrStdout()
+			if outputFormat == "json" || outputFormat == "yaml" {
+				outputWriter = cmd.ErrOrStderr()
+			}
```
murrayju: Another option would be to _always_ write the status messages to stderr. That probably simplifies things a bit, and there's not much impact on the user's experience. Would like to get @nathanjcochran's input on this.

nathanjcochran: Yeah, I vote to be consistent and always output to stderr.

We should probably make the same change for other commands writing to stdout too, but that could be a follow-up (can you make a Linear issue)?

internal/tiger/cmd/service.go
```diff
@@ -411,25 +424,30 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
...
 				// Handle wait behavior
 				if createNoWait {
-					fmt.Fprintf(cmd.OutOrStdout(), "⏳ Service is being created. Use 'tiger service list' to check status.\n")
-					return nil
+					fmt.Fprintf(outputWriter, "⏳ Service is being created. Use 'tiger service list' to check status.\n")
+				} else {
+					// Wait for service to be ready
+					fmt.Fprintf(outputWriter, "⏳ Waiting for service to be ready (wait timeout: %v)...\n", createWaitTimeout)
+					if status, err := waitForServiceReady(client, projectID, serviceID, createWaitTimeout, cmd, outputWriter); err != nil {
+						fmt.Fprintf(outputWriter, "❌ %v\n", err)
```
nathanjcochran: We used to return the error from `waitForServiceReady`, which caused the command to return a non-zero exit code (see [here](https://github.com/timescale/tiger-cli/blob/e5c15520db19ad90d6d4cb4653f48b858e59c9c9/internal/tiger/cmd/root.go#L108-L115)). This is also [documented in the spec](https://github.com/timescale/tiger-cli/blob/e5c15520db19ad90d6d4cb4653f48b858e59c9c9/specs/spec.md?plain=1#L317):

> - **Wait for Completion**: By default, asynchronous service commands (see list above) will wait for the operation to complete before returning, displaying status updates every 10 seconds. Use `--no-wait` to return immediately after the request is accepted, or `--wait-timeout` to specify a custom timeout period. If the wait timeout is exceeded, the command will exit with code 2, but the operation will continue running on the server.

Now we're just logging the error and proceeding to output the service info. That changes the semantics of the command - i.e. with these changes, users won't be able to tell from the exit code whether the service came up successfully or not.

I think we want to maintain the prior semantics with respect to exit codes. Not exactly sure the best way to do that. We could either not output the service info if it doesn't become ready, or we could output the info, but still return the error/exit code from `waitForServiceReady` 🤔.

internal/tiger/cmd/service.go
```diff
@@ -748,26 +797,33 @@ func formatTimePtr(t *time.Time) string {
...
+// statusWriter determines where status messages are written (defaults to cmd.OutOrStdout() if nil)
+func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID string, waitTimeout time.Duration, cmd *cobra.Command, statusWriter io.Writer) (*api.DeployStatus, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), waitTimeout)
 	defer cancel()
 
+	// Determine where to write status messages
+	output := statusWriter
+	if output == nil {
+		output = cmd.OutOrStdout()
+	}
```
nathanjcochran: This pattern, which is present in several of the functions now (i.e. `waitForServiceReady`, `handlePasswordSaving`, `setDefaultService`), is weird to me. I think we should just remove the `cmd` parameter from the function signatures, and have callers pass in `cmd.ErrOrStderr()` as the `statusWriter` instead of passing `nil`. The `cmd` argument doesn't appear to be used for anything else in these functions, besides for getting a default output writer if one isn't provided.

internal/tiger/cmd/service.go
```diff
@@ -587,18 +606,18 @@ Examples:
...
-		return encoder.Encode(sanitizeServiceForOutput(service))
+		return encoder.Encode(prepareServiceForOutput(service, withPassword, cmd))
 	case "yaml":
 		encoder := yaml.NewEncoder(cmd.OutOrStdout())
 		encoder.SetIndent(2)
-		return encoder.Encode(sanitizeServiceForOutput(service))
+		return encoder.Encode(prepareServiceForOutput(service, withPassword, cmd))
 	default: // table format (default)
-		return outputServiceTable(cmd, service)
+		return outputServiceTable(cmd, service, withPassword)
```
nathanjcochran: Rather than passing `withPassword` down into `prepareServiceForOutput` and `outputServiceTable` (where each of those handles conditionally outputting the initial password and calling `buildConnectionString`), I think it might be nice to create a new `OutputService` type (or something like that) which embeds `api.Service` (or just includes the fields we want to output?), but also adds a `ConnectionString` field. We could construct it here at the top of `outputService` (and in the process could `nil` out the `InitialPassword` field if `withPassword` is `false`). Then we could pass that type straight down into `encoder.Encode` and `outputService`, and they could just display what they're given, without as much logic inside.

Just a thought, because I feel like this logic is getting kind of messy. Doesn't have to be done now - could be done in a follow-up later on.

Somewhat related: do we want to output the connection string in the `tiger service list` output now as well? I think `tiger service describe` and `tiger service list` previously returned the same JSON/YAML structures for services, but now `tiger service describe` gets an extra `connection_string` field that doesn't appear in the `tiger service list` output. Probably would be good to keep them consistent, as much as possible

internal/tiger/cmd/service.go
murrayju: Makes sense. I'd say the latter.

internal/tiger/cmd/service.go
murrayju: Looking into this a bit, it seems that a difference with this approach is that the yaml encoder handles `omitempty` quite differently from the json encoder. It seems to output all the `nil` pointer fields as `null`.

```
❯ ./bin/tiger svc list -o yaml     
- service:
    connectionpooler: null
    created: 2025-10-02T16:23:22.161908Z
    endpoint:
      host: a7ecknrg4b.q5pvyszezw.tsdb.cloud.timescale.com
      port: 39669
    forkedfrom: null
    hareplicas: null
    initialpassword: null
    metadata:
      environment: DEV
    name: db-43914
    paused: false
    projectid: q5pvyszezw
    readreplicasets: null
    regioncode: us-east-1
    resources:
      - id: "103153"
        spec:
          cpumillis: null
          memorygbs: null
          volumetype: ""
    serviceid: a7ecknrg4b
    servicetype: TIMESCALEDB
    status: READY
    vpcendpoint: null
  connection_string: postgresql://tsdbadmin@a7ecknrg4b.q5pvyszezw.tsdb.cloud.timescale.com:39669/tsdb?sslmode=require
```

This wasn't a problem previously, because marshalling to a `map[string]interface{}` (and deleting some keys) would omit them.

Claude suggested doing a json marshal/unmarshal first, and pass that to the yaml encoder. That looks better to me:

```
❯ ./bin/tiger svc list -o yaml     
- connection_string: postgresql://tsdbadmin@a7ecknrg4b.q5pvyszezw.tsdb.cloud.timescale.com:39669/tsdb?sslmode=require
  created: "2025-10-02T16:23:22.161908Z"
  endpoint:
    host: a7ecknrg4b.q5pvyszezw.tsdb.cloud.timescale.com
    port: 39669
  metadata:
    environment: DEV
  name: db-43914
  paused: false
  project_id: q5pvyszezw
  region_code: us-east-1
  resources:
    - id: "103153"
      spec:
        volume_type: ""
  service_id: a7ecknrg4b
  service_type: TIMESCALEDB
  status: READY
  vpcEndpoint: null
```

Ends up a big uglier in the code, though.

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/938505fe27270a11893e63a6402eb4ffb5193c55

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/18c47b042b7d584c5d95e92410bad31d614a1b6e

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/18c47b042b7d584c5d95e92410bad31d614a1b6e

internal/tiger/cmd/service.go
nathanjcochran: > Claude suggested doing a json marshal/unmarshal first, and pass that to the yaml encoder.

Yeah, that's more-or-less what we were doing previously. I guess that's the reason why.

We could probably do it by adding `yaml: "...,omitempty"` struct tags in addition to `json` struct tags. See [these docs](https://pkg.go.dev/gopkg.in/yaml.v3#Marshal). However, that's not possible to achieve as long as we're marshaling an `api.Service`, since that type is auto-generated. If we created an `OutputService` type that did _not_ embed `api.Service`, we could do it. Not sure if that's worth it or not though. On the one hand, it might be nice to separate our API types from our output types. But on the other hand, if they're always 1:1, it's just wasted effort to convert from one to the other 🤷‍♂️.

murrayju: I think this is an improvement over what was there: https://github.com/timescale/tiger-cli/pull/32/commits/de7f833f7f0d97c6c63a4a83d27f618fc2a0343d

Could always change this later.

cevian: I am fine with this `Claude suggested doing a json marshal/unmarshal first, and pass that to the yaml encoder. ` sounds like the easiest way forward. Lets just encapsulate in a util function

murrayju: Yeah, there is a util function in https://github.com/timescale/tiger-cli/pull/32/commits/de7f833f7f0d97c6c63a4a83d27f618fc2a0343d. My focus here is on the service commands, though. Probably worth a follow-up effort to make this more broadly consistent.

internal/tiger/cmd/db.go
```diff
@@ -244,8 +269,8 @@ func buildConnectionString(service api.Service, pooled bool, role string, withPa
 		endpoint = service.ConnectionPooler.Endpoint
 	} else {
 		// If pooled was requested but no pooler is available, warn the user
-		if pooled {
-			fmt.Fprintf(cmd.ErrOrStderr(), "⚠️  Warning: Connection pooler not available for this service, using direct connection\n")
+		if pooled && output != nil {
```
nathanjcochran: I think it's probably fine to assume that `output` is not `nil`, and let the code panic if it is (fairly common in Go when dealing with pointer types). Unless you actually have a use case in mind for passing `nil` in?

internal/tiger/cmd/service.go
```diff
@@ -411,25 +426,36 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
...
+					fmt.Fprintf(statusOutput, "⏳ Service is being created. Use 'tiger service list' to check status.\n")
+				} else {
+					// Wait for service to be ready
+					fmt.Fprintf(statusOutput, "⏳ Waiting for service to be ready (wait timeout: %v)...\n", createWaitTimeout)
+					status, err := waitForServiceReady(client, projectID, serviceID, createWaitTimeout, statusOutput)
+					if err != nil {
+						fmt.Fprintf(statusOutput, "❌ %v\n", err)
+					}
+					result = err
+					service.Status = status
```
nathanjcochran:
```suggestion
					service.Status, result = waitForServiceReady(client, projectID, serviceID, createWaitTimeout, statusOutput)
					if result != nil {
						fmt.Fprintf(statusOutput, "❌ %v\n", result)
					}
```

internal/tiger/cmd/service.go
```diff
@@ -586,43 +615,91 @@ Examples:
...
+	if err != nil {
+		return nil, err
+	}
+
+	var jsonArray []map[string]interface{}
+	if err := json.Unmarshal(jsonBytes, &jsonArray); err != nil {
+		return nil, err
+	}
+	return jsonArray, nil
+}
```
nathanjcochran: I think these could be a single generic function instead:

```golang
func toJSON[T any](v any) (T, error) {
	var jsonOut T

	jsonBytes, err := json.Marshal(v)
	if err != nil {
		return jsonOut, err
	}

	if err := json.Unmarshal(jsonBytes, &jsonOut); err != nil {
		return jsonOut, err
	}
	return jsonOut, nil
}
```

Alternatively, if you don't actually care what the output type is, and you just want to round-trip it to/from JSON (which I think is the case?), you could just do this:

```golang
func toJSON(v any) (any, error) {
	jsonBytes, err := json.Marshal(v)
	if err != nil {
		return nil, err
	}

	var jsonOut any
	if err := json.Unmarshal(jsonBytes, &jsonOut); err != nil {
		return nil, err
	}
	return jsonOut, nil
}
```

Side note: I think it's slightly more idiomatic to use `any` than `interface{}` these days (they're the same, `any` is just a newer/clearer alias for `interface{}`)

internal/tiger/cmd/service.go
```diff
@@ -586,43 +615,91 @@ Examples:
...
 	}
 }
 
 // outputServiceTable outputs detailed service information in a formatted table
-func outputServiceTable(cmd *cobra.Command, service api.Service) error {
+func outputServiceTable(cmd *cobra.Command, outputSvc OutputService) error {
 	table := tablewriter.NewWriter(cmd.OutOrStdout())
 	table.Header("PROPERTY", "VALUE")
 
+	service := outputSvc.Service
```
nathanjcochran: I don't think you need to do this, since the `api.Service` is embedded in the `OutputService` - i.e. you should be able to access any of the `api.Service` fields directly on the `outputSvc` variable (which I would probably just name `service`).

internal/tiger/cmd/service.go
```diff
@@ -694,13 +771,23 @@ func outputServiceTable(cmd *cobra.Command, service api.Service) error {
 		table.Append("Created", service.Created.Format("2006-01-02 15:04:05 MST"))
 	}
 
+	// Output password if available
+	if service.InitialPassword != nil {
+		table.Append("Initial Password", *service.InitialPassword)
```
nathanjcochran, line 763: I think this output is also used for the `tiger service describe` command, right? Should it really always say "_Initial_ Password"? It might have been reset via `tiger service update-password`, for instance, so it might not actually be the "initial" password. Same for the JSON/YAML field (it's also currently `initial_password`).

Maybe we should add a `Password`/`password` field to `OutputService` instead?

internal/tiger/cmd/service.go
```diff
@@ -748,26 +847,30 @@ func formatTimePtr(t *time.Time) string {
...
 
 // waitForServiceReady polls the service status until it's ready or timeout occurs
-func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID string, waitTimeout time.Duration, cmd *cobra.Command) error {
+func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID string, waitTimeout time.Duration, output io.Writer) (*api.DeployStatus, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), waitTimeout)
 	defer cancel()
 
+	if output == nil {
+		return nil, fmt.Errorf("output cannot be nil")
+	}
```
nathanjcochran: Again, I wouldn't worry too much about guards like this, personally. This function doesn't check that the passed-in `client` isn't `nil`, for example 🤷‍♂️.

Probably reasonable for this code to panic if someone passes in a `nil` client or output, since that would just be a straight-up programmer mistake.

Just my two cents (it's admittedly a stylistic choice).

internal/tiger/cmd/service.go
```diff
@@ -748,26 +847,30 @@ func formatTimePtr(t *time.Time) string {
...
+
 	ticker := time.NewTicker(10 * time.Second)
 	defer ticker.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
-			return exitWithCode(ExitTimeout, fmt.Errorf("❌ wait timeout reached after %v - service may still be provisioning", waitTimeout))
+			return nil, exitWithCode(ExitTimeout, fmt.Errorf("❌ wait timeout reached after %v - service may still be provisioning", waitTimeout))
```
nathanjcochran: We might also want to return the last status that was polled, here. That way the output will include the more recent/up-to-date status even if we time out while waiting.

internal/tiger/cmd/service.go
murrayju: I got here because of how there's no way to create one and just assign the other, where at one step in the refactor the error was temporary/local.

internal/tiger/cmd/service.go
murrayju: Wonderful. I was under the impression that the type passed to Unmarshal would matter.

internal/tiger/cmd/service.go
murrayju: I did find this confusing/inconsistent. Thanks Claude.

internal/tiger/cmd/service.go
nathanjcochran: > there's no way to create one and just assign the other

Hmm, fwiw, I think you _can_ typically do this, though the behavior is subtle and there might be edge cases I'm not thinking of. See here for example: https://go.dev/play/p/6r_qto0KLjS

This is covered in the [Short variable declarations](https://go.dev/ref/spec#Short_variable_declarations) section of the Go spec. Particularly this part:

> Unlike regular variable declarations, a short variable declaration may redeclare variables provided they were originally declared earlier in the same block (or the parameter lists if the block is the function body) with the same type, and at least one of the non-[blank](https://go.dev/ref/spec#Blank_identifier) variables is new. As a consequence, redeclaration can only appear in a multi-variable short declaration. Redeclaration does not introduce a new variable; it just assigns a new value to the original. The non-blank variable names on the left side of := must be [unique](https://go.dev/ref/spec#Uniqueness_of_identifiers).

internal/tiger/cmd/service.go
nathanjcochran: It typically does, but there's special rules for `any`/`interface{}`, per the [json.Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) docs:

> To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
> - bool, for JSON booleans
> - float64, for JSON numbers
> - string, for JSON strings
> - []any, for JSON arrays
> - map[string]any, for JSON objects
> - nil for JSON null

internal/tiger/cmd/service.go
murrayju: I personally don't like that it's ever called "initial". But for now, it'll be `initial_password` in the `create` output, and just `password` for `describe`.

internal/tiger/cmd/service.go
murrayju: Go is at this weird intersection between Typescript and C. TS would force you to do this (or otherwise validate in the type system). In C, it used to be best practice to aggressively guard against segfaults. I just have to get used to this idea...

internal/tiger/cmd/service.go
nathanjcochran: > I personally don't like that it's ever called "initial"

Fwiw, I would also support making it always be `password` (no "initial"). Your call though.

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/81a0bfb62b67dec275e3e06eb1cfc2a24c076d8f

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/81a0bfb62b67dec275e3e06eb1cfc2a24c076d8f

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/4fd80dcd6f4ef2a584776456ba49ba7edecea1b3

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/dcdb2683325026b251478906f09bf0b3efffdd32

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/dcdb2683325026b251478906f09bf0b3efffdd32

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/dcdb2683325026b251478906f09bf0b3efffdd32

internal/tiger/cmd/db.go
murrayju: https://github.com/timescale/tiger-cli/pull/32/commits/23b210fd6980616c556c71d6d6a75f3a0b1abb6b

internal/tiger/cmd/service.go
nathanjcochran: Yeah, for sure. And I'm not saying you should _never_ guard against `nil` values, just that it's a matter of discretion. If an argument is a pointer or interface type, but you never actually expect it to be `nil`, it's often fair to make that the (often implicit) contract, and just panic if it's violated.

I think it comes down to the fact that pointers in Go serve multiple different purposes:
- They allow you to avoid copying large structs around
- They make function arguments or method receivers mutable (otherwise args are copied in, so modifications are on the copy)
- They can often cause things to be heap allocated instead of stack allocated
- They make something `nil`-able

Sometimes things are pointer types for reasons other than a desire to make them `nil`-able. And codebases are therefore often written with the assumption that those types aren't ever `nil`, and it's a programmer error if they are. It's not like TS where adding `| null` to a type is a very explicit signal that you _actually_ except this type to sometimes be `null`. It's also not quite like C, where segfaults can be legit security concerns (Go's pointers are actually a lot safer than C's, and don't allow a lot of the scary/dangerous stuff).

PR #33, opened 2025-10-02
Add `tiger db connect` message to service create/fork output

nathanjcochran:
See [Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1759334494530009?thread_ts=1759333945.188909&cid=C099S47CSRX) for background.

This adds an extra line to the `tiger service create` and `tiger service fork` output (when `--no-wait` isn't specified) which recommends using `tiger db connect` to connect to the service. This should help users discover the command for connecting to their database, which otherwise might not be obvious.

If `--password-storage` is `none` or the password failed to save, the message is not output (because we can't connect if we failed to save the user's password). If `--no-set-default` is specified, the output includes the service ID in the command (i.e. `tiger db connect ${service_id}`). Otherwise, the output just recommends `tiger db connect` (leveraging the fact that the new service was saved as the default).

For example:

```
$ tiger service create --free
🚀 Creating service 'db-54154' (auto-generated name)...
✅ Service creation request accepted!
📋 Service ID: p9o4zvnk8v
🔐 Password saved to system keyring for automatic authentication
🎯 Set service 'p9o4zvnk8v' as default service.
⏳ Waiting for service to be ready (wait timeout: 30m0s)...
⏳ Service status: QUEUED...
⏳ Service status: QUEUED...
⏳ Service status: QUEUED...
⏳ Service status: QUEUED...
🎉 Service is ready and running!
🔌 Run 'tiger db connect' to connect to your new service
```

### DIFF
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -448,7 +448,7 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 					if result != nil {
 						fmt.Fprintf(statusOutput, "❌ %v\n", result)
 					} else {
-						printConnectMessage(cmd.ErrOrStderr(), passwordSaved, createNoSetDefault, serviceID)
+						printConnectMessage(statusOutput, passwordSaved, createNoSetDefault, serviceID)
 					}
 				}
 
@@ -1332,7 +1332,7 @@ Examples:
 					return err
 				}
 				fmt.Fprintf(statusOutput, "🎉 Service fork completed successfully!\n")
-				printConnectMessage(cmd.ErrOrStderr(), passwordSaved, forkNoSetDefault, serviceID)
+				printConnectMessage(statusOutput, passwordSaved, forkNoSetDefault, serviceID)
 				return nil
 
 			case 401:


internal/tiger/cmd/service.go
```diff
@@ -430,8 +430,11 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 
 				// Wait for service to be ready
 				fmt.Fprintf(cmd.OutOrStdout(), "⏳ Waiting for service to be ready (wait timeout: %v)...\n", createWaitTimeout)
-				return waitForServiceReady(client, projectID, serviceID, createWaitTimeout, cmd)
-
+				if err := waitForServiceReady(client, projectID, serviceID, createWaitTimeout, cmd); err != nil {
+					return err
+				}
+				printConnectMessage(cmd.ErrOrStderr(), passwordSaved, createNoSetDefault, serviceID)
```
nathanjcochran: Using `cmd.ErrOrStderr()` here instead of `cmd.OutOrStdout()` in anticipation of https://github.com/timescale/tiger-cli/pull/32 being merged.

internal/tiger/cmd/service.go
```diff
@@ -914,6 +917,19 @@ func setDefaultService(serviceID string, output io.Writer) error {
...
+func printConnectMessage(output io.Writer, passwordSaved, noSetDefault bool, serviceID string) {
+	if !passwordSaved {
+		// We can't connect if no password was saved, so don't show message
+		return
+	} else if noSetDefault {
+		// If the service wasn't set as the default, include the serviceID in the command
+		fmt.Fprintf(output, "🔌 Run 'tiger db connect %s' to connect to your new service\n", serviceID)
+	} else {
+		// If the service was set as the default, no need to include the serviceID in the command
+		fmt.Fprintf(output, "🔌 Run 'tiger db connect' to connect to your new service\n")
```
cevian, line 929: I would put the service here even if set as default. But don't feel strongly

PR #34, opened 2025-10-02
Fix broken build

nathanjcochran:
The build is currently broken on `main` due to some recent merges that didn't play well together.

This fixes it by resolving some circular references and fixing some missing imports.

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -268,28 +268,28 @@ func buildConnectionString(service api.Service, pooled bool, role string, withPa
 	var connectionString string
 	if withPassword {
 		// Get password from storage if requested
-		storage := util.GetPasswordStorage()
-		password, err := storage.Get(service)
+		storage := password.GetPasswordStorage()
+		passwd, err := storage.Get(service)
 		if err != nil {
 			// Provide specific error messages based on storage type
 			switch storage.(type) {
-			case *util.NoStorage:
+			case *password.NoStorage:
 				return "", fmt.Errorf("password storage is disabled (--password-storage=none)")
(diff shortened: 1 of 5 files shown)


internal/tiger/util/convert.go
```diff
@@ -33,3 +33,17 @@ func AnySlice[T any](in []T) []any {
...
+	if len(ss) == 0 {
+		return nil
+	}
+
+	out := make([]T, len(ss))
+	for i, s := range ss {
+		out[i] = T(s)
+	}
+	return out
+}
```
nathanjcochran, line 49: Made this a generic function to get around having to import the `api` package from `utils`.

internal/tiger/cmd/db.go
```diff
@@ -268,28 +268,28 @@ func buildConnectionString(service api.Service, pooled bool, role string, withPa
 	var connectionString string
 	if withPassword {
 		// Get password from storage if requested
-		storage := util.GetPasswordStorage()
-		password, err := storage.Get(service)
+		storage := password.GetPasswordStorage()
```
graveland, line 271: nit: I'd rename `storage` to `passwordStorage` since just storage is so vague

graveland: Better: I'd probably rename the `password` package to `passwordstore` as well, since `password` is such an obvious local variable name instead of a package name.

nathanjcochran: I don't disagree, but am going to merge this as-is to unblock the build. Will definitely take these suggestions into consideration for future improvements/refactors 👍 (there are lots of things like this that we'd like to straighten out in time 😅).

PR #35, opened 2025-10-02
Add GitHub Actions testing workflow

nathanjcochran:
Adds a GitHub Actions workflow for building and testing the code. Also runs `shellcheck` on the `install.sh` script.

I also:
- Fixed a race condition that appeared when the tests were run with the `-race` flag, which seemed to be caused by `viper.WatchConfig()` (possibly https://github.com/spf13/viper/issues/174). Rather than relying on `viper.WatchConfig()` to reload the config file into viper's internal state when it changes, we now manually reload it whenever `config.Load()` is called. This ensures that MCP server tool calls (which always reload the config with `config.Load()`) will always be using the most up-to-date config state, which was the original purpose of adding `viper.WatchConfig()` in #20.
- Fixed a race in the auth tests where the test OAuth server wasn't always ready by the time we made a request to it (shows up more often in CI because the machine has more limited resources)
- Fixed a `shellcheck` issue in the `install.sh` script that was only failing in CI, because the GitHub Actions ubuntu image users an older version of shellcheck by default (see https://www.shellcheck.net/wiki/SC2002).

Closes [AGE-135](https://linear.app/tigerdata/issue/AGE-135/tiger-cli-add-testlint-workflow-to-github-actions)


.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Start D-Bus and keyring
+        run: |
+          # Start D-Bus session
+          eval $(dbus-launch --sh-syntax)
+          echo "DBUS_SESSION_BUS_ADDRESS=$DBUS_SESSION_BUS_ADDRESS" >> $GITHUB_ENV
+
+          # Initialize gnome-keyring with a test password
+          echo 'test' | gnome-keyring-daemon --unlock
+          eval $(echo 'test' | gnome-keyring-daemon --start --components=secrets)
+          echo "GNOME_KEYRING_CONTROL=$GNOME_KEYRING_CONTROL" >> $GITHUB_ENV
```
nathanjcochran, line 38: This was necessary to get the tests that rely on the keyring to pass in CI. Tbh, Claude wrote this code and I didn't take time to understand it, but it did make the tests pass 😅. It's definitely possible that some of this is wonky or unnecessary.

scripts/install.sh
```diff
@@ -298,7 +298,7 @@ verify_checksum() {
 
     # Format checksum for validation: "hash  filename"
     local formatted_checksum
-    formatted_checksum=$(printf "%s  %s\n" "$(cat "${checksum_file}" | tr -d '[:space:]')" "${filename}")
+    formatted_checksum=$(printf "%s  %s\n" "$(tr -d '[:space:]' < "${checksum_file}")" "${filename}")
```
nathanjcochran, line 301: Addresses https://www.shellcheck.net/wiki/SC2002, which was only showing up in CI because the GitHub Actions ubuntu image includes an [older version of shellcheck](https://github.com/actions/runner-images/blob/ba87e27210c2791016bbc815395798de8412e903/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L306), and the shellcheck docs say:

> This suggestion was enabled by default up to and including ShellCheck 0.10.0.
>
> In later versions, it is [optional](https://www.shellcheck.net/wiki/optional). It must be explicitly enabled with a [directive](https://www.shellcheck.net/wiki/directive) enable=useless-use-of-cat in a # shellcheck comment or .shellcheckrc

PR #36, opened 2025-10-02
AGE-182: only write explicit changes to config file

murrayju:
The config system had a problem where the _effective_ config (after resolving CLI args, env vars, and defaults) would be written to the config file when _any_ config was intended to be saved.

This caused several problems, most annoying of which being that temporarily passed CLI flags would get persisted (like `-o json`).

The solution is to never use the global `viper` instance to write the config file. Instead, a new instance is created, and only populated with the *file* content before being updated as needed. Then this can be written without carrying in unwanted values.

### DIFF
--- a/internal/tiger/config/config.go
+++ b/internal/tiger/config/config.go
@@ -6,7 +6,6 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strconv"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


internal/tiger/config/config.go
```diff
@@ -183,57 +243,101 @@ func setBool(key, val string) (bool, error) {
...
+			fieldValue := v.Field(i)
+			if fieldValue.CanSet() {
+				fieldValue.Set(reflect.ValueOf(value))
+				viper.Set(key, value)
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("field not found: %s", key)
+}
```
nathanjcochran: I don't love this reflection approach, tbh. Generally best to avoid reflection if possible imo ("Reflection is never clear" is one of the [Go proverbs](https://go-proverbs.github.io/)). I'd honestly prefer a big switch statement (even if it looks redundant) like before. Can the switch just be extracted from the `Set` call and used for this purpose too? I guess not, because `Set` takes a `string` value? But maybe things could be restructured somehow? 🤔

I could probably be convinced that reflection is fine here, if you have strong feelings. But it did stand out to me.

murrayju: In a vacuum, I might have argued for reflection for this function. But, thinking about it more, we already had a switch-case for `Set` that does some validation, and we might as well re-use that here.

https://github.com/timescale/tiger-cli/pull/36/commits/d0e098d6abba1780917671c13149074bc9272332

PR #37, opened 2025-10-03
Fix: Suppress MCP "Method not found" errors for unsupported endpoints

cevian:
## Summary
- Suppresses benign "Method not found" errors when MCP docs server doesn't implement optional resources/resource_templates endpoints
- Fixes error handling in StartStdio to properly handle nil errors
- Logs unsupported endpoints at debug level instead of error level

## Problem
When starting the MCP server (`tiger mcp start`), users were seeing error messages for endpoints that the docs MCP server doesn't implement:
```
{"message":"Failed to register resources from docs MCP server","error":"failed to list resources from remote server: calling \"resources/list\": Method not found"}
{"message":"Failed to register resource templates from docs MCP server","error":"failed to list resource templates from remote server: calling \"resources/templates/list\": Method not found"}
```

These errors are benign since resources and resource templates are optional MCP features, but they were confusing for users.

## Solution
Added `isMethodNotFoundError()` function that detects JSON-RPC "Method not found" errors (code -32601) and suppresses them for the specific endpoints we're calling. The errors are now logged at debug level instead.

## Implementation Notes
The implementation uses string matching because the actual error type (`github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.WireError`) is in an internal package that we cannot import. The function includes detailed documentation explaining why other approaches (errors.Is, errors.As, type assertion) don't work in this case.

## Test Plan
- [x] Verified MCP server starts without error messages: `go run cmd/tiger/main.go mcp start`
- [x] Verified errors are still logged at debug level: `go run cmd/tiger/main.go mcp start --debug`
- [x] Verified other functionality still works (prompts are registered successfully)

### DIFF
--- a/internal/tiger/mcp/proxy.go
+++ b/internal/tiger/mcp/proxy.go
@@ -44,7 +44,6 @@ func isMethodNotFoundError(err error) bool {
 
 	errStr := err.Error()
 	// Check for the specific "Method not found" JSON-RPC error
-	// We also verify it's for the resources endpoints we're calling
 	return strings.Contains(errStr, "Method not found")
 }
 
@@ -87,8 +86,7 @@ func (s *Server) registerDocsProxy(ctx context.Context) {
 	}
 
 	if err := proxyClient.RegisterResources(ctx, s.mcpServer); err != nil {
-		// Check if this is a "Method not found" error (code -32601)
-		// We use reflection to check the error code since wireError is unexported
+		// Check if this is a "Method not found" error as those are expected in servers that don't have any resources
 		if isMethodNotFoundError(err) {
 			logging.Debug("Resources not supported by remote MCP server")
 		} else {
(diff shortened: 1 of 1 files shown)


internal/tiger/mcp/proxy.go
```diff
@@ -12,6 +13,40 @@ import (
...
+//  4. Using errors.As() with a local struct that has the same shape as WireError doesn't work
+//     because errors.As requires type identity, not structural compatibility.
+//
+// 5. Using reflection to access the Code field would work but adds complexity and runtime overhead.
+//
+// Therefore, we use string matching on the error message. This is brittle but works for our
+// specific use case where we're checking for unsupported resources/resource_templates methods.
+// The error chain looks like:
+//   - "failed to list resources from remote server: calling \"resources/list\": Method not found"
+//   - "failed to list resource templates from remote server: calling \"resources/templates/list\": Method not found"
```
nathanjcochran, line 39: Do we really need a comment this long? Can't we just say "the error type is from an internal package"? 😅

internal/tiger/cmd/mcp.go
```diff
@@ -221,7 +221,8 @@ func startStdioServer(ctx context.Context) error {
 	}
 
 	// Start the stdio transport
-	if err := server.StartStdio(ctx); !errors.Is(err, context.Canceled) {
+	err = server.StartStdio(ctx)
+	if err != nil && !errors.Is(err, context.Canceled) {
 		return fmt.Errorf("failed to start MCP server: %w", err)
 	}
```
nathanjcochran, line 227: I'd vote for `if err := server.StartStdio(ctx); err != nil && !errors.Is(err, context.Canceled)` so the error is scoped to the if-block.

PR #41, opened 2025-10-03
Make `tiger_service_create` MCP tool set default service ID

nathanjcochran:
Makes the `tiger_service_create` MCP tool call set the default service ID if the `set_default` parameter is `true` (default: `true`), similar to how the `tiger service create` command works.

I wasn't sure if we wanted this at first, but after playing around, I think we do. It's really nice to be able to easily see service details with `tiger service describe` or get the connection string with `tiger db connection-string` immediately after the agent creates a service via the MCP tool.

### DIFF
--- a/specs/spec_mcp.md
+++ b/specs/spec_mcp.md
@@ -174,6 +174,7 @@ Create a new database service.
 - `replicas` (number, optional): Number of high-availability replicas - default: 0
 - `wait` (boolean, optional): Wait for service to be ready - default: false
 - `timeout` (number, optional): Timeout for waiting in minutes - default: 30
+- `set_default` (boolean, optional): Set the newly created service as the default service for future commands - default: true
 
 **Returns:** Service object with creation status and details.
 


internal/tiger/cmd/service.go
```diff
@@ -902,12 +902,7 @@ func handlePasswordSaving(service api.Service, initialPassword string, output io
 }
 
 // setDefaultService sets the given service as the default service in the configuration
-func setDefaultService(serviceID string, output io.Writer) error {
-	cfg, err := config.Load()
-	if err != nil {
-		return fmt.Errorf("failed to load config: %w", err)
-	}
```
nathanjcochran, line 909: Idk why we were re-loading the config here. I don't think it's necessary to reload it, so I updated the code to just pass in the config that was already loaded.

internal/tiger/mcp/server.go
```diff
@@ -84,18 +84,18 @@ func (s *Server) createAPIClient() (*api.ClientWithResponses, error) {
...
-		return "", fmt.Errorf("failed to load config: %w", err)
+		return nil, fmt.Errorf("failed to load config: %w", err)
 	}
 
 	if cfg.ProjectID == "" {
-		return "", fmt.Errorf("project ID is required. Please run 'tiger auth login' with --project-id")
+		return nil, fmt.Errorf("project ID is required. Please run 'tiger auth login' with --project-id")
 	}
-	return cfg.ProjectID, nil
+	return cfg, nil
```
nathanjcochran, line 98: Updated this to return the full config instead of just the project ID (while still validating that the project ID is set).

internal/tiger/mcp/service_tools.go
```diff
@@ -497,11 +502,21 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
...
 
+		// Set as default service if requested (defaults to true)
+		if input.SetDefault {
+			if err := cfg.Set("service_id", serviceID); err != nil {
+				// Log warning but don't fail the service creation
+				logging.Debug("MCP: Failed to set service as default", zap.Error(err))
+			} else {
+				logging.Debug("MCP: Set service as default", zap.String("service_id", serviceID))
+			}
+		}
```
nathanjcochran, line 513: This is the main change.

internal/tiger/cmd/service.go
murrayju: It doesn't "re-load" so much as grab a copy. Agreed that this is weird/unnecessary, and passing in is better. There are still things like this I'd like to improve about the config, but I was avoiding broader changes in #36

internal/tiger/mcp/server.go
murrayju: This function is weird, and probably unnecessary. Why is it attached to `Server`, when `s` isn't even used? More config weirdness to clean up later, I'm not worried about it now.

nathanjcochran: This function is called at the start of every MCP tool call, and it does encapsulate the common projectID validation and error messages so they aren't repeated every time, which seems worthwhile to me. If anything, there's probably more shared logic that we could encapsulate in this function (though the name and return type would probably need to change).

> Why is it attached to Server, when s isn't even used?

You're right that it currently doesn't _need_ to have a receiver, but it also doesn't really bother me, fwiw. Lots of people do this as a way of organizing/namespacing functions within a package. I think long-term, it probably makes sense to be related to the `Server` type. For instance, if we weren't using the global `viper` instance (and I'd personally prefer if we weren't, as even [viper itself discourages its use](https://pkg.go.dev/github.com/spf13/viper#readme-viper-or-vipers)), we'd probably need to access the viper instance (or something encapsulating it) via the `s *Server` receiver in order to load the config.

PR #42, opened 2025-10-04
Add MCP tool for querying a service

nathanjcochran:
Adds a new `db_execute_query` tool (I omitted the `tiger_` prefix, per the changes in #30) that is capable of executing a SQL query and returning the results.

I made a handful of small changes compared to the original spec, such as:
- Making the service ID parameter required, instead of optional. None of the other MCP tools currently utilize the default service, and given that it could be risky to run a SQL query against the wrong database, I figured it would be better to make it explicit. The LLM is able to list services and get service IDs, so it shouldn't really cause much friction.
- Changing `timeout` to `timeout_seconds` for clarity (similar changes were made in #30).
- Adding `role` and `pooled` parameters for parity with the `tiger db connect` flags.
- Returning column types as well as names in the response.
- Returning "rows affected" instead of "row count". Rows affected essentially _is_ the row count for row-returning queries (e.g. `SELECT`), but also shows the number of rows modified for `INSERT`/`UPDATE`/`DELETE` queries, making it useful in those scenarios as well.
- Returning the execution time as a string in Go's standard `time.Duration` output format, rather than in milliseconds. For long-running queries, this will hopefully make the duration easier for both humans and LLMs to parse and understand.

I moved the logic for building connection strings from the `internal/tiger/cmd` package into `internal/tiger/password` so that it could be shared with the MCP tools. That seemed like a fairly reasonable place to put it, even if not entirely ideal. The code depends on the `internal/tiger/api` package, so it can't go in `util`, and I didn't want to make another package just for the connection string logic (I think having a bunch of very small packages is kind of an anti-pattern in Go). I think we may need to rethink our package organization at some point, but this seemed good enough for now 🤷‍♂️.

Note that there was talk about also adding an MCP tool for executing a query from a file (see [here](https://iobeam.slack.com/archives/C09BFEDCWBW/p1759597691222019?thread_ts=1759528276.214829&cid=C09BFEDCWBW)). I did not do that in this PR, but it should be easy to do in a follow-up if we want it.

Closes [AGE-159](https://linear.app/tigerdata/issue/AGE-159/tiger-cli-database-tool-calls)

### DIFF
--- a/internal/tiger/mcp/db_tools.go
+++ b/internal/tiger/mcp/db_tools.go
@@ -95,9 +95,9 @@ func (s *Server) registerDatabaseTools() {
 	mcp.AddTool(s.mcpServer, &mcp.Tool{
 		Name:  "db_execute_query",
 		Title: "Execute SQL Query",
-		Description: `Execute a SQL query against a service database.
+		Description: `Execute a single SQL query against a service database.
 
-This tool connects to a PostgreSQL database service in TigerData Cloud and executes the provided SQL query, returning the results with column names, row data, and execution metadata. Perfect for data exploration, schema inspection, and database operations.
+This tool connects to a PostgreSQL database service in TigerData Cloud and executes the provided SQL query, returning the results with column names, row data, and execution metadata. Multi-statement queries are not supported.
 
 WARNING: Use with caution - this tool can execute any SQL statement including INSERT, UPDATE, DELETE, and DDL commands. Always review queries before execution.`,
 		InputSchema:  DBExecuteQueryInput{}.Schema(),


internal/tiger/cmd/db.go
```diff
@@ -192,8 +205,13 @@ Examples:
...
 
-			// Build connection string for testing
-			connectionString, err := buildConnectionString(service, dbTestConnectionPooled, dbTestConnectionRole, false, cmd.ErrOrStderr())
+			// Build connection string for testing with password (if available)
+			connectionString, err := password.BuildConnectionString(service, password.ConnectionStringOptions{
+				Pooled:       dbTestConnectionPooled,
+				Role:         dbTestConnectionRole,
+				PasswordMode: password.PasswordOptional,
+				WarnWriter:   cmd.ErrOrStderr(),
+			})
```
nathanjcochran, line 214: This is the only place we use `PasswordOptional`. That causes `BuildConnectionString` to attempt to include the password if possible, but to return the connection string without the password if the password can't be found.

The previous code would build the connection string without the password, but then `testDatabaseConnection` would call `buildConnectionConfig`, which parsed the connection string into a config struct and attempted to add the password to the config, if possible. That seemed like a pretty roundabout way of doing it (building a connection string just to parse it back into a config), so I got rid of that logic and added the `PasswordOptional` option when building the connection string.

That being said, I'm not 100% convinced it's worth it to return the connection string without a password here. The old code in `buildConnectionConfig` had this comment:

> // Note: If keyring password retrieval fails, we let pgx try without it
> // This allows fallback to other authentication methods

But I'm not sure what "other authentication methods" it's talking about here, or if it's really worth worrying about it. If we _don't_ need to support this, we could get rid of `PasswordOptional` and replace `PasswordMode` with a `PasswordRequired` boolean, which would probably make the code a little more straightforward/clear.

cevian: The other method I originally had in mind was PGPASSWORD env variable being set in the shell or a .pgpassword file (even if not set by the cli). I still think its worth allowing this kind of fallback.

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+	schema.Properties["execution_time"].Description = "Query execution time as a human-readable duration string"
+	schema.Properties["execution_time"].Examples = []any{"123ms", "1.5s", "45.2µs"}
+
+	return schema
+}
+
+// registerDatabaseTools registers database operation tools with comprehensive schemas and descriptions
+func (s *Server) registerDatabaseTools() {
+	mcp.AddTool(s.mcpServer, &mcp.Tool{
+		Name:  "db_execute_query",
```
cevian, line 96: as a followup/discussion point: should we have a separate tool for "read-only" that make sure you are using a read-only role or connection? Alternatively, we can have another input parameter "enforce read-only"

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/timescale/tiger-cli/internal/tiger/logging"
+	"github.com/timescale/tiger-cli/internal/tiger/password"
+	"github.com/timescale/tiger-cli/internal/tiger/util"
+)
+
+// DBExecuteQueryInput represents input for tiger_db_execute_query
+type DBExecuteQueryInput struct {
+	ServiceID      string `json:"service_id"`
+	Query          string `json:"query"`
```
cevian, line 22: I think we should also take in query parameter array for $1, $2 substitution to help llm avoid sql injection

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+	// Connect to database
+	conn, err := pgx.Connect(queryCtx, connString)
+	if err != nil {
+		return nil, DBExecuteQueryOutput{}, fmt.Errorf("failed to connect to database: %w", err)
+	}
+	defer conn.Close(context.Background())
+
+	// Execute query and measure time
+	startTime := time.Now()
+	rows, err := conn.Query(queryCtx, input.Query)
```
cevian: do we also want to set a statement_timeout on the postgres side as well or does pgx do that automatically?

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+	schema.Properties["timeout_seconds"].Minimum = util.Ptr(0.0)
+	schema.Properties["timeout_seconds"].Default = util.Must(json.Marshal(30))
+	schema.Properties["timeout_seconds"].Examples = []any{10, 30, 60}
+
+	schema.Properties["role"].Description = "Database role/username to connect as"
+	schema.Properties["role"].Default = util.Must(json.Marshal("tsdbadmin"))
+	schema.Properties["role"].Examples = []any{"tsdbadmin", "readonly", "postgres"}
+
+	schema.Properties["pooled"].Description = "Use connection pooling (if available for the service)"
+	schema.Properties["pooled"].Default = util.Must(json.Marshal(false))
```
cevian, line 51: I think default should be true?

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+
+	return schema
+}
+
+// registerDatabaseTools registers database operation tools with comprehensive schemas and descriptions
+func (s *Server) registerDatabaseTools() {
+	mcp.AddTool(s.mcpServer, &mcp.Tool{
+		Name:  "db_execute_query",
+		Title: "Execute SQL Query",
+		Description: `Execute a SQL query against a service database.
```
cevian:
```suggestion
		Description: `Execute a single SQL query against a service database.
```

internal/tiger/mcp/db_tools.go
```diff
@@ -0,0 +1,225 @@
...
+
+This tool connects to a PostgreSQL database service in TigerData Cloud and executes the provided SQL query, returning the results with column names, row data, and execution metadata. Perfect for data exploration, schema inspection, and database operations.
```
cevian:
```suggestion
This tool connects to a PostgreSQL database service in TigerData Cloud and executes the provided SQL query, returning the results with column names, row data, and execution metadata. Multi-statement queries not supported.
```

internal/tiger/mcp/db_tools.go
nathanjcochran: I considered this too, but on further reflection, I'm not really sure what it would be guarding against. If the entire SQL query is crafted by the LLM anyways, what's the value of having it pass parameters separately? It isn't really like a normal backend, where you have predefined "safe" queries that you want to parameterize with "unsafe" user-provided input. In this case, both the queries and the parameters are coming from the same LLM.

That being said, I'm not opposed to supporting parameters, as a way of giving the LLM more flexibility. For example, if a user provided a query with placeholders, it would allow the LLM to just execute it directly with specific parameters without having to modify the query. So yeah, I can implement this (I just think it's worth pointing out that I don't really think it affects security in any meaningful way).

internal/tiger/mcp/db_tools.go
nathanjcochran: When the `queryCtx` is cancelled, `pgx` closes the connection and cancels the query under the hood (see [here](https://github.com/jackc/pgx/issues/1794#issuecomment-1806824969)). We should not need to set `statement_timeout`.

internal/tiger/mcp/db_tools.go
nathanjcochran: It [defaults to `false`](https://github.com/timescale/tiger-cli/blob/6b901dc0e331dca7482ca70d951e04a3c2fe1469/internal/tiger/cmd/db.go#L161) for `tiger db connect`. What's your rationale for the difference?

internal/tiger/mcp/db_tools.go
nathanjcochran: Yeah, I think a separate input parameter is probably the way to go. I do think we should do that in a follow-up though.

internal/tiger/mcp/db_tools.go
nathanjcochran: Support for parameterized queries added here: https://github.com/timescale/tiger-cli/pull/42/commits/178a33d2c4c181e528c2fc7eb40620d4066cd7f7

internal/tiger/mcp/db_tools.go
nathanjcochran: https://github.com/timescale/tiger-cli/pull/42/commits/20462b37df3392ac769a9b8fbec982a8cf491855

internal/tiger/mcp/db_tools.go
nathanjcochran: https://github.com/timescale/tiger-cli/pull/42/commits/20462b37df3392ac769a9b8fbec982a8cf491855

internal/tiger/mcp/db_tools.go
cevian: I think thats wrong there too. create an issue https://linear.app/tigerdata/issue/AGE-200/make-pooled=true-the-default-in-climcp

PR #46, opened 2025-10-07
Log forwarded 4xx errors

Askir:
Relies on: https://github.com/timescale/savannah-public/pull/49
The errors for these cases are still suboptimal:
- Creating more than 2 free services (insufficient permission ->  Error: Unauthorized. Entitlement check has failed.)
- Forking more than 2 free services (insufficient permission -> Error: Unauthorized. Entitlement check has failed.)
- Forking before a backup is read (still gives a 500)
- Forking to an invalid (too early timestamp) (still gives a 500)

Will dive a bit deeper into graphql to try get better errors. At least it should be propagated with this.

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -325,6 +325,10 @@ func getServiceDetails(cmd *cobra.Command, args []string) (api.Service, error) {
 		return api.Service{}, exitWithErrorFromStatusCode(resp.StatusCode(), resp.JSON4XX)
 	}
 
+	if resp.JSON200 == nil {
+		return api.Service{}, fmt.Errorf("empty response from API")
+	}
+
 	return *resp.JSON200, nil
 }
 
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -194,6 +194,12 @@ func buildServiceListCmd() *cobra.Command {
 				return nil
 			}
 
+			if resp.JSON200 == nil {
+				fmt.Fprintln(statusOutput, "🏜️  No services found! Your project is looking a bit empty.")
+				fmt.Fprintln(statusOutput, "🚀 Ready to get started? Create your first service with: tiger service create")
+				return nil
+			}
+
 			// Output services in requested format
 			return outputServices(cmd, services, cfg.Output, withPassword)
 		},
--- a/internal/tiger/cmd/integration_test.go
+++ b/internal/tiger/cmd/integration_test.go
@@ -83,6 +83,7 @@ func executeIntegrationCommand(args ...string) (string, error) {
 // TestServiceLifecycleIntegration tests the complete authentication and service lifecycle:
 // login -> whoami -> create -> describe -> update-password -> delete -> logout
 func TestServiceLifecycleIntegration(t *testing.T) {
+	config.SetTestServiceName(t)
 	// Check for required environment variables
 	publicKey := os.Getenv("TIGER_PUBLIC_KEY_INTEGRATION")
 	secretKey := os.Getenv("TIGER_SECRET_KEY_INTEGRATION")
@@ -399,8 +400,8 @@ func TestServiceLifecycleIntegration(t *testing.T) {
 		}
 
 		// Check that error indicates service not found
-		if !strings.Contains(err.Error(), "not found") && !strings.Contains(err.Error(), "404") {
-			t.Errorf("Expected 'not found' error for deleted service, got: %v", err)
+		if !strings.Contains(err.Error(), "no service with that id exists") {
+			t.Errorf("Expected 'no service with that id exists' error for deleted service, got: %v", err)
 		}
 
 		// Check that it returns the correct exit code (this should be required)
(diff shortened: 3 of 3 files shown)


internal/tiger/api/client_util.go
```diff
@@ -94,14 +95,41 @@ func ValidateAPIKeyWithClient(client ClientWithResponsesInterface, projectID str
...
+}
+
+// FormatAPIErrorFromBody attempts to parse an API error from a response body.
+// If the body contains a valid API error with a message, it will be used; otherwise the fallback message is returned.
+func FormatAPIErrorFromBody(body []byte, fallback string) error {
+	if len(body) > 0 {
+		var apiErr Error
+		if err := json.Unmarshal(body, &apiErr); err == nil {
+			if apiErr.Message != nil && *apiErr.Message != "" {
+				return fmt.Errorf("%s", *apiErr.Message)
```
nathanjcochran: There's a simpler way to do this (we don't really need to use a format string here, since the format is just `%s`):

```suggestion
				return errors.New(*apiErr.Message)
```

Might need to import the `errors` package at the top of the file if it's not already.

internal/tiger/api/client_util_test.go
```diff
@@ -118,3 +118,115 @@ func TestValidateAPIKey_Integration(t *testing.T) {
...
+				t.Errorf("Expected error message '%s', got '%s'", tt.expected, err.Error())
+			}
+		})
+	}
+}
+
+// Helper function for creating string pointers
+func strPtr(s string) *string {
+	return &s
+}
```
nathanjcochran: Can use the generic [util.Ptr](https://github.com/timescale/tiger-cli/blob/2b7037b143c3d2f6ead7d7c630387d47a9724133/internal/tiger/util/convert.go#L10-L12) function instead (Claude always want to add helper functions like this - we should probably add something to the `CLAUDE.md` file about using the existing methods from the `util` package instead).

internal/tiger/cmd/db.go
```diff
@@ -330,13 +330,17 @@ func getServiceDetails(cmd *cobra.Command, args []string) (api.Service, error) {
...
 	case 404:
-		return api.Service{}, exitWithCode(ExitServiceNotFound, fmt.Errorf("service '%s' not found in project '%s'", serviceID, projectID))
+		return api.Service{}, exitWithCode(ExitServiceNotFound, api.FormatAPIError(resp.JSON404, fmt.Sprintf("service '%s' not found in project '%s'", serviceID, projectID)))
 	default:
-		return api.Service{}, fmt.Errorf("API request failed with status %d", resp.StatusCode())
+		statusCode := resp.StatusCode()
+		if statusCode >= 400 && statusCode < 500 {
+			return api.Service{}, api.FormatAPIErrorFromBody(resp.Body, fmt.Sprintf("API request failed with status %d", statusCode))
+		}
+		return api.Service{}, fmt.Errorf("API request failed with status %d", statusCode)
```
nathanjcochran: I think this is fine for now, but I also think it's worth mentioning that if the `openapi.yaml` spec accurately reflected all of the status codes that could be returned from each endpoint (e.g. like it does [here](https://github.com/timescale/tiger-cli/blob/2b7037b143c3d2f6ead7d7c630387d47a9724133/openapi.yaml#L379-L382) and [here](https://github.com/timescale/tiger-cli/blob/2b7037b143c3d2f6ead7d7c630387d47a9724133/openapi.yaml#L1119-L1136)), the generated client library would parse the error responses for us, and they could be all be accessed via `resp.JSONXXX` fields, similar to how you're accessing [resp.JSON404](https://github.com/timescale/tiger-cli/blob/2b7037b143c3d2f6ead7d7c630387d47a9724133/internal/tiger/api/client.go#L2347) above.

We could even go further and add a method like this in `./internal/tiger/api/client_util.go`:

```go
func (e *Error) Error() string {
	if e.Message == nil {
		return "unknown error"
	}
	return *e.Message
}
```

That would make all of those error types (e.g. `resp.JSON404`) automatically implement the go `error` interface, so they could be passed directly to `exitWithCode`. Then we could get rid of the `FormatAPIError` and `FormatAPIErrorFromBody` methods (I'm not sure how important the fallback messages really are, but I would hope they aren't actually necessary 🤷‍♂️)

nathanjcochran (approved):
Left a couple minor comments, but no show-stoppers. Looks good to me ✅ 

I definitely think there's more we could do to clean up the error handling, but that could be done in a follow-up. This gets the job done for now 👍.

internal/tiger/cmd/service.go
```diff
@@ -211,13 +215,17 @@ func buildServiceListCmd() *cobra.Command {
...
 			case 404:
-				return fmt.Errorf("project not found")
+				return api.FormatAPIError(resp.JSON404, "project not found")
 			default:
-				return fmt.Errorf("API request failed with status %d", resp.StatusCode())
+				statusCode := resp.StatusCode()
+				if statusCode >= 400 && statusCode < 500 {
+					return api.FormatAPIErrorFromBody(resp.Body, fmt.Sprintf("API request failed with status %d", statusCode))
+				}
+				return fmt.Errorf("API request failed with status %d", statusCode)
```
nathanjcochran: This is often repeated. Can we encapsulate this in a helper function?

PR #47, opened 2025-10-07
Fix test isolation issues with keyring storage

cevian:
## Summary
- Fixes test failures when running `go test ./... -count=1` due to keyring race conditions
- Tests from different packages were competing for the same keyring service name
- Implemented unique keyring namespaces per test using `t.Name()`

## Problem
Tests were failing when run without cache (`-count=1`) because:
- All tests shared the same "tiger-cli-test" keyring service name
- When tests from different packages ran in parallel, they would interfere with each other
- This caused random failures where tests would get wrong API keys or fail to find expected values

## Solution
Implemented `SetTestServiceName(t *testing.T)` function that:
- Creates unique keyring service names based on test name (e.g., "tiger-test-TestAuthLogin")
- Automatically registers cleanup via `t.Cleanup()` - no manual cleanup needed
- Prevents race conditions when tests run in parallel across packages

## Changes
- **api_key.go**: Added `SetTestServiceName()` with automatic cleanup
- **7 test files**: Updated to call `SetTestServiceName(t)` in setup functions
- Removed all manual `ResetTestServiceName()` calls - cleanup is now automatic

## Test Plan
- [x] Run `go test ./... -count=1` - all tests pass
- [x] Run `go test ./... -count=3` - tests pass consistently
- [x] Run tests with parallel packages (default) - no race conditions
- [x] Verify keyring isolation by checking unique service names per test


internal/tiger/config/api_key.go
```diff
@@ -12,22 +12,40 @@ import (
...
+// The cleanup is automatically registered with t.Cleanup()
+func SetTestServiceName(t *testing.T) {
+	testServiceNameOverride = "tiger-test-" + t.Name()
+
+	// Automatically clean up when the test finishes
+	t.Cleanup(func() {
+		testServiceNameOverride = ""
+	})
+}
+
```
nathanjcochran, line 48: One potential worry I have with this approach is that there could be race conditions where multiple parallel tests call `SetTestServiceName` concurrently, thereby overwriting the value of `testServiceNameOverride` set by the other tests.

I guess that's not an issue right now, because we don't have any parallel tests within the same package, and tests in different packages are fully isolated from each other (i.e. they don't share memory or this `testServiceNameOverride` variable). Right?

Still, it seems like a potential gotcha for down the road. I feel like it would be better to somehow pass the keyring service name into the calls that access the keyring (even though that would be annoying in many ways).

In any case, I think this is fine for now. Definitely an improvement over what we had 👍

cevian: Yeah this needs a better refactor but merging for now

PR #49, opened 2025-10-07
AGE-124: env output format for services

murrayju:
This adds support for `env` as an output format. It is only supported on individual service outputs, and other usages will fail with an error.

The output uses the standard `PG*` environment variable names, to increase the liklihood of being directly usable in `.env` files.

image

Like the other output formats, this only controls what is written to `stdout`. The user is then able to redirect to a file, if so desired.

```bash
tiger svc create --free -o env --with-password >> .env
```

As part of this, I removed the global `--output` flag, and instead implemented individual flags on the commands that use it. This allowed for more clarity and accuracy in validation around the presence of the `env` value.

I also noticed that the `tiger config show` command had a bug in that it was affected by any CLI args (or env vars) passed to it. We really want this to reflect the config in the file (and defaults if unset). I overhauled how this works, and also added options for `--no-defaults` and `--with-env` to control the behavior a bit.

image

image

image

image

image

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -442,7 +442,10 @@ main() {
     binary_path="$(extract_archive "${archive_name}" "${tmp_dir}" "${platform}")"
 
     # Copy binary to install directory
+    # Remove existing binary first to prevent errors related
+    # to swapping out a currently executing binary
     log_info "Installing to ${install_dir}..."
+    rm -f "${install_dir}/${BINARY_NAME}"
     cp "${binary_path}" "${install_dir}/${BINARY_NAME}"
 
     # Verify installation


internal/tiger/cmd/service.go
```diff
@@ -620,6 +620,10 @@ type OutputService struct {
 	api.Service
 	ConnectionString *string `json:"connection_string,omitempty" yaml:"connection_string,omitempty"`
 	Password         *string `json:"password,omitempty" yaml:"password,omitempty"`
+	Role             *string `json:"role,omitempty" yaml:"role,omitempty"`
+	Host             *string `json:"host,omitempty" yaml:"host,omitempty"`
+	Port             *int    `json:"port,omitempty" yaml:"port,omitempty"`
+	Database         *string `json:"database,omitempty" yaml:"database,omitempty"`
```
nathanjcochran: Is there a reason these need to be pointers? Could we just compare against the zero value instead to check if they're set? Same with the pre-existing `ConnectionString` and `Password` fields.

Alternatively, you could just embed `password.ConnectionDetails` in this struct instead of duplicating all of these fields (that's probably what I'd recommend).

internal/tiger/cmd/root.go
```diff
@@ -71,7 +71,7 @@ tiger auth login
 	// Add persistent flags
 	cmd.PersistentFlags().StringVar(&configDir, "config-dir", config.GetDefaultConfigDir(), "config directory")
 	cmd.PersistentFlags().BoolVar(&debug, "debug", false, "enable debug logging")
-	cmd.PersistentFlags().VarP((*outputFlag)(&output), "output", "o", "output format (json, yaml, table)")
+	cmd.PersistentFlags().VarP((*outputFlag)(&output), "output", "o", "output format (json, yaml, env, table)")
```
nathanjcochran: I have some misgivings about listing `env` as an output format for the global `--output` flag, even though that format is not supported by all commands.

Can we either get rid of `--output` as a global flag (and just define it for the commands that actually use it), or see if there's a way to override it per-command? I'm honestly not sure what viper does if you re-define a flag with the same name for a subcommand 🤔.

internal/tiger/cmd/service.go
```diff
@@ -796,31 +824,37 @@ func prepareServiceForOutput(service api.Service, withPassword bool, output io.W
...
+		outputSvc.Role = &connectionDetails.Role
+		outputSvc.Host = &connectionDetails.Host
+		outputSvc.Port = &connectionDetails.Port
+		outputSvc.Database = &connectionDetails.Database
+		if connectionDetails.Password != "" {
+			outputSvc.Password = &connectionDetails.Password
+		}
+	} else if output != nil {
+		fmt.Fprintf(output, "⚠️  Warning: Failed to get connection details: %v\n", err)
+	}
```
nathanjcochran: Minor, but it's idiomatic to have the error-handling condition first. I think you can also scope both `connectionDetails` and `err` to the conditional block (I don't _think_ the connection details are used anywhere else?):


```suggestion
	if connectionDetails, err := password.GetConnectionDetails(service, opts); err != nil {
	    fmt.Fprintf(output, "⚠️  Warning: Failed to get connection details: %v\n", err)
	} else {
		outputSvc.Role = &connectionDetails.Role
		outputSvc.Host = &connectionDetails.Host
		outputSvc.Port = &connectionDetails.Port
		outputSvc.Database = &connectionDetails.Database
		if connectionDetails.Password != "" {
			outputSvc.Password = &connectionDetails.Password
		}
	}
```

internal/tiger/config/output.go
```diff
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-var ValidOutputFormats = []string{"json", "yaml", "table"}
+var ValidOutputFormats = []string{"json", "yaml", "env", "table"}
```
nathanjcochran: Does this mean `env` is a valid output format in the config file? i.e. `tiger config set output env` is valid?

I think it probably shouldn't be, since that would cause several commands to fail with an error (but again, this is kind of why I feel like re-using the global output option for this is kind of wonky).

internal/tiger/password/connection.go
```diff
@@ -90,21 +67,19 @@ func BuildConnectionString(service api.Service, opts ConnectionStringOptions) (s
...
-	host = *endpoint.Host
 
-	if endpoint.Port != nil {
-		port = *endpoint.Port
-	} else {
-		port = 5432 // Default PostgreSQL port
+	details := &ConnectionDetails{
+		Role:     opts.Role,
+		Host:     *endpoint.Host,
+		Port:     5432,   // Default PostgreSQL port
```
nathanjcochran, line 79: I know this was pre-existing logic, but I'm honestly not convinced it's right. If we fail to return a port from the API, I kind of doubt 5432 is just magically going to be the correct port. 

Then again, maybe it is 🤷‍♂️. Can revisit later.

internal/tiger/cmd/service.go
```diff
@@ -796,31 +824,37 @@ func prepareServiceForOutput(service api.Service, withPassword bool, output io.W
...
-	}
-
-	// Build connection string
-	passwordMode := password.PasswordExclude
 	if withPassword {
-		passwordMode = password.PasswordRequired
+		outputSvc.Password = outputSvc.InitialPassword
 	}
-	connectionString, err := password.BuildConnectionString(service, password.ConnectionStringOptions{
+	outputSvc.InitialPassword = nil
```
nathanjcochran, line 800: If I'm understanding correctly, it looks like we always discard `InitialPassword` here, and then look the password up using the provided password storage mechanism in `password.GetConnectionDetails`.

I think we should pass `InitialPassword` into `GetConnectionDetails` instead (e.g. add an `InitialPassword` field to `ConnectionDetailsOptions`) and use that if provided, rather than always looking it up via the password storage mechanism. That way, a user who set `--password-storage=none` could still get their password back in the output when creating a service (which I don't think is currently possible, given this implementation).

internal/tiger/password/connection.go
```diff
@@ -139,17 +113,62 @@ func BuildConnectionString(service api.Service, opts ConnectionStringOptions) (s
...
 		// Password is handled separately via PGPASSWORD env var or ~/.pgpass file
 		// This ensures credentials are never visible in process arguments
-		connectionString = fmt.Sprintf("postgresql://%s@%s:%d/%s?sslmode=require", opts.Role, host, port, database)
+		connectionString = fmt.Sprintf("postgresql://%s@%s:%d/%s?sslmode=require", opts.Role, details.Host, details.Port, details.Database)
+	default:
+		// Include password in connection string
+		connectionString = fmt.Sprintf("postgresql://%s:%s@%s:%d/%s?sslmode=require", opts.Role, details.Password, details.Host, details.Port, details.Database)
 	}
 
 	return connectionString, nil
```
nathanjcochran: Minor, but I'd just return from within the switch statement, rather than setting a local variable and returning it later

```suggestion
	// Build connection string in PostgreSQL URI format
	switch details.Password {
	case "":
            // Build connection string without password (default behavior)
            // Password is handled separately via PGPASSWORD env var or ~/.pgpass file
            // This ensures credentials are never visible in process arguments
	    return fmt.Sprintf("postgresql://%s@%s:%d/%s?sslmode=require", opts.Role, details.Host, details.Port, details.Database), nil
	default:
	    // Include password in connection string
	    return fmt.Sprintf("postgresql://%s:%s@%s:%d/%s?sslmode=require", opts.Role, details.Password, details.Host, details.Port, details.Database), nil
}
```

nathanjcochran (commented):
I still have some misgivings about `env` as an output format, given output is a global config option/flag, and `env` is not universally valid.

I think maybe we should consider:
- Not allowing `env` via `tiger config set output`, since that would break some commands, and I don't think it's likely that users would want to configure the env format globally anyways.
- Getting rid of the global `--output` flag and replacing it with per-command `--output` flags. That way, they could support different options and it wouldn't feel as weird (honestly, it's kind of weird that it's a global flag anyways, since not all commands support it).
- For command that support the flag, still falling back to the global config option (the value set via `tiger config set output`) if not otherwise specified.

Otherwise, I think this largely makes sense. Though I also think maybe we should add support to `tiger db connection-string` (or a new command, if `connection-string` feels to weird?) for outputting these env vars, so users can do it later too (not just when initially creating a service).

Left a handful of other mostly-minor comments below.

murrayju:
Yeah, I'd be happy to get rid of the global `--output` flag, I don't think it makes sense. @cevian can you weigh in on that?

> so users can do it later too (not just when initially creating a service)

It works with `tiger svc describe` in this PR.

internal/tiger/cmd/service.go
murrayju: I'm not sure. _Every_ field in `api.Service` is a pointer, so it felt like these should be too. I figured it was related to how `omitempty` works, but I don't actually know that for a fact. This is directly related to the discussion we had in Slack yesterday. I feel like consistency is the most important thing here. I'd rather address this later by making them _all_ non-pointers (if that's the direction we want)

> could just embed `password.ConnectionDetails`

Yeah, I agree, if we can make the pointer-ness the same for both structs.

nathanjcochran: > _Every_ field in `api.Service` is a pointer

The fields in `api.Service` are pointers because that code is generated from the `openapi.yaml` spec via [openapi-codegen](https://github.com/oapi-codegen/oapi-codegen), and it has a default behavior of making optional fields pointer types (which is a reasonable default for the codegen library because some applications _might_ need to distinguish between an omitted/null JSON value and a zero value that was actually provided).

It could be fixed by updating the [Service](https://github.com/timescale/tiger-cli/blob/e8f0a89b3c706c064d3cb5c77110af545c878609/openapi.yaml#L766-L844) schema in the `openapi.yaml` spec to include a `required` field that lists the fields that should always be present (like we do in other places - e.g. [here](https://github.com/timescale/tiger-cli/blob/e8f0a89b3c706c064d3cb5c77110af545c878609/openapi.yaml#L926-L932)). Imo, it would be nice to take the time to set that correctly for all types/fields in the spec. Alternatively, it could also be fixed by setting the [x-go-type-skip-optional-pointer](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#x-go-type-skip-optional-pointer---do-not-add-a-pointer-type-for-optional-fields-in-structs) extension to `true` (which would allow us to generate non-pointer types for optional fields, if we don't need to distinguish omitted/`null` JSON values from zero values for optional fields).

But in general, it's just the result of an imperfect `openapi.yaml` schema, and I don't think it should be taken as authoritative (note that if you do want to update that schema, make sure to keep it synchronized with [the copy in savannah-public](https://github.com/timescale/savannah-public/blob/main/openapi.yaml))

> I figured it was related to how omitempty works

The `omitempty` tag will omit a field from marshaled JSON output if it contains the empty/zero value. For pointer types that `nil`, for strings it's `""`, for ints it's `0`, etc. So it should work fine whether or not the fields are pointers, so long as we don't ever need to output the zero value itself (i.e. `0`, `""`, etc.)

> I feel like consistency is the most important thing here

In general I agree, but I think it's okay if generated code adheres to different patterns (since we don't have as much control over it). I think we should mostly try to be consistent and idiomatic in the code we're writing ourselves.

murrayju: Sounds great to me 👍 
https://github.com/timescale/tiger-cli/pull/49/commits/9175597dcad14a29dac2df7c89bd1cf2ebc9619b

internal/tiger/password/connection.go
murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/bbd781a9829a955a48c773dacd95d031e151731b

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/ee206701a301180b1d499684fcd85e3ada3e8a01

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/ee206701a301180b1d499684fcd85e3ada3e8a01

internal/tiger/config/output.go
murrayju: Addressed in https://github.com/timescale/tiger-cli/pull/49/commits/43c43aea437fe49b79d9178b2df5bd25409073d0

internal/tiger/cmd/root.go
murrayju: Removed as a global flag: https://github.com/timescale/tiger-cli/pull/49/commits/43c43aea437fe49b79d9178b2df5bd25409073d0

I had issues with attempting to register individual flags with viper, it seemed to just not work. Was easy enough to check the cli flag in the cli command handler directly, fall back to the (global) config if not set.

internal/tiger/password/connection.go
murrayju: I'm indifferent. I'd rather not change here.

internal/tiger/cmd/config.go
```diff
@@ -133,68 +178,55 @@ func buildConfigCmd() *cobra.Command {
...
+	}
+	if cfg.ConsoleURL != "" {
+		table.Append("console_url", cfg.ConsoleURL)
+	}
+	if cfg.GatewayURL != "" {
+		table.Append("gateway_url", cfg.GatewayURL)
+	}
+	if cfg.DocsMCP {
+		table.Append("docs_mcp", fmt.Sprintf("%t", cfg.DocsMCP))
+	}
```
nathanjcochran: I think this is probably a bug. It looks like we only show `docs_mcp` in the output now if it's `true`. Same for `debug` and `analytics` below. I don't think that's right.

If I'm understanding correctly, the purpose of this change (only including fields in the output if they're not set to their zero value) is to support the `--no-defaults` case, where we only want to show the fields that are set in the user's `config.yaml` file. For most of the other fields, checking for the zero value is a decent way of checking whether the field is set in the user's config. However, it breaks down for booleans (and technically also if the user has set any of the string options to `""` in their config).

For example, imagine a user has set `docs_mcp: false` in their config (the default is `true`). Given this implementation, when they run `tiger config show --no-defaults`, they _won't_ see `docs_mcp` in the output at all.

I think this probably affects the JSON and YAML outputs too, for more-or-less the same reason (because of the `omitempty` YAML/JSON struct tags you added, since `false` is the empty/zero value for booleans).

I think this is one of those cases in Go where we _do_ need to distinguish between an omitted/`nil` value and a zero value (e.g. `false`). The normal way of doing that would be to make all of the `Config` fields pointers. But imo, that's going to cause major pain everywhere else we use the `Config`. So instead, we might want to make a new `ConfigOutput` struct, or something like that, which has all of the same fields but as pointers, that's just used for `tiger config show`.

internal/tiger/cmd/config_test.go
```diff
@@ -646,8 +665,9 @@ func TestConfigCommands_Integration(t *testing.T) {
 		t.Fatalf("Failed to show config after unset: %v", err)
 	}
 
+	result = make(map[string]interface{})
```
nathanjcochran: We should probably add something to `CLAUDE.md` about always using `any` instead of `interface{}`:

```suggestion
	result = make(map[string]any)
```

internal/tiger/cmd/root.go
nathanjcochran: > I had issues with attempting to register individual flags with viper, it seemed to just not work. Was easy enough to check the cli flag in the cli command handler directly, fall back to the (global) config if not set.

Hmm, weird. I would have expected that to work. You were binding them to `viper` outside of the command's `RunE` function, right? (because I think it's too late if it happens within `RunE`, since [this](https://github.com/timescale/tiger-cli/blob/83c89544e8608dbb9e064e7198fa9bc6b6e1de53/internal/tiger/cmd/root.go#L59-L68) has already been triggered at that point - I think it probably needs to happen right after the flags are defined in the command builder function).

In any case, the current approach is fine for now. Would be nice to get it using viper correctly, but can improve later.

internal/tiger/cmd/service.go
```diff
@@ -760,31 +797,35 @@ func prepareServiceForOutput(service api.Service, withPassword bool, output io.W
...
+	} else {
+		outputSvc.ConnectionDetails = *connectionDetails
+	}
+
+	if connectionString, err := password.BuildConnectionString(service, opts); err != nil {
+		if output != nil {
+			fmt.Fprintf(output, "⚠️  Warning: Failed to get connection string: %v\n", err)
+		}
+	} else {
+		outputSvc.ConnectionString = connectionString
```
nathanjcochran: Given that `BuildConnectionString` just calls `GetConnectionDetails` internally and then formats the result as a connection string, it's kind of weird to me that we're calling both of them here sequentially (it's duplicate work).

Here's a thought: get rid of `BuildConnectionString` altogether, and just make a `String()` function on the `ConnectionDetails` type that returns the connection string:

```go
func (d *ConnectionDetails) String() string {
	if d.Password == "" {
		// Build connection string without password (default behavior)
		// Password is handled separately via PGPASSWORD env var or ~/.pgpass file
		// This ensures credentials are never visible in process arguments
		return fmt.Sprintf("postgresql://%s@%s:%d/%s?sslmode=require", d.Role, d.Host, d.Port, d.Database)
	}

	// Include password in connection string
	return fmt.Sprintf("postgresql://%s:%s@%s:%d/%s?sslmode=require", d.Role, d.Password, d.Host, d.Port, d.Database)
}
```

That way, we'd just call `GetConnectionDetails` everywhere, and then call `connectionDetails.String()` to get the connection string:

```go
if connectionDetails, err := password.GetConnectionDetails(service, opts); err != nil {
	if output != nil {
		fmt.Fprintf(output, "⚠️  Warning: Failed to get connection details: %v\n", err)
	}
} else {
	outputSvc.ConnectionDetails = *connectionDetails
	outputSvc.ConnectionString = connectionDetails.String()
}
```

NOTE: code not tested - wrote by hand in the GitHub UI

internal/tiger/config/output.go
```diff
@@ -5,13 +5,18 @@ import (
 	"strings"
 )
 
-var ValidOutputFormats = []string{"json", "yaml", "table"}
+var validOutputFormats = []string{"json", "yaml", "table"}
+var validOutputFormatsWithEnv = append(append([]string(nil), validOutputFormats...), "env")
```
nathanjcochran: Haha wait, does this not work?


```suggestion
var validOutputFormatsWithEnv = append(validOutputFormats, "env")
```

I guess that could lead to both slices sharing the same backing array, but I don't think it should really matter here, since we're never mutating the slices 🤷‍♂️.

But if we're worried about that and want to be as safe as possible, I think this would be semantically clearer (need to `import "slices"`):

```suggestion
var validOutputFormatsWithEnv = append(slices.Clone(validOutputFormats), "env"))
```

internal/tiger/cmd/config_test.go
murrayju: Probably, but also I think GitHub copilot autocompleted this for me when I started typing (which I don't think reads that file)

internal/tiger/cmd/root.go
murrayju: > You were binding them to viper outside of the command's RunE function, right? 

Yes, had it immediately after registering the flag. Claude thought it is related to the order in which we initialize viper, but it wasn't able to find a solution. I do suspect we still aren't doing something quite right, but I think that can be a follow up effort.

internal/tiger/config/output.go
murrayju: Yeah, you are right. I clearly didn't know the basics of slices, and neither does Claude
image

murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/1f46c6ac89fbcbd4b987178ccf9ff8ac376bffa8

internal/tiger/cmd/service.go
murrayju: 👍 I like this

internal/tiger/cmd/config_test.go
nathanjcochran: Yeah, it's not a big deal. `interface{}` was the only way to do it for a long time, so it's probably all over the training data for these LLMs. It's still valid, but `any` is just newer and more idiomatic these days.

internal/tiger/config/output.go
nathanjcochran: Haha yeah, I think Claude probably suggested that as a way to ensure the new slice doesn't share the same backing array as the original (which can lead to subtle and confusing bugs if you're not careful). And Claude probably doesn't know about the `slices` package because it's relatively new (came out with generics). But in our case, we can just do the simple thing, because we know we're not going to be mutating either slice.

internal/tiger/cmd/service.go
murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/e4d41889d49d7d70801b7de69b73f4fa1299abc6

internal/tiger/cmd/config_test.go
murrayju: https://github.com/timescale/tiger-cli/pull/49/commits/43ee46bb5b2f2574ccd307cbe9eb08a998924734

There are plenty of other places in the code that do this, too. I wonder if we can add a linter rule or something?

internal/tiger/cmd/config.go
murrayju: Yeah, good call: https://github.com/timescale/tiger-cli/pull/49/commits/c6244cb17faf3d8032e7efa042f44e68055fa91e

PR #50, opened 2025-10-08
Update `service_create` and `service_show` to optionally return password

nathanjcochran:
See [Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1759528069142429) for background.

Updates the `service_create` and `service_show` MCP tools to function more like `tiger service create` and `tiger service describe`, following the changes made in #32.

Specifically, both MCP tools now return a `connection_string` field in the response. Additionally, they accept a new `with_password` parameter, which defaults to `false`. If set to `true`, a `password` field is returned in the response, and the password is also included in the returned connection string. This makes it possible for AI assistants to get access to a service's password and connection string, which was previously not possible via the MCP tools (which made it challenging for AI assistants to connect to a database or configure an application with database credentials).

I also tried to clean up and de-duplicate some repeated code.

Closes [AGE-205](https://linear.app/tigerdata/issue/AGE-205/tiger-cli-make-service-create-mcp-tool-return-connection-string-and)

### DIFF
--- a/internal/tiger/mcp/service_tools.go
+++ b/internal/tiger/mcp/service_tools.go
@@ -379,16 +379,6 @@ func (s *Server) handleServiceGet(ctx context.Context, req *mcp.CallToolRequest,
 		Service: s.convertToServiceDetail(service),
 	}
 
-	// Include password in ServiceDetail if requested
-	// Note: service_show doesn't have access to InitialPassword, so we fetch from storage
-	if input.WithPassword {
-		if passwd, err := password.GetPassword(service); err != nil {
-			logging.Debug("MCP: Failed to retrieve password from storage", zap.Error(err))
-		} else {
-			output.Service.Password = passwd
-		}
-	}
-
 	// Always include connection string in ServiceDetail
 	// Password is embedded in connection string only if with_password=true
 	if details, err := password.GetConnectionDetails(service, password.ConnectionDetailsOptions{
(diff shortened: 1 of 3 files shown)


internal/tiger/mcp/service_tools.go
```diff
@@ -79,14 +79,24 @@ func setServiceIDSchemaProperties(schema *jsonschema.Schema) {
...
+	schema.Properties["with_password"].Description = "Whether to include the password in the response and connection string."
+	schema.Properties["with_password"].Default = util.Must(json.Marshal(false))
+	schema.Properties["with_password"].Examples = []any{false, true}
+}
+
 // ServiceGetInput represents input for service_get
 type ServiceGetInput struct {
-	ServiceID string `json:"service_id"`
+	ServiceID    string `json:"service_id"`
+	WithPassword bool   `json:"with_password,omitempty"`
```
murrayju, line 92: What does `omitempty` do for us on an input type? Does it make it optional?

nathanjcochran: Yes. But this is really just a convention of the [github.com/google/jsonschema-go](https://pkg.go.dev/github.com/google/jsonschema-go/jsonschema) library that [github.com/modelcontextprotocol/go-sdk/mcp](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp) uses under the hood (for normal JSON unmarshalling, omitempty doesn't have any effect).

See [these docs](https://pkg.go.dev/github.com/google/jsonschema-go/jsonschema#For) for more info on how `jsonschema-go` generates JSON schemas from structs. In particular:

> Structs have schema type "object", and disallow additionalProperties. Their properties are derived from exported struct fields, using the struct field JSON name. Fields that are marked "omitempty" are considered optional; all other fields become required properties.

PR #53, opened 2025-10-10
Fix `service_create` status/connection string bug when waiting

nathanjcochran:
Fixes a small bug I didn't notice when I first implemented #50.

Specifically, when the `service_create` MCP tool is called with `with_password: true` and `wait: true`, the password and connection string weren't being returned in the output, because after we finished polling to check if the service is ready, we replaced the entire service struct (which we had already modified with the connection string/password).

This fixes it by making the MCP version of `waitForServiceReady` just return the latest service status, rather than the entire service. This makes it work more like the command version of `waitForServiceReady` (we should really consolidate them, but I didn't want to deal with the logging/output differences right now - can tackle it if/when we overhaul logging in general).

Also fixed a very small bug in the command version of `waitForServiceReady`. See comment below.

### DIFF
--- a/internal/tiger/cmd/service_test.go
+++ b/internal/tiger/cmd/service_test.go
@@ -1252,7 +1252,7 @@ func TestWaitForServiceReady_Timeout(t *testing.T) {
 	cmd.SetErr(errBuf)
 
 	// Test waitForServiceReady with very short timeout to trigger timeout quickly
-	_, err = waitForServiceReady(client, "test-project-123", "svc-12345", 100*time.Millisecond, cmd.ErrOrStderr())
+	_, err = waitForServiceReady(client, "test-project-123", "svc-12345", 100*time.Millisecond, nil, cmd.ErrOrStderr())
 
 	// Should return an error with ExitTimeout
 	if err == nil {


internal/tiger/cmd/service.go
```diff
@@ -841,14 +841,14 @@ func formatTimePtr(t *time.Time) string {
 }
 
 // waitForServiceReady polls the service status until it's ready or timeout occurs
-func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID string, waitTimeout time.Duration, output io.Writer) (*api.DeployStatus, error) {
+func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID string, waitTimeout time.Duration, initialStatus *api.DeployStatus, output io.Writer) (*api.DeployStatus, error) {
```
nathanjcochran, line 844: This fixes a small bug I noticed when in the command version of `waitForServiceReady`, where the `lastStatus` returned from the command would be `nil` if it timed out before the first poll. Unlikely in practice, but technically possible if the `--wait-timeout` is less than 10 seconds (our polling interval).

The fix is just to pass in the service's initial status (from the create/fork request), and set the initial value of `lastStatus` to that.

internal/tiger/mcp/service_tools.go
```diff
@@ -558,10 +557,10 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
 	if input.Wait {
 		timeout := time.Duration(*input.TimeoutMinutes) * time.Minute
 
-		output.Service, err = s.waitForServiceReady(apiClient, cfg.ProjectID, serviceID, timeout, serviceStatus)
-		if err != nil {
+		if status, err := s.waitForServiceReady(apiClient, cfg.ProjectID, serviceID, timeout, service.Status); err != nil {
 			output.Message = fmt.Sprintf("Error: %s", err.Error())
 		} else {
+			output.Service.Status = util.DerefStr(status)
```
nathanjcochran, line 563: This is the meat of the fix. Rather than overwriting `output.Service` with the latest polling result, we just overwrite `output.Service.Status` now (ensuring we don't overwrite the connection string/password set above).

PR #54, opened 2025-10-10
AGE-120: display message when new version available

murrayju:
This pulls the latest release version number from the (configurable) S3 bucket url, compares it to the current version, and outputs a message when outdated. This is performed at most once per hour (also configurable), with the last check time stored in the config file.

We attempt to determine the installation method based on the current binary's path, and output a helpful message for _how_ to upgrade as well (e.g. `brew upgrade tiger-cli`).

I made `tiger version` skip this implicit check, instead adding an explicit `tiger version --check` that forces checking (regardless of last check time).

image

Finally, I also implemented some `--output` options for the `version` command, to conform to table formatting by default, with `json`, `yaml`, and `bare` (_only_ the version number) to making parsing for scripting tasks easier.

image

Includes the new version when `--check` is used
image

`-o bare` will be easiest to just capture the version number
image

### DIFF
--- a/internal/tiger/cmd/config_test.go
+++ b/internal/tiger/cmd/config_test.go
@@ -7,6 +7,7 @@ import (
 	"slices"
 	"strings"
 	"testing"
+	"time"
 
 	"gopkg.in/yaml.v3"
 
@@ -106,6 +107,8 @@ password_storage: pgpass
 func TestConfigShow_JSONOutput(t *testing.T) {
 	tmpDir, _ := setupConfigTest(t)
 
+	now := time.Now()
+
 	// Create config file with JSON output format
 	configContent := `api_url: https://json.api.com/v1
 project_id: json-project
@@ -113,7 +116,8 @@ output: json
 analytics: true
 password_storage: none
 version_check_interval: 1h
-`
+version_check_last_time: ` + now.Format(time.RFC3339) + "\n"
+
 	configFile := config.GetConfigFile(tmpDir)
 	if err := os.WriteFile(configFile, []byte(configContent), 0644); err != nil {
 		t.Fatalf("Failed to write config file: %v", err)
@@ -132,20 +136,21 @@ version_check_interval: 1h
 
 	// Verify ALL JSON keys and their expected values
 	expectedValues := map[string]interface{}{
-		"api_url":                "https://json.api.com/v1",
-		"console_url":            "https://console.cloud.timescale.com",
-		"gateway_url":            "https://console.cloud.timescale.com/api",
-		"docs_mcp":               true,
-		"docs_mcp_url":           "https://mcp.tigerdata.com/docs",
-		"project_id":             "json-project",
-		"service_id":             "",
-		"output":                 "json",
-		"analytics":              true,
-		"password_storage":       "none",
-		"debug":                  false,
-		"config_dir":             tmpDir,
-		"releases_url":           "https://cli.tigerdata.com",
-		"version_check_interval": float64(3600000000000), // JSON unmarshals time.Duration as nanoseconds (1 hour = 3600000000000ns)
+		"api_url":                 "https://json.api.com/v1",
+		"console_url":             "https://console.cloud.timescale.com",
+		"gateway_url":             "https://console.cloud.timescale.com/api",
+		"docs_mcp":                true,
+		"docs_mcp_url":            "https://mcp.tigerdata.com/docs",
+		"project_id":              "json-project",
+		"service_id":              "",
+		"output":                  "json",
+		"analytics":               true,
+		"password_storage":        "none",
+		"debug":                   false,
+		"config_dir":              tmpDir,
+		"releases_url":            "https://cli.tigerdata.com",
+		"version_check_interval":  float64(3600000000000), // JSON unmarshals time.Duration as nanoseconds (1 hour = 3600000000000ns)
+		"version_check_last_time": now.Format(time.RFC3339),
 	}
 
 	for key, expectedValue := range expectedValues {
@@ -155,7 +160,6 @@ version_check_interval: 1h
 	}
 
 	// Ensure no extra keys are present
-	delete(result, "version_check_last_time") // this always changes
 	if len(result) != len(expectedValues) {
 		t.Errorf("Expected %d keys in JSON output, got %d", len(expectedValues), len(result))
 	}
@@ -164,13 +168,16 @@ version_check_interval: 1h
 func TestConfigShow_YAMLOutput(t *testing.T) {
 	tmpDir, _ := setupConfigTest(t)
 
+	now := time.Now()
+
 	// Create config file with YAML output format
 	configContent := `api_url: https://yaml.api.com/v1
 project_id: yaml-project
 output: yaml
 analytics: false
 password_storage: keyring
-`
+version_check_last_time: ` + now.Format(time.RFC3339) + "\n"
+
 	configFile := config.GetConfigFile(tmpDir)
 	if err := os.WriteFile(configFile, []byte(configContent), 0644); err != nil {
 		t.Fatalf("Failed to write config file: %v", err)
@@ -182,37 +189,46 @@ password_storage: keyring
 	}
 
 	// Parse YAML output
-	var result map[string]interface{}
+	var result map[string]any
 	if err := yaml.Unmarshal([]byte(output), &result); err != nil {
 		t.Fatalf("Failed to parse YAML output: %v", err)
 	}
 
 	// Verify ALL YAML keys and their expected values
-	expectedValues := map[string]interface{}{
-		"api_url":                "https://yaml.api.com/v1",
-		"console_url":            "https://console.cloud.timescale.com",
-		"gateway_url":            "https://console.cloud.timescale.com/api",
-		"docs_mcp":               true,
-		"docs_mcp_url":           "https://mcp.tigerdata.com/docs",
-		"project_id":             "yaml-project",
-		"service_id":             "",
-		"output":                 "yaml",
-		"analytics":              false,
-		"password_storage":       "keyring",
-		"debug":                  false,
-		"config_dir":             tmpDir,
-		"releases_url":           "https://cli.tigerdata.com",
-		"version_check_interval": "24h0m0s", // YAML serializes time.Duration as string
+	expectedValues := map[string]any{
+		"api_url":                 "https://yaml.api.com/v1",
+		"console_url":             "https://console.cloud.timescale.com",
+		"gateway_url":             "https://console.cloud.timescale.com/api",
+		"docs_mcp":                true,
+		"docs_mcp_url":            "https://mcp.tigerdata.com/docs",
+		"project_id":              "yaml-project",
+		"service_id":              "",
+		"output":                  "yaml",
+		"analytics":               false,
+		"password_storage":        "keyring",
+		"debug":                   false,
+		"config_dir":              tmpDir,
+		"releases_url":            "https://cli.tigerdata.com",
+		"version_check_interval":  "24h0m0s", // YAML serializes time.Duration as string
+		"version_check_last_time": now,
 	}
 
 	for key, expectedValue := range expectedValues {
-		if result[key] != expectedValue {
-			t.Errorf("Expected %s '%v', got %v", key, expectedValue, result[key])
+		switch expectedValue.(type) {
+		case time.Time:
+			// YAML unmarshals time.Time as time.Time type, so we need to compare differently
+			if expectedValue.(time.Time).Format(time.RFC3339) != result[key].(time.Time).Format(time.RFC3339) {
+				t.Errorf("foo Expected %s '%v', got %v", key, expectedValue, result[key])
+			}
+		default:
+			// Other types can be compared directly
+			if result[key] != expectedValue {
+				t.Errorf("Expected %s '%v', got %v", key, expectedValue, result[key])
+			}
 		}
 	}
 
 	// Ensure no extra keys are present
-	delete(result, "version_check_last_time") // this always changes
 	if len(result) != len(expectedValues) {
 		t.Errorf("Expected %d keys in YAML output, got %d", len(expectedValues), len(result))
 	}


internal/tiger/cmd/config.go
```diff
@@ -216,6 +217,15 @@ func outputTable(cfg *config.ConfigOutput, cmd *cobra.Command) error {
...
 		table.Append("config_dir", *cfg.ConfigDir)
 	}
+	if cfg.VersionCheckURL != nil {
+		table.Append("version_check_url", *cfg.VersionCheckURL)
+	}
+	if cfg.VersionCheckInterval != nil {
+		table.Append("version_check_interval", fmt.Sprintf("%d seconds", *cfg.VersionCheckInterval))
+	}
+	if cfg.VersionCheckLastTime != nil {
+		table.Append("version_check_last_time", time.Unix(int64(*cfg.VersionCheckLastTime), 0).Format(time.RFC1123))
```
murrayju: I figured it'd be nice for the table output format to be more human readable. Open to a different formatting choice, or just the raw number if people think that's better.

Image

internal/tiger/cmd/root.go
```diff
@@ -33,14 +35,13 @@ tiger auth login
...
-			}
-
 			cfg, err := config.Load()
 			if err != nil {
-				logging.Error("failed to load config", zap.Error(err))
-				return err
+				return fmt.Errorf("failed to load config: %w", err)
+			}
+
+			if err := logging.Init(cfg.Debug); err != nil {
```
murrayju, line 43: I fixed this unrelated bug, where `debug` from the config file/env was not being used (only the cli flag worked)

scripts/install.sh
```diff
@@ -381,7 +381,7 @@ verify_installation() {
 
     # Test that the binary is executable and get version
     local installed_version
-    if installed_version=$("${binary_path}" version 2>/dev/null | head -n1 || echo ""); then
+    if installed_version=$("${binary_path}" version -o bare 2>/dev/null | head -n1 || echo ""); then
```
murrayju: To fix this:

Image

internal/tiger/cmd/root.go
```diff
@@ -51,8 +52,22 @@ tiger auth login
...
 		},
-		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			cfg, err := config.Load()
+			if err != nil {
+				return fmt.Errorf("failed to load config: %w", err)
+			}
+
+			// Check for updates on any command
+			// Skip check if running the version command (it handles via --check flag)
```
cevian: don't feel strongly but why not just reuse --skip-update-check on `version` seems counterintuitive to check that less often than other commands

internal/tiger/cmd/version.go
```diff
@@ -1,24 +1,100 @@
...
+
+			output := cmd.OutOrStdout()
+			switch outputFormat {
+			case "json":
+				encoder := json.NewEncoder(output)
+				encoder.SetIndent("", "  ")
+				return encoder.Encode(versionOutput)
+			case "yaml":
+				encoder := yaml.NewEncoder(output)
+				encoder.SetIndent(2)
```
cevian: do we set this indent consistently across yaml? if so let's make a linear ticket to unify into util func

internal/tiger/config/config.go
```diff
@@ -357,6 +368,55 @@ func (c *Config) updateField(key string, value any) (any, error) {
...
+
+	case "version_check_interval":
+		switch v := value.(type) {
+		case int:
+			if v < 0 {
+				return nil, fmt.Errorf("version_check_interval must be non-negative (0 to disable)")
+			}
+			c.VersionCheckInterval = v
+			validated = v
+		case string:
```
cevian, line 388: I am pretty sure there is a standard golang way to parse string intervals. we should use that

internal/tiger/config/config.go
```diff
@@ -357,6 +368,55 @@ func (c *Config) updateField(key string, value any) (any, error) {
...
+
+	case "version_check_last_time":
+		switch v := value.(type) {
+		case int64:
+			c.VersionCheckLastTime = v
+			validated = v
+		case int:
+			c.VersionCheckLastTime = int64(v)
+			validated = int64(v)
+		case string:
```
cevian, line 412: again golang has ways to parse date/times and we should use that

internal/tiger/config/config.go
```diff
@@ -15,60 +15,71 @@ import (
...
+	ProjectID            string `mapstructure:"project_id" yaml:"project_id"`
+	ServiceID            string `mapstructure:"service_id" yaml:"service_id"`
+	Output               string `mapstructure:"output" yaml:"output"`
+	Analytics            bool   `mapstructure:"analytics" yaml:"analytics"`
+	PasswordStorage      string `mapstructure:"password_storage" yaml:"password_storage"`
+	Debug                bool   `mapstructure:"debug" yaml:"debug"`
+	ConfigDir            string `mapstructure:"config_dir" yaml:"-"`
+	VersionCheckURL      string `mapstructure:"version_check_url" yaml:"version_check_url"`
+	VersionCheckInterval int    `mapstructure:"version_check_interval" yaml:"version_check_interval"`
+	VersionCheckLastTime int64  `mapstructure:"version_check_last_time" yaml:"version_check_last_time"`
```
cevian: This cannot be a time.Time()? and the interval can't be a time.Duration

internal/tiger/config/config.go
```diff
@@ -15,60 +15,71 @@ import (
...
+	ServiceID            *string `mapstructure:"service_id" json:"service_id,omitempty" yaml:"service_id,omitempty"`
+	Output               *string `mapstructure:"output" json:"output,omitempty" yaml:"output,omitempty"`
+	Analytics            *bool   `mapstructure:"analytics" json:"analytics,omitempty" yaml:"analytics,omitempty"`
+	PasswordStorage      *string `mapstructure:"password_storage" json:"password_storage,omitempty" yaml:"password_storage,omitempty"`
+	Debug                *bool   `mapstructure:"debug" json:"debug,omitempty" yaml:"debug,omitempty"`
+	ConfigDir            *string `mapstructure:"config_dir" json:"config_dir,omitempty" yaml:"config_dir,omitempty"`
+	VersionCheckURL      *string `mapstructure:"version_check_url" json:"version_check_url,omitempty" yaml:"version_check_url,omitempty"`
+	VersionCheckInterval *int    `mapstructure:"version_check_interval" json:"version_check_interval,omitempty" yaml:"version_check_interval,omitempty"`
+	VersionCheckLastTime *int64  `mapstructure:"version_check_last_time" json:"version_check_last_time,omitempty" yaml:"version_check_last_time,omitempty"`
```
cevian: again lets use golang time types here if possible?

internal/tiger/config/config.go
```diff
@@ -15,60 +15,71 @@ import (
...
+	DefaultConsoleURL           = "https://console.cloud.timescale.com"
+	DefaultGatewayURL           = "https://console.cloud.timescale.com/api"
+	DefaultDocsMCP              = true
+	DefaultDocsMCPURL           = "https://mcp.tigerdata.com/docs"
+	DefaultOutput               = "table"
+	DefaultAnalytics            = true
+	DefaultPasswordStorage      = "keyring"
+	DefaultDebug                = false
+	DefaultVersionCheckURL      = "https://tiger-cli-releases.s3.amazonaws.com/install/latest.txt"
+	DefaultVersionCheckInterval = 3600 // 1 hour in seconds (0 to disable)
```
cevian: I would set the interval to a time.Hour*24

internal/tiger/version/check.go
```diff
@@ -0,0 +1,287 @@
...
+
+// DetectInstallMethod determines how Tiger CLI was installed
+func detectInstallMethod(binaryPath string) InstallMethod {
+	// Check for development build
+	if strings.Contains(binaryPath, "go-build") {
+		return InstallMethodDevelopment
+	}
+
+	// Check for Homebrew installation
+	if strings.Contains(binaryPath, "/homebrew/") ||
```
cevian: need to support `/usr/local/Cellar/` too

internal/tiger/version/check.go
```diff
@@ -0,0 +1,287 @@
...
+	vLatest, err := semver.NewVersion(latest)
+	if err != nil {
+		return false
+	}
+
+	return vLatest.GreaterThan(vCurrent)
+}
+
+// DetectInstallMethod determines how Tiger CLI was installed
+func detectInstallMethod(binaryPath string) InstallMethod {
```
cevian, line 132: dislike this but don't know if there is a better way. We could run `tiger config set install_method deb` and similar in post-install scripts. but that has its own complexity (which I think is better but wont inisist).

internal/tiger/version/check.go
```diff
@@ -0,0 +1,287 @@
...
+	case InstallMethodDeb:
+		return "sudo apt update && sudo apt install tiger-cli"
+	case InstallMethodRPM:
+		// Try to detect which package manager is available
+		if _, err := exec.LookPath("dnf"); err == nil {
+			return "sudo dnf update tiger-cli"
+		}
+		return "sudo yum update tiger-cli"
+	case InstallMethodInstallSh:
+		return "curl -fsSL https://tiger-cli-releases.s3.amazonaws.com/install/install.sh | sh"
```
cevian: will have to remember to replace when changing url (Not to self)

internal/tiger/cmd/root.go
murrayju: Claude suggested this, and I thought maybe it made sense where you often check the version in scripts, and might not want the update check.

But, I do agree it would be simpler to just be consistent, and we have the flags to control the behavior as needed.

internal/tiger/cmd/version.go
murrayju: Yes, it is in a few places now. I'll just add a helper now.

internal/tiger/config/config.go
murrayju: Do you mean that you would only want to show the message at most once per day?

internal/tiger/version/check.go
murrayju: The GitHub cli uses [detection](https://github.com/cli/cli/blob/trunk/internal/ghcmd/cmd.go#L169) (similar to this), but they also only bother to detect homebrew.

I do think good detection logic is going to be more reliable than writing to the config file, which can easily be altered/reset. Either way, it is easy to think of ways to break it... we could bail on attempts to be helpful, and instead show a more generic "update according to your install method" message (or no message at all).

This is again a case of code written by Claude, that I tested and appears to work well. I'm not convinced that there is a good reason to rework it a different way.

internal/tiger/version/check.go
murrayju: This helps, but I think we'll also change the subpath
https://github.com/timescale/tiger-cli/pull/54/commits/7b25e54e8e8f95ff10dee9f6aa9a66011d2506c8

internal/tiger/cmd/root.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/13be93cf78a514fa35edf502f9e22529ec05df6c

internal/tiger/cmd/version.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/042735a984be10d43eb5920fa8d1dda625889e4b

internal/tiger/version/check.go
murrayju: Borrowed some code from the GH cli to use `brew --prefix` for better detection, regardless of install dir.

https://github.com/timescale/tiger-cli/pull/54/commits/b30627bf04c1c62f258d208e928fb3f8cf3f2056

internal/tiger/cmd/root.go
murrayju: Oh, another reason was to avoid checking twice when doing the explicit check. So, handling that slightly different from before:

https://github.com/timescale/tiger-cli/pull/54/commits/76fd8e275468399051f61e35d6dde58b9a956377

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/10c1ec64c356d2d83550a8b564b776287fabc235

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/10c1ec64c356d2d83550a8b564b776287fabc235

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/10c1ec64c356d2d83550a8b564b776287fabc235

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/10c1ec64c356d2d83550a8b564b776287fabc235

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/e9237d08cbbcf71811ae447cb792bde2eba0a2ca

internal/tiger/cmd/config.go
```diff
@@ -216,17 +218,14 @@ func outputTable(cfg *config.ConfigOutput, cmd *cobra.Command) error {
...
 	}
+	if cfg.ReleasesURL != nil {
+		table.Append("releases_url", *cfg.ReleasesURL)
+	}
+	if cfg.VersionCheckInterval != nil {
+		table.Append("version_check_interval", cfg.VersionCheckInterval.String())
+	}
+	if cfg.VersionCheckLastTime != nil {
+		table.Append("version_check_last_time", cfg.VersionCheckLastTime.Format(time.RFC1123))
+	}
```
nathanjcochran, line 229: This is not super important, but I think we should consider organizing the order in which these config values are output. Either grouping them (so that all of the URLs are together, for instance), or maybe just alphabetizing them? Doesn't have to happen in this PR, just something that occurred to me.

internal/tiger/cmd/config_test.go
```diff
@@ -131,18 +132,20 @@ password_storage: none
...
+		"docs_mcp_url":           "https://mcp.tigerdata.com/docs",
+		"project_id":             "json-project",
+		"service_id":             "",
+		"output":                 "json",
+		"analytics":              true,
+		"password_storage":       "none",
+		"debug":                  false,
+		"config_dir":             tmpDir,
+		"releases_url":           "https://cli.tigerdata.com",
+		"version_check_interval": float64(3600000000000), // JSON unmarshals time.Duration as nanoseconds (1 hour = 3600000000000ns)
```
nathanjcochran: This test doesn't appear to check for the presence of `version_check_last_time`. Might be good to include (I think all of the other config options are included here 🤔). Same for the other tests below.

If the problem is that `version_check_last_time` is non-deterministic, we might want to stub out `time.Now` and mock it in these tests. That's usually done like this:

1. Do this in the code itself, and call `now()` instead of `time.Now()`

```
var now = time.Now
```

2. Do this to mock it in the tests (and don't forget to reset it when the test is finished, either via a `defer` call or [T.Cleanup](https://pkg.go.dev/testing#T.Cleanup)):
```
now = func() time.Time {
    return time.Date(2025, time.October, 14, 12, 0, 0, 0, time.UTC)
}
```

internal/tiger/cmd/version.go
```diff
@@ -2,23 +2,89 @@ package cmd
...
+			case "bare":
+				fmt.Fprintln(output, versionOutput.Version)
+				return nil
+			default:
+				return outputVersionTable(output, versionOutput)
+			}
 		},
 	}
+
+	cmd.Flags().BoolVar(&checkVersion, "check", false, "Check for the latest version")
```
cevian: I still don't get why we don't just use the existing global. but don't feel too strongly

internal/tiger/config/config.go
```diff
@@ -357,6 +369,59 @@ func (c *Config) updateField(key string, value any) (any, error) {
...
+			t, err := time.Parse(time.RFC3339, v)
+			if err != nil {
+				// Try parsing as unix timestamp
+				i, err := strconv.ParseInt(v, 10, 64)
+				if err != nil {
+					return nil, fmt.Errorf("invalid version_check_last_time value: %s (must be RFC3339 timestamp or unix timestamp)", v)
+				}
+				t = time.Unix(i, 0)
+			}
+			c.VersionCheckLastTime = t
```
cevian, line 426: I would add a sanity check that it's before now() or before now() and a day

cevian (approved):
a few optional comments

internal/tiger/cmd/version.go
murrayju: This is to force a check, regardless of last check time. The global is `--skip-update-check`. Maybe you are thinking `--skip-update-check=false`? That doesn't quite seem clear enough to me, plus it is much more to type.

murrayju: This should help: https://github.com/timescale/tiger-cli/pull/54/commits/f322c168c0b33356fc287c8fa2146b3028633a74

internal/tiger/config/config.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/cd05cc90d0acb4621c4f807f466915681631ad28

internal/tiger/cmd/config.go
murrayju: The yaml/json output is in the order as defined in the struct. That was _nearly_ alphabetical already, so I abc sorted those. Then I sorted the table output to match.

https://github.com/timescale/tiger-cli/pull/54/commits/c60d7aaf20e730f081b175accac0940c63ffc4d7

image

internal/tiger/cmd/config_test.go
murrayju: https://github.com/timescale/tiger-cli/pull/54/commits/13d73c6227c23ba1ceca624eaaf1445c48cf1419

PR #55, opened 2025-10-13
Remove `--free`, implement `shared` CPU/memory and automatic defaults

nathanjcochran:
See [Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1759958456029109?thread_ts=1759768841.513939&cid=C099S47CSRX) and [AGE-208](https://linear.app/tigerdata/issue/AGE-208/tiger-cli-remove-free-support-shared-cpumemory-add-auto-options) for background.

This implements the changes in [AGE-208](https://linear.app/tigerdata/issue/AGE-208/tiger-cli-remove-free-support-shared-cpumemory-add-auto-options), with one relatively minor difference: I did not implement explicit "auto" options for each flag. Instead, I just removed the default value for each flag, and am interpreting the zero value (i.e. when the flag is omitted) to mean "auto". This was just a little easier to implement as a first pass, as it meant that I didn't need to update as much of the `--help` text to include references to "auto" or explain what it means. I think it would be easy to add explicit "auto" options as a follow-up, if we're sure we want it.

This PR:
- Gets rid of the `--free` flag.
- Adds support for specifying `shared` for `--cpu` and `--memory` (this is how you create a free services now).
- Gets rid of the explicit defaults for `--addons`, `--region`, `--cpu`, and `--memory`
   - These parameters are no longer sent in the API request if they aren't specified.
   - The default values are now determined in the API layer.
   - The `--cpu` and `--memory` defaults now depend on what type of plan you're on (the free plan defaults to `shared` CPU/memory, whereas paid plans default to `0.5 CPU/ 2 GB`.
   - The `--addons` and `--region` defaults depend on what type of service you're creating.
      - If you're creating a free service (shared CPU/memory), `--addons` defaults to `time-series,ai`, whereas if you're creating a paid service, it defaults to `time-series`.
      - `--region` currently always defaults to `us-east-1`, but we intend to have it intelligently choose the best region for paid services in the future.
- Makes corresponding changes to the `service_create` MCP tool.

### How to test:

- Check out https://github.com/timescale/savannah-public/pull/51, build and run it: `go install ./... && server`
- Check out this branch and build it: `go install ./...`
- Configure your CLI to point to your local `savannah-public` server: `tiger config set api_url http://localhost:8080/public/api/v1`
- Create a service: `tiger service create` (default service type depends on your plan)
- Create a free service: `tiger service create --cpu=shared`

### Related PRs:

- https://github.com/timescale/timescale-graphql-client/pull/8
- https://github.com/timescale/savannah-public/pull/51

### DIFF
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -244,6 +244,10 @@ func buildServiceCreateCmd() *cobra.Command {
 		Short: "Create a new database service",
 		Long: `Create a new database service in the current project.
 
+The default type of service created depends on your plan:
+- Free plan: Creates a service with shared CPU/memory and the 'time-series' and 'ai' add-ons
+- Paid plans: Creates a service with 0.5 CPU / 2 GB memory and the 'time-series' add-on
+
 By default, the newly created service will be set as your default service for future
 commands. Use --no-set-default to prevent this behavior.
 
--- a/internal/tiger/mcp/service_tools.go
+++ b/internal/tiger/mcp/service_tools.go
@@ -259,6 +259,10 @@ func (s *Server) registerServiceTools() {
 		Title: "Create Database Service",
 		Description: `Create a new database service in TigerData Cloud with specified type, compute resources, region, and HA options.
 
+The default type of service created depends on the user's plan:
+- Free plan: Creates a service with shared CPU/memory and the 'time-series' and 'ai' add-ons
+- Paid plans: Creates a service with 0.5 CPU / 2 GB memory and the 'time-series' add-on
+
 Default behavior: Returns immediately while service provisions in background (recommended).
 Setting wait=true will block for a few minutes until ready - only use if user explicitly needs immediate access.
 timeout_minutes: Wait duration in minutes (only relevant with wait=true).


internal/tiger/cmd/service.go
```diff
@@ -253,7 +252,7 @@ Examples:
   tiger service create
 
   # Create a free TimescaleDB service
-  tiger service create --name free-db --free
+  tiger service create --name free-db --cpu shared --memory shared
```
Askir:
```suggestion
  tiger service create --name free-db --cpu shared
```
Better to give an example without redundant params no?

internal/tiger/cmd/service.go
```diff
@@ -463,15 +426,14 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
...
 	// Add flags
 	cmd.Flags().StringVar(&createServiceName, "name", "", "Service name (auto-generated if not provided)")
-	cmd.Flags().StringSliceVar(&createAddons, "addons", []string{util.AddonTimeSeries}, fmt.Sprintf("Addons to enable (%s, or 'none' for PostgreSQL-only)", strings.Join(util.ValidAddons(), ", ")))
-	cmd.Flags().StringVar(&createRegionCode, "region", "us-east-1", "Region code")
-	cmd.Flags().IntVar(&createCpuMillis, "cpu", 500, "CPU allocation in millicores")
-	cmd.Flags().IntVar(&createMemoryGbs, "memory", 2, "Memory allocation in gigabytes")
+	cmd.Flags().StringSliceVar(&createAddons, "addons", nil, fmt.Sprintf("Addons to enable (%s, or 'none' for PostgreSQL-only)", strings.Join(util.ValidAddons(), ", ")))
+	cmd.Flags().StringVar(&createRegionCode, "region", "", "Region code")
+	cmd.Flags().StringVar(&createCpuMillis, "cpu", "", "CPU allocation in millicores or 'shared'")
+	cmd.Flags().StringVar(&createMemoryGBs, "memory", "", "Memory allocation in gigabytes or 'shared'")
```
Askir, line 436: I think this is a bit confusing with the default behavior not being mentioned?

internal/tiger/cmd/service.go
```diff
@@ -282,12 +281,9 @@ Examples:
...
 
-Free Tier:
-  When using --free, resource flags (--cpu, --memory, --replicas, --addons, --region) cannot be specified
-
-Allowed CPU/Memory Configurations (non-free services):
-  0.5 CPU (500m) / 2GB    |  1 CPU (1000m) / 4GB    |  2 CPU (2000m) / 8GB    |  4 CPU (4000m) / 16GB
-  8 CPU (8000m) / 32GB    |  16 CPU (16000m) / 64GB  |  32 CPU (32000m) / 128GB
+Allowed CPU/Memory Configurations:
+  shared / shared       |  0.5 CPU (500m) / 2GB    |  1 CPU (1000m) / 4GB     |  2 CPU (2000m) / 8GB
+  4 CPU (4000m) / 16GB  |  8 CPU (8000m) / 32GB    |  16 CPU (16000m) / 64GB  |  32 CPU (32000m) / 128GB
```
Askir, line 290: I think at least here we should document the default behavior. I'd actually prefer to have auto/auto as a specific option but even if we don't have it we should explain what not providing anything does.

internal/tiger/cmd/service.go
nathanjcochran: Good call. Done here: https://github.com/timescale/tiger-cli/pull/55/commits/8f9a6d1184ee4c80152bdeba7574b1f0963d6de3

internal/tiger/cmd/service.go
nathanjcochran: I agree, but I think it's going to be hard to fit an explanation of the default behavior into the flag descriptions (whereas previously, it was easy to just list a default value). We can list `auto` as the default in a follow-up PR if we think that's clearer (I think that change should be backwards-compatible), but I still don't think it would really explain much.

So I think it needs to go in the main `--help` text. How does this look for now (at least as a first pass - can definitely improve later)? https://github.com/timescale/tiger-cli/pull/55/commits/346aae9bbbfc4e4c68a6ad4ad73c0c371181e82a

internal/tiger/cmd/service.go
nathanjcochran: Yeah, agree. See my other comment above.

PR #58, opened 2025-10-15
Add spec for `tiger db create role` command 

cevian:
## Summary
- Adds comprehensive specification for `tiger db create role` command
- Documents read-only role creation for safe AI agent database access
- Defines command structure, options, and behavior

## Key Features Documented
- **Role creation**: Optional positional service ID, required `--name` flag
- **Read-only enforcement**: Permanent role-based read-only via `tsdb_admin.read_only_role`
- **Grant inheritance**: `--from` flag accepts comma-separated list of roles
- **Performance safety**: `--statement-timeout` for query limits
- **Password options**: Auto-generate, explicit value, `TIGER_NEW_PASSWORD` env var, or interactive prompt

## Design Decisions
- `--name` flag is required (not positional) to prevent accidental role creation with service IDs
- Service ID is optional positional parameter (uses default service if omitted)
- Password auto-generation is default behavior for better security
- Read-only mode uses role-level enforcement that cannot be bypassed

## Documentation Includes
- Multiple usage examples covering common scenarios
- Detailed explanation of read-only mode implementation
- Use cases for AI agents and automated tools
- Safety patterns combining read-only + statement timeouts
- Password behavior precedence (flag > env var > auto-generate)

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -427,6 +427,17 @@ func createRoleWithOptions(ctx context.Context, conn *pgx.Conn, roleName, rolePa
 			roleName); err != nil {
 			return fmt.Errorf("failed to grant tsdbadmin privileges: %w", err)
 		}
+
+		// Grant any other roles (besides tsdbadmin) if specified
+		// This is necessary because the special functions don't support IN ROLE clause
+		for _, role := range otherRoles {
+			grantSQL := fmt.Sprintf("GRANT %s TO %s",
+				pgx.Identifier{role}.Sanitize(),
+				pgx.Identifier{roleName}.Sanitize())
+			if _, err := tx.Exec(ctx, grantSQL); err != nil {
+				return fmt.Errorf("failed to grant role %s: %w", role, err)
+			}
+		}
 	} else {
 		// Use standard CREATE ROLE for non-tsdbadmin cases
 		// Fail if password contains a single quote (we don't support escaping)
@@ -435,6 +446,7 @@ func createRoleWithOptions(ctx context.Context, conn *pgx.Conn, roleName, rolePa
 		}
 		// Wrap password in single quotes for SQL literal
 		quotedPassword := "'" + rolePassword + "'"
+		// IN ROLE clause handles all role grants, so no need for separate GRANT statements
 		createSQL := buildCreateRoleSQL(roleName, quotedPassword, fromRoles)
 		if _, err := tx.Exec(ctx, createSQL); err != nil {
 			return fmt.Errorf("failed to create role: %w", err)
@@ -446,16 +458,6 @@ func createRoleWithOptions(ctx context.Context, conn *pgx.Conn, roleName, rolePa
 			if _, err := tx.Exec(ctx, alterSQL); err != nil {
 				return fmt.Errorf("failed to configure read-only mode: %w", err)
 			}
-		}
-	}
-
-	// Grant any other roles (besides tsdbadmin) if specified
-	for _, role := range otherRoles {
-		grantSQL := fmt.Sprintf("GRANT %s TO %s",
-			pgx.Identifier{role}.Sanitize(),
-			pgx.Identifier{roleName}.Sanitize())
-		if _, err := tx.Exec(ctx, grantSQL); err != nil {
-			return fmt.Errorf("failed to grant role %s: %w", role, err)
 		}
 	}
 
@@ -563,7 +565,7 @@ func outputCreateRoleResult(cmd *cobra.Command, roleName string, readOnly bool, 
 func buildDbCreateRoleCmd() *cobra.Command {
 	var roleName string
 	var readOnly bool
-	var fromRoles string
+	var fromRoles []string
 	var statementTimeout time.Duration
 	var passwordFlag string
 	var output string
@@ -603,7 +605,7 @@ Examples:
   tiger db create role --name ai_analyst --read-only --from app_role
 
   # Create a read-only role inheriting from multiple roles
-  tiger db create role --name ai_analyst --read-only --from app_role,readonly_role
+  tiger db create role --name ai_analyst --read-only --from app_role --from readonly_role
 
   # Create a read-only role with statement timeout
   tiger db create role --name ai_analyst --read-only --statement-timeout 30s
@@ -612,8 +614,7 @@ Examples:
   tiger db create role --name ai_analyst --password=my-secure-password
 
   # Create a role with password from environment variable
-  export TIGER_NEW_PASSWORD=my-secure-password
-  tiger db create role --name ai_analyst
+  TIGER_NEW_PASSWORD=my-secure-password tiger db create role --name ai_analyst
 
 Technical Details:
 This command executes PostgreSQL statements in a transaction to create and configure the role.
@@ -623,7 +624,7 @@ CREATE ROLE Options Used:
   - PASSWORD: Always set (from flag, env var, or auto-generated)
   - IN ROLE: Added when --from flag is provided to inherit grants from existing roles
 
-GUCs (PostgreSQL Configuration Parameters) That May Be Set:
+PostgreSQL Configuration Parameters That May Be Set:
   - tsdb_admin.read_only_role: Set to 'true' when --read-only flag is used
     (enforces permanent read-only mode for the role)
   - statement_timeout: Set when --statement-timeout flag is provided
@@ -679,19 +680,8 @@ GUCs (PostgreSQL Configuration Parameters) That May Be Set:
 			}
 			defer conn.Close(ctx)
 
-			// Parse from roles
-			var fromRolesList []string
-			if fromRoles != "" {
-				for _, role := range strings.Split(fromRoles, ",") {
-					role = strings.TrimSpace(role)
-					if role != "" {
-						fromRolesList = append(fromRolesList, role)
-					}
-				}
-			}
-
 			// Create the role with all options in a transaction
-			if err := createRoleWithOptions(ctx, conn, roleName, rolePassword, readOnly, statementTimeout, fromRolesList); err != nil {
+			if err := createRoleWithOptions(ctx, conn, roleName, rolePassword, readOnly, statementTimeout, fromRoles); err != nil {
 				return fmt.Errorf("failed to create role: %w", err)
 			}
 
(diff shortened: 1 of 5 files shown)


specs/spec.md
```diff
@@ -407,8 +413,81 @@ tiger db save-password svc-12345 --password your-password --password-storage key
...
+
+**Password Behavior:**
+By default, if neither `--password` flag nor `TIGER_NEW_PASSWORD` environment variable is provided, a secure random password will be automatically generated. The password is saved according to the `--password-storage` setting (keyring, pgpass, or none).
+
+**Options for `create role`:**
+- `--name`: Role name to create (required)
+- `--read-only`: Enable permanent read-only enforcement via `tsdb_admin.read_only_role` setting
+- `--from`: Comma-separated list of roles to inherit grants from (e.g., `app_role,readonly_role`)
+- `--statement-timeout`: Set statement timeout for the role (accepts any duration format: "30s", "5m", etc.)
```
nathanjcochran, line 455: I'm assuming `--statement-timeout` would just be implemented as `ALTER ROLE <name> SET statement_timeout = <timeout>`, right?

If that's the case, are we sure we want a flag just for setting this one specific configuration parameter? Or should we make something more generic, like `--config statement_timeout=5s`? I know having an `=` within a flag value is kind of weird, but I've seen other CLIs that do it, and this would be _way_ more flexible. Could also make it a `[]string` flag, so it supports setting multiple configuration parameters at once (e.g. via `--config key=value,key=value` or `--config key=value --config key=value`). Just a thought.

specs/spec.md
```diff
@@ -407,8 +413,81 @@ tiger db save-password svc-12345 --password your-password --password-storage key
...
+- **Survives reconnection**: The read-only enforcement persists across all connections, connection poolers, and database forks.
+
+**Use Cases:**
+- **AI data analysis**: Give AI agents like Claude Code safe access to production data for analysis without risk of accidental modifications
+- **Schema discovery**: Allow AI coding assistants to explore database structure to write better migrations, without write access
+- **Automated reporting**: Run analytics queries against current production data without forking
+- **Development tools**: Provide read-only access to tools that need to inspect but not modify data
+
+**Combining with `--from` for Safe Table Access:**
+The `--from` flag allows you to create a read-only role that inherits the same table access as your application role:
```
nathanjcochran, line 476: What is the default behavior if `--from` is not specified? What permissions does the new role have in that case? We should probably document the exact behavior (it's not immediately clear to me, as someone who hasn't worked a _ton_ with Postgres roles)

specs/spec.md
cevian: I want to be opinionated here and isolate the things we think are important and make them explicit flags

nathanjcochran: Cool, works for me!

scripts/test-integration.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
cevian, line 1: this file helps claude code run the integration tests correctly

internal/tiger/cmd/integration_test.go
```diff
@@ -731,17 +1203,17 @@ func TestAuthenticationErrorsIntegration(t *testing.T) {
 	}{
 		{
 			name: "db connection-string",
-			args: []string{"db", "connection-string", "non-existent-service", "--project-id", projectID},
+			args: []string{"db", "connection-string", "non-existent-service"},
 		},
 		{
 			name: "db connect",
-			args: []string{"db", "connect", "non-existent-service", "--project-id", projectID},
+			args: []string{"db", "connect", "non-existent-service"},
```
cevian, line 1192: fixes like this are for things that were broken from before

internal/tiger/cmd/db.go
```diff
@@ -348,6 +351,388 @@ Examples:
...
+
+	// Grant any other roles (besides tsdbadmin) if specified
+	for _, role := range otherRoles {
+		grantSQL := fmt.Sprintf("GRANT %s TO %s",
+			pgx.Identifier{role}.Sanitize(),
+			pgx.Identifier{roleName}.Sanitize())
+		if _, err := tx.Exec(ctx, grantSQL); err != nil {
+			return fmt.Errorf("failed to grant role %s: %w", role, err)
+		}
+	}
```
nathanjcochran: Isn't this duplicative with the `IN ROLE` clause added by `buildCreateRoleSQL`, which is used in the non-tsdbadmin case? Perhaps this logic should only exist in the `hasTsdbadmin` conditional block above?

internal/tiger/cmd/db.go
```diff
@@ -348,6 +351,388 @@ Examples:
...
+
+  # Create a read-only role with statement timeout
+  tiger db create role --name ai_analyst --read-only --statement-timeout 30s
+
+  # Create a role with specific password
+  tiger db create role --name ai_analyst --password=my-secure-password
+
+  # Create a role with password from environment variable
+  export TIGER_NEW_PASSWORD=my-secure-password
+  tiger db create role --name ai_analyst
```
nathanjcochran: There's probably not many cases where it actually makes sense to export your new password in an env var (which would make it available to all subsequent commands in the shell too). I'd vote to change this to:

```suggestion
  TIGER_NEW_PASSWORD=my-secure-password tiger db create role --name ai_analyst
```

internal/tiger/cmd/db.go
```diff
@@ -348,6 +351,388 @@ Examples:
...
+
+Technical Details:
+This command executes PostgreSQL statements in a transaction to create and configure the role.
+
+CREATE ROLE Options Used:
+  - LOGIN: Always enabled to allow the role to connect
+  - PASSWORD: Always set (from flag, env var, or auto-generated)
+  - IN ROLE: Added when --from flag is provided to inherit grants from existing roles
+
+GUCs (PostgreSQL Configuration Parameters) That May Be Set:
```
nathanjcochran: Not sure that the average user knows what a GUC is, or that mentioning it here provides much value. Why not just:


```suggestion
PostgreSQL Configuration Parameters That May Be Set:
```

internal/tiger/cmd/db.go
```diff
@@ -348,6 +351,388 @@ Examples:
...
+
+			// Output result in requested format
+			return outputCreateRoleResult(cmd, roleName, readOnly, statementTimeout, fromRolesList, cfg.Output)
+		},
+	}
+
+	// Add flags
+	cmd.Flags().StringVar(&roleName, "name", "", "Role name to create (required)")
+	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Enable permanent read-only enforcement via tsdb_admin.read_only_role")
+	cmd.Flags().StringVar(&fromRoles, "from", "", "Comma-separated list of roles to inherit grants from (e.g., app_role,readonly_role)")
```
nathanjcochran: You could use a [StringSliceVar](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringSliceVar) here instead, which would allow you to get rid of the parsing logic above, and would also allow it to support passing multiple values via multiple flags (e.g. `--from role1 --from role2`). That's what we use for the [`--addons` flag](https://github.com/timescale/tiger-cli/blob/ac94addc562fdad55c523876f28ab3f8b839d110/internal/tiger/cmd/service.go#L426), so probably good to use it here too for consistency.

internal/tiger/cmd/integration_test.go
```diff
@@ -355,11 +355,483 @@ func TestServiceLifecycleIntegration(t *testing.T) {
...
+	})
+
 	t.Run("DeleteService", func(t *testing.T) {
 		if serviceID == "" {
 			t.Skip("No service ID available for deletion")
 		}
 
+		// Clean up roles before deleting the service
+		cleanupRoles()
+
```
nathanjcochran: Is it really necessary to drop the roles right before we delete the service? Isn't deleting the service enough? Could probably get rid of the `cleanupRoles()` function entirely, I think (since this is the only place it's called).

scripts/test-integration.sh
nathanjcochran: I'm fine with this, but I also think this script could probably be greatly simplified. Seems like a lot of code for something that really just loads some env vars from a `.env` file and calls `go test` 🤷‍♂️.

Like, I think this entire script could probably just be:

```bash
#!/bin/bash

env $(cat .env | grep -v '^#' | xargs) go test ./... -run Integration "$@"
```

The `$@` would let you pass any `go test` flags through (so you wouldn't need special handling for `-v`, and you could provide a different value for `-run` that would override the first/default value provided).

Alternatively, we could add a [TestMain](https://pkg.go.dev/testing#hdr-Main) function that loads the env vars from the `.env` file itself, if they exist (assuming that was the main issue preventing Claude Code from running the tests correctly)? That way the integration tests would "just work" when running `go test` if the `.env` file was there. Though I guess we'd have to do that in every package that contains integration tests, which is maybe not ideal.

scripts/test-integration.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Check if .env exists
+if [ ! -f .env ]; then
+    echo -e "${RED}Error: .env file not found${NC}"
+    echo ""
+    echo "Please create a .env file with the required environment variables:"
+    echo "  TIGER_PUBLIC_KEY_INTEGRATION=your-public-key"
+    echo "  TIGER_SECRET_KEY_INTEGRATION=your-secret-key"
+    echo "  TIGER_PROJECT_ID_INTEGRATION=your-project-id"
+    echo "  TIGER_API_URL_INTEGRATION=https://api.tigerdata.cloud"
```
nathanjcochran:
```suggestion
    echo "  TIGER_API_URL_INTEGRATION=https://console.cloud.timescale.com/public/api/v1"
```

scripts/test-integration.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Parse arguments
+TEST_PATTERN="${1:-Integration}"  # Default to all integration tests
```
nathanjcochran: I don't think the `Integration` pattern actually matches all integration tests. See [TestServiceNotFound](https://github.com/timescale/tiger-cli/blob/bc504202987842c768d10bec29a4bc9ad2a2d5ba/internal/tiger/cmd/integration_test.go#L959-L969) for instance.

scripts/test-integration.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo -e "${RED}  Build failed${NC}"
+    exit 1
+fi
+
+echo ""
+echo -e "${BLUE}Running tests matching: ${TEST_PATTERN}${NC}"
+echo -e "${BLUE}────────────────────────────────────────────────────────${NC}"
+
+# Run the tests
+if go test ./internal/tiger/cmd \
```
nathanjcochran: Any reason this is hardcoded to `./internal/tiger/cmd`, instead of just doing `./...` (which would handle integration tests in other packages too)?

```suggestion
if go test ./... \
```

scripts/test-integration.sh
cevian: Renamed. to have Integration

PR #60, opened 2025-10-15
Add db save-password command with role support

cevian:
# NOTE: THIS INCLUDES A BREAKING CHANGE TO PASSWORD STORAGE

## Summary
- Implements new `tiger db save-password` command to save database passwords
- Adds support for storing passwords for different database roles independently
- Provides three password input methods with clear precedence

### New Features
- **save-password command**: New command under `tiger db` to save passwords to configured storage
- **Role-based storage**: Support for storing passwords per database role (tsdbadmin, readonly, custom roles)
- **Multiple input methods**:
  1. `--password=value` flag (highest precedence)
  2. `TIGER_NEW_PASSWORD` environment variable
  3. Interactive prompt (when neither provided)

### Implementation Details
- Modified password storage interface to include role parameter
- Updated all storage backends (KeyringStorage, PgpassStorage, NoStorage) with role support
- Added comprehensive test coverage for all input methods and storage backends
- Updated spec documentation to reflect the new command and its behavior

### Examples
```bash
# Save password with explicit value
tiger db save-password svc-12345 --password=mypass

# Save password for custom role
tiger db save-password svc-12345 --password=mypass --role readonly

# Use environment variable
export TIGER_NEW_PASSWORD=mypass
tiger db save-password svc-12345

# Interactive prompt
tiger db save-password svc-12345
Enter password: <user types password>
```

## Test plan
- [x] All existing tests pass
- [x] New tests for save-password command with explicit password
- [x] New tests for save-password with environment variable
- [x] New tests for interactive password prompt
- [x] New tests for custom roles
- [x] New tests for different storage backends (keyring, pgpass)
- [x] Error handling tests (no auth, no service ID, empty password)

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -12,10 +11,12 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 
 	"github.com/timescale/tiger-cli/internal/tiger/api"
 	"github.com/timescale/tiger-cli/internal/tiger/config"
 	"github.com/timescale/tiger-cli/internal/tiger/password"
+	"github.com/timescale/tiger-cli/internal/tiger/util"
 )
 
 var (
@@ -24,6 +25,16 @@ var (
 
 	// getServiceDetailsFunc can be overridden for testing
 	getServiceDetailsFunc = getServiceDetails
+
+	// checkStdinIsTTY can be overridden for testing to bypass TTY detection
+	checkStdinIsTTY = func() bool {
+		return util.IsTerminal(os.Stdin)
+	}
+
+	// readPasswordFromTerminal can be overridden for testing to inject password input
+	readPasswordFromTerminal = func(fd int) ([]byte, error) {
+		return term.ReadPassword(fd)
+	}
 )
 
 func buildDbConnectionStringCmd() *cobra.Command {
@@ -301,17 +312,18 @@ Examples:
 				// Use environment variable
 				passwordToSave = envPassword
 			} else {
-				// Interactive prompt
-				fmt.Fprint(cmd.OutOrStdout(), "Enter password: ")
-				scanner := bufio.NewScanner(cmd.InOrStdin())
-				if scanner.Scan() {
-					passwordToSave = scanner.Text()
-				} else {
-					if err := scanner.Err(); err != nil {
-						return fmt.Errorf("failed to read password: %w", err)
-					}
-					return fmt.Errorf("failed to read password: EOF")
+				// Interactive prompt - check if we're in a terminal
+				if !checkStdinIsTTY() {
+					return fmt.Errorf("TTY not detected - password required. Use --password flag or TIGER_NEW_PASSWORD environment variable")
 				}
+
+				fmt.Fprint(cmd.OutOrStdout(), "Enter password: ")
+				bytePassword, err := readPasswordFromTerminal(int(os.Stdin.Fd()))
+				if err != nil {
+					return fmt.Errorf("failed to read password: %w", err)
+				}
+				fmt.Fprintln(cmd.OutOrStdout()) // Print newline after hidden input
+				passwordToSave = string(bytePassword)
 				if passwordToSave == "" {
 					return fmt.Errorf("password cannot be empty")
 				}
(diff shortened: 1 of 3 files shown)


internal/tiger/password/password_storage.go
```diff
@@ -20,15 +20,18 @@ func getPasswordServiceName() string {
...
 	}
 	if service.ProjectId == nil {
 		return "", fmt.Errorf("project ID is required")
 	}
+	if role == "" {
+		return "", fmt.Errorf("role is required")
+	}
 
-	return fmt.Sprintf("password-%s-%s", *service.ProjectId, *service.ServiceId), nil
+	return fmt.Sprintf("password-%s-%s-%s", *service.ProjectId, *service.ServiceId, role), nil
```
cevian, line 34: This is the cause of the breaking change

internal/tiger/cmd/db.go
```diff
@@ -229,6 +233,98 @@ Examples:
...
+			}
+
+			fmt.Fprintf(cmd.OutOrStdout(), "Password saved successfully for service %s (role: %s)\n",
+				*service.ServiceId, dbSavePasswordRole)
+			return nil
+		},
+	}
+
+	// Add flags for db save-password command
+	cmd.Flags().StringVar(&dbSavePasswordValue, "password", "", "Password to save")
```
murrayju: Feel like we should allow `-p` as a shorter option

internal/tiger/cmd/db.go
```diff
@@ -229,6 +233,98 @@ Examples:
...
+				fmt.Fprint(cmd.OutOrStdout(), "Enter password: ")
+				scanner := bufio.NewScanner(cmd.InOrStdin())
+				if scanner.Scan() {
+					passwordToSave = scanner.Text()
+				} else {
+					if err := scanner.Err(); err != nil {
+						return fmt.Errorf("failed to read password: %w", err)
+					}
+					return fmt.Errorf("failed to read password: EOF")
+				}
```
nathanjcochran: Should we use [term.ReadPassword](https://pkg.go.dev/golang.org/x/term#ReadPassword) instead (like we're doing [here](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/internal/tiger/cmd/auth.go#L238-L239))?

Also, this should probably error out if stdin is not a terminal/TTY (like we do in [other places](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/internal/tiger/cmd/auth.go#L218-L220) before showing an interactive prompt).

internal/tiger/cmd/db.go
```diff
@@ -229,6 +233,98 @@ Examples:
...
+				if passwordToSave == "" {
+					return fmt.Errorf("password cannot be empty")
+				}
+			}
+
+			// Save password using configured storage
+			storage := password.GetPasswordStorage()
+			err = storage.Save(service, passwordToSave, dbSavePasswordRole)
+			if err != nil {
```
nathanjcochran: Minor but I think it's more idiomatic to scope `err` to the conditional block if possible:


```suggestion
			if err := storage.Save(service, passwordToSave, dbSavePasswordRole); err != nil {
```

internal/tiger/cmd/db.go
```diff
@@ -229,6 +233,98 @@ Examples:
...
+				return fmt.Errorf("failed to save password: %w", err)
+			}
+
+			fmt.Fprintf(cmd.OutOrStdout(), "Password saved successfully for service %s (role: %s)\n",
```
nathanjcochran: Fwiw, we switched to outputting most "status messages" to stderr recently, since a lot of commands now output structured data to stdio for use in scripting (e.g. when using the `-o json` flag). That's not really a concern here, but we might want to output messages like this to stderr anyways for consistency. Should probably update `CLAUDE.md` with something about that convention.

internal/tiger/cmd/service.go
```diff
@@ -828,7 +828,7 @@ func waitForServiceReady(client *api.ClientWithResponses, projectID, serviceID s
 func handlePasswordSaving(service api.Service, initialPassword string, output io.Writer) bool {
 	// Note: We don't fail the service creation if password saving fails
 	// The error is handled by displaying the appropriate message below
-	result, _ := password.SavePasswordWithResult(service, initialPassword)
+	result, _ := password.SavePasswordWithResult(service, initialPassword, "tsdbadmin")
```
nathanjcochran, line 829: Not a big deal, but `tsdbadmin` is hard-coded in several places now. Might be worth making a constant at some point? Doesn't have to be now, just a thought.

internal/tiger/password/password_storage.go
```diff
@@ -334,17 +346,24 @@ func GetPasswordStorage() PasswordStorage {
...
 			Message: "No password provided",
 		}, nil
 	}
+	if role == "" {
+		return PasswordStorageResult{
+			Success: false,
+			Method:  "none",
+			Message: "Role is required",
+		}, fmt.Errorf("role is required")
+	}
```
nathanjcochran, line 363: I think this check is duplicative, as I think all of the `storage.Save` (called below) methods do the same validation 🤷‍♂️.

specs/spec.md
```diff
@@ -31,6 +31,7 @@ mv tiger /usr/local/bin/
 - `TIGER_CONFIG_DIR`: Configuration directory (default: ~/.config/tiger)
 - `TIGER_OUTPUT`: Default output format (json, yaml, table)
 - `TIGER_ANALYTICS`: Enable/disable usage analytics (true/false)
+- `TIGER_NEW_PASSWORD`: Password to use for save-password command (used when --password flag not provided)
```
nathanjcochran: This was added to the "Global Configuration" section, but it's not global, right? It's just for the `tiger db save-password` command?

internal/tiger/password/password_storage.go
cevian: I'd rather keep this for safety as it never makes sense to save without a role

internal/tiger/cmd/service.go
cevian: Ill do this in a diff pr

PR #63, opened 2025-10-15
Increase polling frequency

nathanjcochran:
This PR increases the polling frequency when creating or deleting services from every 10 seconds to every 1 second.

Additionally, to avoid spamming the console with lots of `Service status: QUEUED...` messages (which were previously output every time we polled), I implemented a more dynamic status message that updates in real-time and shows a little loading spinner thingy:

https://github.com/user-attachments/assets/60664360-b6ef-4552-b9ba-d635e155ae24

If the output is not a tty, it doesn't try to render the dynamic loading spinner/message, but instead prints the status message on a new line every time the message changes. This is to prevent weird terminal control characters from showing up in the output when it's redirected to a file.

I also fixed some buggy and duplicate error output.

Closes [AGE-188](https://linear.app/tigerdata/issue/AGE-188/tiger-cli-increase-polling-frequency)

### DIFF
--- a/internal/tiger/cmd/spinner.go
+++ b/internal/tiger/cmd/spinner.go
@@ -12,28 +12,31 @@ import (
 // spinnerFrames defines the animation frames for the spinner
 var spinnerFrames = []string{"⢎ ", "⠎⠁", "⠊⠑", "⠈⠱", " ⡱", "⢀⡰", "⢄⡠", "⢆⡀"}
 
-type Spinner struct {
-	// Populated when output is a terminal
-	program *tea.Program
+type Spinner interface {
+	// Update changes the spinner's displayed message.
+	Update(message string, args ...any)
 
-	// Populated when output is not a terminal
-	output io.Writer
-	model  *spinnerModel
+	// Stop terminates the spinner program and waits for it to finish.
+	Stop()
 }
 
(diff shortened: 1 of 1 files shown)


internal/tiger/cmd/spinner.go
```diff
@@ -0,0 +1,126 @@
...
+
+	s.program.Quit()
+	s.program.Wait()
+}
+
+func (s *Spinner) println() {
+	fmt.Fprintln(s.output, s.model.View())
+}
+
+func isTerminal(w io.Writer) bool {
```
murrayju: I wrote a similar function for the version check stuff. Should consolidate these.

Reviewing on mobile, too hard to provide a link, sorry.

nathanjcochran: Good call! Done here: https://github.com/timescale/tiger-cli/pull/63/commits/ac42720a17c5a0315cb4b376fc05f176fea7413d.

I decided to rely on the official [golang.org/x/term](https://pkg.go.dev/golang.org/x/term) package rather than the [github.com/mattn/go-isatty](https://pkg.go.dev/github.com/mattn/go-isatty) package you were using. I think the functionality should be essentially the same, but I trust the official package slightly more.

PR #64, opened 2025-10-16
Flag to exit with nonzero code when update available

murrayju:
This improves scriptability to leverage an exit code when an update is available.

Exits with code `7` when doing an explicit check: `tiger version --check`
image

Will exit with `0` if there is no update
image


cevian (approved):
Please update the spec document as well as I believe it lists error codes

PR #65, opened 2025-10-16
Preserve stderr color output when redirecting stdout

murrayju:
The `fatih/color` package detects when stdout is not connected to a tty and disables itself. Since we are writing colors to stderr instead, we have to override that.

Before
image

After
image

Additionally, I wired up a global `--no-color` flag, and a `no_color` config option. When these are set, color is omitted regardless of tty attachment.

image

(includes an env var)
image

### DIFF
--- a/internal/tiger/cmd/root.go
+++ b/internal/tiger/cmd/root.go
@@ -23,7 +23,6 @@ func buildRootCmd() *cobra.Command {
 	var passwordStorage string
 	var skipUpdateCheck bool
 	var colorFlag bool
-	var noColorFlag bool
 
 	cmd := &cobra.Command{
 		Use:   "tiger",
@@ -53,9 +52,6 @@ tiger auth login
 				zap.Bool("debug", cfg.Debug),
 			)
 
-			if noColorFlag && cmd.Flag("no-color").Changed {
-				cfg.UpdateField("color", false)
-			}
 			if !cfg.Color {
 				color.NoColor = true
 			}
@@ -103,9 +99,6 @@ tiger auth login
 	cmd.PersistentFlags().StringVar(&passwordStorage, "password-storage", config.DefaultPasswordStorage, "password storage method (keyring, pgpass, none)")
 	cmd.PersistentFlags().BoolVar(&skipUpdateCheck, "skip-update-check", false, "skip checking for updates on startup")
 	cmd.PersistentFlags().BoolVar(&colorFlag, "color", true, "enable colored output")
-	cmd.PersistentFlags().MarkHidden("color")
-	cmd.PersistentFlags().BoolVar(&noColorFlag, "no-color", false, "disable colored output")
-	cmd.MarkFlagsMutuallyExclusive("color", "no-color")
 
 	// Bind flags to viper
 	viper.BindPFlag("debug", cmd.PersistentFlags().Lookup("debug"))


cevian (approved):
We should also probably use color in more places but thats for another PR

internal/tiger/cmd/config_test.go
```diff
@@ -143,6 +143,7 @@ version_check_last_time: ` + now.Format(time.RFC3339) + "\n"
 		"docs_mcp_url":            "https://mcp.tigerdata.com/docs",
 		"project_id":              "json-project",
 		"service_id":              "",
+		"no_color":                false,
```
nathanjcochran: My personal opinion is that the config parameter/flag should be `color`, not `no_color`. I know we have some [convention](https://github.com/timescale/tiger-cli/blob/863f334dae83487086fad6f4c8a6f6ca7246602c/CLAUDE.md?plain=1#L651-L658) around flags being `--no-whatever` instead of `--whatever=false`, but I feel like that pattern breaks down a bit when it's also a config option. My reasoning is:

1. We already have an `analytics` config option that defaults to `true` (meaning that disabling it via the flag is `--analytics=false`). So there's existing precedent for this pattern too (and imo, we should try to be consistent for boolean config options, and I would not like `no_analytics`).
2. In the config, the boolean value always has to be specified, so you don't get the same ergonomic benefit of the `--no-config` flag. And setting `"no_color": true` in the config feels less clear/nice to me than `"color": false`.
3. Perhaps my biggest complaint: if a user sets `"no_color": true` in their config, but wants to re-enable color for a single command via the flag, they have to use a double-negative: `--no-color=false`, which I _really_ don't like as a way to say "use color". If the config/flag is `color`, we avoid the possibility for ever needing a double-negative.

That's just my take though. If everybody else disagrees, I'm fine with whatever.

Otherwise this PR LGTM 👍

nathanjcochran: Also, some well-known commands (like `ls`) take a `--color` flag.

murrayju: Some CLI's treat `--no-X` the same as `--X=false`. I think that'd be my preference, with the config value then just being called `color`, like you suggested.

nathanjcochran: I'd rather not have to deal with two different flags that are related but are mutually exclusive (e.g. what happens if both are provided? etc.). In other words, I agree with [this](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/CLAUDE.md?plain=1#L670-L675).

Personally, I don't think `--color=false` is that bad 🤷‍♂️. It's the same as `--analytics=false`, which is currently a thing. I feel like the `--no-whatever` convention is good for flags that aren't also config options, but imo, I think it's fine to break that convention for config option flags, and I feel like it's most clear if the flags and config options are 1:1.

murrayju: Whelp, I already coded it up to support both, and I don't think it looks too bad: https://github.com/timescale/tiger-cli/pull/65/commits/c95f9da2e9ceea90bb3f787b58f957f776ed6b03

I added `--no-color` as a "hidden" flag, and it's like 2 lines to handle it.

> what happens if both are provided

`--no-color` simply has higher precedence. I'd be fine with making this case error, too. I really don't think this is something to worry about.

> In other words, I agree with [this](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/CLAUDE.md?plain=1#L670-L675).

lol, [just above that](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/CLAUDE.md?plain=1#L655-L656), it says

```
1. **Default Positive Behavior** - The positive behavior is the default (no flag needed)
2. **Explicit Negative Override** - Use `--no-<feature>` to disable the default behavior
```

Which would guide us back towards what I had originally. I think that doc is self-contradicting in a lot of ways.

If you still feel strongly, I can remove it.

murrayju: Thinking about it more, `--no-color` is the one I want to appear in help (it is the most likely to be used), so I hid `--color` instead. Also marked them mutually exclusive, because why not.

https://github.com/timescale/tiger-cli/pull/65/commits/32517998d31acd1ecdb8fc2b2ec0c9dc48af92a3

nathanjcochran: > lol, [just above that](https://github.com/timescale/tiger-cli/blob/419c587d7f56ca51358195d9186b122c91f6956e/CLAUDE.md?plain=1#L655-L656), it says

Yeah, I linked that in my very first comment above 😉

nathanjcochran: Fwiw, I'm pretty opposed to supporting both `--color` and `--no-color`. I think it's over-complicating things. I think we should just use `--color` everywhere to match the config (all the other global flags match the corresponding config values) and to match `--analytics`.

I also think that showing `--no-color` in the help text will push people towards `--no-color=false`, which is the double-negative thing I really don't like (see #3 in my original comment).

If it's a problem for people, we can easily add `--no-color` as a second flag in the future. But if we do, I think we should do it across the board (i.e. `--no-analytics`, etc.). But I'm just not convinced it's worth it right now.

murrayju: Ok, I don't have a strong enough opinion to argue about it.

https://github.com/timescale/tiger-cli/pull/65/commits/1ee4a95bcf4542e9cb1ffd0cfbf417e26868d843

nathanjcochran: Thanks 🙏. I'm open to bringing it back later if everyone agrees that's a pattern we want to follow for global boolean flags. I just want to keep things simple for now

internal/tiger/config/config.go
```diff
@@ -244,10 +250,10 @@ func setBool(key, val string) (bool, error) {
 	return b, nil
 }
 
-// updateField updates the field in the Config struct corresponding to the given key.
+// UpdateField updates the field in the Config struct corresponding to the given key.
 // It accepts either a string (from user input) or a typed value (string/bool from defaults).
 // The function validates the value and updates both the struct field and viper state.
-func (c *Config) updateField(key string, value any) (any, error) {
```
nathanjcochran, line 250: Why did this function have to be exported? I'm not seeing where it's being used outside the package now 🤔

PR #66, opened 2025-10-17
AGE-212: eliminate PasswordMode in favor of boolean

murrayju:
This eliminates the `PasswordMode` enum, replacing with a simple `WithPassword` boolean. The `GetConnectionDetails` will no longer fail in any case where the password is unavailable, but the caller can still handle this by checking by a blank password (as is done by `tiger db connection-string --with-password`). In other cases, the connection string is simply built without the password in it.

### DIFF
--- a/internal/tiger/mcp/service_tools.go
+++ b/internal/tiger/mcp/service_tools.go
@@ -490,7 +490,8 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
 		logging.Debug("MCP: Failed to build connection string", zap.Error(err))
 	} else {
 		if input.WithPassword && details.Password == "" {
-			logging.Debug("MCP: Requested password but password not available")
+			// This should not happen since we have InitialPassword, but check just in case
+			logging.Error("MCP: Requested password but password not available")
 		}
 		output.Service.ConnectionString = details.String()
 	}


internal/tiger/cmd/db.go
```diff
@@ -140,7 +144,7 @@ Examples:
 			details, err := password.GetConnectionDetails(service, password.ConnectionDetailsOptions{
 				Pooled:       dbConnectPooled,
 				Role:         dbConnectRole,
-				PasswordMode: password.PasswordExclude,
+				WithPassword: false,
```
nathanjcochran: Because the zero value for a boolean is false, you can also just omit `WithPassword` from the struct initialization entirely. Same for other places where it's explicitly being set to false now. Just a matter of preference/style though, really 🤷‍♂️.

internal/tiger/mcp/service_tools.go
```diff
@@ -381,7 +381,7 @@ func (s *Server) handleServiceGet(ctx context.Context, req *mcp.CallToolRequest,
 	if details, err := password.GetConnectionDetails(service, password.ConnectionDetailsOptions{
 		Pooled:       false,
 		Role:         "tsdbadmin",
-		PasswordMode: password.GetPasswordMode(input.WithPassword),
+		WithPassword: input.WithPassword,
```
nathanjcochran, line 383: Is it going to confuse the LLM if it requests a password from the `service_get` tool call but it doesn't get one back (and there's no indication as to why)? Should we return an error instead, and just let the LLM try again without `with_password: true` (assuming it even wants the service details without the password)?

internal/tiger/cmd/service.go
```diff
@@ -737,7 +737,7 @@ func prepareServiceForOutput(service api.Service, withPassword bool, output io.W
 	opts := password.ConnectionDetailsOptions{
 		Pooled:          false,
 		Role:            "tsdbadmin",
-		PasswordMode:    password.GetPasswordMode(withPassword),
+		WithPassword:    withPassword,
```
nathanjcochran, line 740: So `tiger db connection-string --with-password` will error out if the password can't be found, but `tiger service get --with-password` won't? Given the scripting use case (e.g. `tiger svc get -o env --with-password`), is that the right behavior? Genuine question (I'm not sure).

I definitely agree that `tiger service list --with-password` shouldn't error out if one service doesn't have a password, though (although I'm honestly wondering if being able to list all of your services along with their passwords really makes sense in the first place 🤔)

internal/tiger/mcp/db_tools.go
```diff
@@ -150,7 +150,7 @@ func (s *Server) handleDBExecuteQuery(ctx context.Context, req *mcp.CallToolRequ
 	details, err := password.GetConnectionDetails(service, password.ConnectionDetailsOptions{
 		Pooled:       input.Pooled,
 		Role:         input.Role,
-		PasswordMode: password.PasswordRequired, // MCP always requires password
+		WithPassword: true,
```
nathanjcochran, line 153: Nice! This is a case where I think the "optional" behavior makes a ton of sense, since `pgx` should automatically check the `pgpass` file if a password isn't in the connection string (though one of us should probably verify that at some point). So it essentially allows for falling back to the `pgpass` file if a password can't be found in the keyring.

We might also want to think about falling back to the keyring if the configured password storage mechanism is `pgpass` but the password can't be found there (in case someone switches their password storage from `keyring` to `pgpass` after creating a service, for example). Definitely doesn't have to be part of this PR - just a thought for the future.

nathanjcochran (approved):
Left a couple questions about whether we should error out in other cases besides `tiger db connection-string --with-password` when the password can't be found, but overall LGTM! Thanks for doing this!

internal/tiger/mcp/db_tools.go
cevian: also pgx will fallback to env variables too so good change

internal/tiger/mcp/service_tools.go
cevian: Yeah I would also prefer an error. The mcp is heavily discouraged from asking for this anyway

internal/tiger/cmd/service.go
cevian: I would error here and would remove the `--with-password` flag for list entirely

cevian (approved):

internal/tiger/mcp/db_tools.go
murrayju: > We might also want to think about falling back to the keyring

Yeah, we discussed this somewhere else. I think we should _read_ passwords from _all_ storage options, with the config then only affecting where we write, and where we read _first_.

internal/tiger/cmd/service.go
nathanjcochran: Yeah, I think I agree with that.

murrayju: We currently don't error at all if we totally fail to get the connection details, we just [log a warning](https://github.com/timescale/tiger-cli/blob/003d20ebb339e61b9237a26b7538e191b2a2253b/internal/tiger/cmd/service.go#L746). I do feel that outputting as much of the service info as possible is better than failing outright. A missing password feels _less_ severe than missing the hostname. Should this just also be a warning?

nathanjcochran: I'd be fine with a warning.

I do think there are cases where people might want to run `tiger svc get -o env --with-password` and have it be "best effort" with respect to the password, so I think that's reasonable. At least a warning will explain to the user why the result didn't include a password, which might otherwise be confusing 👍.

Should we do the same for `tiger db connection-string --with-password`? I do think it would be nice to be consistent. Unless there's a good argument for why these two commands should have different behavior?

internal/tiger/cmd/db.go
murrayju: https://github.com/timescale/tiger-cli/pull/66/commits/1b991a7d08827dfdbe5090ba3796da99dd785cfc

internal/tiger/mcp/service_tools.go
murrayju: https://github.com/timescale/tiger-cli/pull/66/commits/003d20ebb339e61b9237a26b7538e191b2a2253b

internal/tiger/cmd/service.go
murrayju: Removed `svc list --with-password` https://github.com/timescale/tiger-cli/pull/66/commits/caee332ed32720104a5b8af40d0238411b0b000b

murrayju: I guess I really don't know when [service.Endpoint might be missing](https://github.com/timescale/tiger-cli/blob/caee332ed32720104a5b8af40d0238411b0b000b/internal/tiger/password/connection.go#L41), or [endpoint.Host](https://github.com/timescale/tiger-cli/blob/caee332ed32720104a5b8af40d0238411b0b000b/internal/tiger/password/connection.go#L57). If those are expected to be "impossible", it would be better to just error out (in these cases as well as missing password)

murrayju: And this [pooler logic](https://github.com/timescale/tiger-cli/blob/caee332ed32720104a5b8af40d0238411b0b000b/internal/tiger/password/connection.go#L49-L52) is related too. If you do `tiger db connection-string --pooled`, it outputs a warning and falls back to a direct connection if not available. We should consider being consistent about strictness, whether we error or have fallback behavior.

murrayju: > a good argument for why these two commands should have different behavior

The only argument I have is that `tiger db connection-string --with-password` outputs exactly one thing (which it can't format as requested), whereas `tiger svc get --with-password` outputs many things, of which 2 are affected. One could argue that there is no need for both of these commands in the first place, but this difference (along with the `--pooled` flag support) could be _the_ reason to have both 🤷 

I do feel that many things about the CLI are not self-consistent, so it is hard for me to say how any one command should behave.

nathanjcochran: I also don't know whether it's possible for `service.Endpoint` or `endpoint.Host` to be missing. I haven't really been worrying about that code path, but I don't think it's likely to happen in practice 😅.

And fwiw, I agree that we should try to be consistent about strictness with `--pooled` and `--with-password`, whatever we decide.

murrayju: Discussion continued [in Slack](https://iobeam.slack.com/archives/C099S47CSRX/p1760715047755529).

The decision is that `svc get` and `db connection-string` should fail if the passed flags can't be respected (`--with-password` and `--pooled`)

For `svc create`, it should only error out if the service fails to be created. Anything else is just written to stderr as a warning.

https://github.com/timescale/tiger-cli/pull/66/commits/9fad7bc91f0a9add7eb9edd377ff5224ba90d2f1

internal/tiger/mcp/service_tools.go
```diff
@@ -489,6 +489,9 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
 	}); err != nil {
 		logging.Debug("MCP: Failed to build connection string", zap.Error(err))
 	} else {
+		if input.WithPassword && details.Password == "" {
+			logging.Debug("MCP: Requested password but password not available")
+		}
```
nathanjcochran, line 495: Might be worth leaving a comment saying that we never expect this to actually happen. Because we pass `InitialPassword` into `GetConnectionDetails`, we should always get a password out, unless the API response didn't include the `initialPassword`, which would be a major problem 😅.

internal/tiger/cmd/service.go
```diff
@@ -568,9 +568,12 @@ type OutputService struct {
 }
 
 // outputService formats and outputs a single service based on the specified format
-func outputService(cmd *cobra.Command, service api.Service, format string, withPassword bool) error {
+func outputService(cmd *cobra.Command, service api.Service, format string, withPassword bool, strict bool) error {
```
nathanjcochran, line 571: Haha almost feels like we want an enum with three values again, instead of two booleans (since `withPassword: false` + `strict: true` is basically meaningless) 😅. But this is totally fine for now.

nathanjcochran (approved):
LGTM!

internal/tiger/cmd/service.go
murrayju: I had the same thought, but I actually think the code to map an enum to a bool is more complex than two bools with a don't care case.

internal/tiger/mcp/service_tools.go
murrayju: https://github.com/timescale/tiger-cli/pull/66/commits/b00216335ee03fb38f74b903af48bdb9e80cea3d

PR #70, opened 2025-10-17
Allow setting environment tags for forks and create

Askir:


cevian (approved):
But we also need to add visibility into the tags in list/get/describe. Can be a separate PR but need doing

PR #74, opened 2025-10-17
Remove `project_id` config option and global flag

nathanjcochran:
See [Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1760718673826009) for background.

This PR removes the `project_id` config option and the `--project-id` global flag, since users cannot actually change their project id after logging in, given that the API key is scoped to a particular project. This has been confusing in the past, since the presence of the config option and flag imply that users should be able to change it post-login.

Instead, the project id is now stored in the keyring (or file fallback) along with the api key. Specifically, the value stored in the keyring/file fallback is now a JSON object with `api_key` and `project_id` fields. I think this will also give us more flexibility going forwards (e.g. if we want to support different authentication mechanicsms in a backwards-compatible way, we can store different fields in that JSON object, and handle them accordingly).

Additionally, I updated `tiger auth status` to display the project ID, since it will no longer be visible in the `tiger config show` output. I think this will hopefully make it more clear that your login is scoped to a particular project, and prevent users from erroneously trying to change the project ID after login.

Note that while this is a breaking change in terms of how the api key and project ID are stored/used, it also changes the name of the keyring entry and fallback file (to `credentials`), meaning that users should simply appear to be logged out after updating to this version of the CLI (and will therefore have to log in again). That's perhaps not _ideal_, but it at least shouldn't leave them in a weird/broken state.


internal/tiger/config/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+		panic("test must call SetTestServiceName() to set a unique keyring service name")
+	}
+
+	return keyringServiceName
+}
+
+// SetTestServiceName sets a unique service name for testing based on the test name
+// This allows tests to use unique service names to avoid conflicts when running in parallel
+// The cleanup is automatically registered with t.Cleanup()
+func SetTestServiceName(t *testing.T) {
```
murrayju, line 47: I feel like this could be in the test file. Lots of test related code leaking into the implementation looks off to me, but maybe this is a common go pattern 🤷

internal/tiger/config/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+
+	data, err := os.ReadFile(credentialsFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", "", ErrNotLoggedIn
+		}
+		return "", "", fmt.Errorf("failed to read credentials file: %w", err)
+	}
+
+	credentials := strings.TrimSpace(string(data))
```
murrayju: Any particular reason the file needs trimming but the keyring does not?

murrayju (approved):
lgtm, definitely simpler and more flexible this way. Just a couple minor comments.

I didn't test anything.

internal/tiger/cmd/db.go
```diff
@@ -282,10 +265,10 @@ func getServiceDetails(cmd *cobra.Command, args []string) (api.Service, error) {
 
 	cmd.SilenceUsage = true
 
-	// Get API key for authentication
-	apiKey, err := getAPIKeyForDB()
+	// Get API key and project ID for authentication
+	apiKey, projectID, err := getCredentialsForDB()
 	if err != nil {
-		return api.Service{}, exitWithCode(ExitAuthenticationError, fmt.Errorf("authentication required: %w", err))
+		return api.Service{}, exitWithCode(ExitAuthenticationError, fmt.Errorf("authentication required: %w. Please run 'tiger auth login'", err))
```
cevian, line 391: nice error code addition

internal/tiger/config/credentials.go
```diff
@@ -0,0 +1,204 @@
...
+
+// removeCredentialsFromKeyring removes credentials from keyring (test helper)
+func removeCredentialsFromKeyring() {
+	keyring.Delete(GetServiceName(), keyringUsername)
+}
+
+// removeCredentialsFile removes credentials file
+func removeCredentialsFile() error {
+	configDir := GetConfigDir()
+	credentialsFile := fmt.Sprintf("%s/credentials", configDir)
```
cevian: Nit: this comes up a few time I'd make a util func to get the credentials file name.

cevian (approved):
Looks great and a valuable cleanup. Only a few nits

internal/tiger/config/credentials.go
nathanjcochran: The reason it's an exported function is because it's called from tests in _other_ packages. So it cannot be unexported or in a `_test.go` file. It needs to be exported from the package itself, so other packages (or their tests) can call it.

I agree that this is not ideal, though. I generally don't like test-related functionality leaking into a package's exported surface area. The pattern was discussed a bit [here](https://github.com/timescale/tiger-cli/pull/47#discussion_r2411960037) too, but it works for now, and fixing it is relatively low-priority at the moment.

internal/tiger/config/credentials.go
nathanjcochran: Ask Claude lol. 

But no, it shouldn't be necessary imo. When the json is parsed, leading and trailing whitespace is ignored. So I think calling `string.TrimSpace` is unnecessary. Removed here: https://github.com/timescale/tiger-cli/pull/74/commits/2b5720301cc65f34a0c41774d2e8cf36db748e99

internal/tiger/config/credentials.go
nathanjcochran: Good call! Fixed here: https://github.com/timescale/tiger-cli/pull/74/commits/76c63679ddfc3098278081f1041464b93da679eb

PR #80, opened 2025-10-21
Run `go mod tidy --diff` in CI

nathanjcochran:
Runs `go mod tidy --diff` in the CI pipeline, to ensure that the `go.mod` and `go.sum` files correctly reflects the modules needed by the source code (i.e. no missing modules and no extra unused modules). This is basically a way to make sure people run `go mod tidy` after adding/removing dependencies.

Per `go help mod tidy`:

> The -diff flag causes tidy not to modify go.mod or go.sum but
> instead print the necessary changes as a unified diff. It exits
> with a non-zero code if the diff is not empty.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,12 @@ jobs:
           eval $(echo 'test' | gnome-keyring-daemon --start --components=secrets)
           echo "GNOME_KEYRING_CONTROL=$GNOME_KEYRING_CONTROL" >> $GITHUB_ENV
 
+      - name: ShellCheck
+        run: shellcheck scripts/install.sh
+
+      - name: Mod
+        run: go mod tidy --diff
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -37,6 +37,12 @@ jobs:
           eval $(echo 'test' | gnome-keyring-daemon --start --components=secrets)
           echo "GNOME_KEYRING_CONTROL=$GNOME_KEYRING_CONTROL" >> $GITHUB_ENV
 
+      - name: ShellCheck
+        run: shellcheck scripts/install.sh
```
nathanjcochran, line 41: Moved this up from below because it's extremely fast, and better to give fast feedback on the install script than make people wait for all the tests to run.

murrayju (approved):

PR #84, opened 2025-10-23
Refactor context handling

graveland:
Stop using context.Background() and use a root context that's got signal.NotifyContext() on it. This allows control-C to interrupt everything.

### DIFF
--- a/cmd/tiger/main.go
+++ b/cmd/tiger/main.go
@@ -50,7 +50,7 @@ func notifyContext(parent context.Context) (context.Context, context.CancelFunc)
 	go func() {
 		select {
 		case sig := <-sigChan:
-			logging.Info("Received interrupt signal", zap.Stringer("signal", sig))
+			logging.Info("Received interrupt signal, press control-C again to exit", zap.Stringer("signal", sig))
 			signal.Stop(sigChan) // Restore default signal handling behavior
 			cancel()
 		case <-ctx.Done():
--- a/internal/tiger/cmd/mcp.go
+++ b/internal/tiger/cmd/mcp.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"time"
 
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
@@ -277,10 +276,8 @@ func startHTTPServer(ctx context.Context, host string, port int) error {
 	<-ctx.Done()
 
 	// Shutdown server gracefully
-	logging.Info("Shutting down HTTP server...")
-	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer shutdownCancel()
-	if err := httpServer.Shutdown(shutdownCtx); err != nil {
+	logging.Info("Gracefully shutting down HTTP server..., press control-C twice to immediately shutdown")
+	if err := httpServer.Shutdown(context.Background()); err != nil {
 		return fmt.Errorf("failed to shut down HTTP server: %w", err)
 	}
 


cmd/tiger/main.go
```diff
@@ -1,7 +1,35 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/timescale/tiger-cli/internal/tiger/cmd"
+)
 
 func main() {
-	cmd.Execute()
+	if err := run(); err != nil {
```
graveland, line 18: These changes are to make sure that the context is actually cancelled. os.Exit() skips cleanup entirely, so by doing it in a function w/ defer and only exiting in a single location, we allow cleanup to happen. It could be improved still: I haven't investigated the MCP bits at all, but a waitgroup to wait for cleanup to finish before exiting could potentially be added too.

internal/tiger/cmd/mcp.go
```diff
@@ -312,27 +309,16 @@ func getListener(host string, startPort int) (net.Listener, int, error) {
...
-		case <-ctx.Done():
-		}
+		<-ctx.Done()
+		logging.Info("Received interrupt signal, shutting down MCP server...")
 	}()
 
-	return ctx, func() {
-		cancel()
-		signal.Stop(sigChan)
-	}
```
nathanjcochran, line 337: A couple comments:
1. I feel like this function could probably be removed now. It doesn't seem to be doing anything meaningful anymore, besides logging that a signal was received. But that could be misleading now, since all we _really_ know is that the context was cancelled. Additionally, the log message could also race with other things waiting for `<- ctx.Done()` now, which could make for confusing (out-of-order) logs. Perhaps we should actually move the original function to `main.go` and use it instead of `signal.NotifyContext` instead? (so that we still log something when the signal is received, which is nice when cleanup can take awhile - such as in the MCP server graceful shutdown case).
2. There's a subtle bug here, I think. Previously, the cancel func that gets returned would call `signal.Stop(sigChan)`, thereby restoring the default signal handling behavior. This meant that a second ctrl-c after `stop()` would "hard kill" the MCP server if graceful shutdown was taking too long. We explicitly leveraged that fact [here](https://github.com/timescale/tiger-cli/blob/0bc7923a2fd7a926a10f02292e18d6d348bcc6ac/internal/tiger/cmd/mcp.go#L278-L285). But I don't think that works correctly anymore, since the cancel func no longer restores the default signal handling behavior.

So my advice would be to remove this function entirely, and instead create a similar function in `main.go` and use it instead of `signal.NotifyContext`:

```go
// notifyContext sets up graceful shutdown handling and returns a context and
// cleanup function. This is nearly identical to [signal.NotifyContext], except
// that it logs a message when a signal is received and also restores the default
// signal handling behavior.
func notifyContext(parent context.Context) (context.Context, context.CancelFunc) {
	ctx, cancel := context.WithCancel(parent)

	sigChan := make(chan os.Signal, 1)
	signal.Notify(sigChan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
	go func() {
		select {
		case sig := <-sigChan:
			logging.Info("Received interrupt signal",
				zap.Stringer("signal", sig),
			)
            signal.Stop(sigChan) // Restore default signal handling behavior
			cancel()
		case <-ctx.Done():
		}
	}()

	return ctx, func() {
		cancel()
		signal.Stop(sigChan)
	}
}

```

graveland: I'm fine with that. This could be fixed in that function by changing:
```go
if err := httpServer.Shutdown(context.Background()); err != nil {
```
to
```go
shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
defer shutdownCancel()
if err := httpServer.Shutdown(shutdownCtx); err != nil {
```

internal/tiger/cmd/root.go
```diff
@@ -82,8 +85,7 @@ tiger auth login
 	initConfigFunc := func() {
 		configDirFlag := cmd.PersistentFlags().Lookup("config-dir")
 		if err := config.SetupViper(config.GetEffectiveConfigDir(configDirFlag)); err != nil {
-			fmt.Fprintf(os.Stderr, "Error setting up config: %v\n", err)
-			os.Exit(1)
+			panic(fmt.Errorf("error setting up config: %w", err))
```
nathanjcochran: I don't _love_ this being a panic. I'm not sure if it can realistically happen in normal circumstances, but I think it would probably be better to return an error instead. Is there a way we can move this out of `cobra.OnInitialize` so it's possible to return the error? I'm not sure of the exact order of operations re: when `OnInitialize` runs vs this top-level `PersistentPreRunE` function, but I feel like having this code in `PersistentPreRunE` would probably be functionally equivalent?

graveland: I didn't like that either, but I also haven't used `OnInitialize` before so I didn't want to change things even more :) I'd prefer it being in `PersistenPreRunE` too.

graveland: panic at least is catchable compared to os.Exit :)

graveland: Taking it out of OnInitialize() seems to break the interaction between cobra/viper/pflags. I think panic is nicer than os.Exit at least.

internal/tiger/cmd/auth.go
```diff
@@ -242,13 +239,42 @@ func promptForCredentials(consoleURL string, creds credentials) (credentials, er
...
 }
+
+func readString(ctx context.Context, prompt string, args ...any) (string, error) {
+	valCh := make(chan string)
+	errCh := make(chan error)
+	defer func() { close(valCh); close(errCh) }()
+	go func() {
+		reader := bufio.NewReader(os.Stdin)
+		_, _ = fmt.Fprintf(os.Stdout, prompt, args...)
+		val, err := reader.ReadString('\n')
```
nathanjcochran: A couple small comments:

There are other cases where we read from stdin in a different manner. For example, we call `term.ReadPassword` above on line 232, but we'll want that to respect the context too. Can we therefore pass in the "read" function (e.g. `read func() (string, error)`) as an argument, to make this more generic?

Also, I'd remove the logic for prompting (and the `prompt` and `args` parameters) and leave that up to the caller, unless there's a good reason for it to be part of this function.

graveland: yeah... I had a for {} around it so it wouldn't continue with an empty value previously, so the prompt wanted to be passed.

internal/tiger/cmd/mcp.go
```diff
@@ -285,11 +275,12 @@ func startHTTPServer(ctx context.Context, host string, port int) error {
...
 	// called here and via the deferred call above.
 	<-ctx.Done()
-	stop()
 
 	// Shutdown server gracefully
 	logging.Info("Shutting down HTTP server...")
-	if err := httpServer.Shutdown(context.Background()); err != nil {
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer shutdownCancel()
+	if err := httpServer.Shutdown(shutdownCtx); err != nil {
```
nathanjcochran: I'm not sure we want a 5 second timeout here. MCP servers using the HTTP transport will keep long-running connections open until the user session (e.g. a Claude Code session connected to the MCP server) ends. I think the default behavior should be to wait for those user sessions to end, not to exit early and break them. The user can always do a second ctrl-c if they _do_ want to exit early and break them.

internal/tiger/cmd/service.go
nathanjcochran: This is another place where we probably want to show a different error message for `context.Canceled` now.

internal/tiger/cmd/mcp.go
graveland: We could, but how discoverable is control-C x 2? Control-C or sigterm means to shut down to me?

graveland: Just pushed, checkout the changes :)

internal/tiger/cmd/auth.go
```diff
@@ -243,12 +247,40 @@ func promptForCredentials(consoleURL string, creds credentials) (credentials, er
...
+
+	select {
+	case <-ctx.Done():
+		return "", ctx.Err()
+	case err := <-errCh:
+		return "", err
+	case val := <-valCh:
+		return val, nil
+	}
+}
```
nathanjcochran: I think there are some other places where we need to use this function too:

- https://github.com/timescale/tiger-cli/blob/b6ced91c10103dca753b17ac88a7f5c6a8bfa8f7/internal/tiger/cmd/service.go#L924
- https://github.com/timescale/tiger-cli/blob/b6ced91c10103dca753b17ac88a7f5c6a8bfa8f7/internal/tiger/cmd/db.go#L324

nathanjcochran: There are some other places where we read from stdin too, but they use bubbletea and handle ctrl-c explicitly, so I _think_ we should be good there:

- https://github.com/timescale/tiger-cli/blob/b6ced91c10103dca753b17ac88a7f5c6a8bfa8f7/internal/tiger/cmd/mcp_install.go#L372-L373
- https://github.com/timescale/tiger-cli/blob/b6ced91c10103dca753b17ac88a7f5c6a8bfa8f7/internal/tiger/cmd/oauth.go#L308-L309

nathanjcochran (approved):
Left a comment about a couple other places where we should use the new `readString` function, so that reading from stdin can be canceled via ctrl-c. Might want to move that `readString` function into the `util` package or something, given that we'll probably end up using it in a bunch of different places.

Other than that, this LGTM!

PR #85, opened 2025-10-23
Add initial serviceID and name completion

graveland:
It's a quick add for serviceID's in service commands. I haven't looked at other places it could be used yet, but there's no doubt a bunch. We should think about how the completion should be displayed as well (ie., show `serviceID(name)`? Mix serviceIDs and names in the same list?

### DIFF
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -72,6 +72,7 @@ Examples:
 
   # Get service details in YAML format
   tiger service get svc-12345 --output yaml`,
+		ValidArgsFunction: serviceIDCompletion,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Get config
 			cfg, err := config.Load()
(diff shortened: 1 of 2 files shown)


internal/tiger/cmd/service.go
```diff
@@ -1287,3 +1291,66 @@ Examples:
...
+	results := make([]string, 0, len(services))
+	for _, service := range services {
+		if service.ServiceId != nil && strings.HasPrefix(*service.ServiceId, toComplete) {
+			results = append(results, *service.ServiceId)
+		}
+	}
+	return results, cobra.ShellCompDirectiveNoFileComp
+}
+
+func serviceNameCompletion(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
nathanjcochran: To be clear, this function isn't used anywhere yet, right? Do we actually need it right now? Doesn't seem like it would be hard to add later if/when we need it.

graveland: Correct, I'll just zap it

PR #87, opened 2025-10-24
Add analytics tracking

nathanjcochran:
This PR adds support for tracking analytics events.

- Events are not sent if analytics is disabled via the `analytics` config option, the `--analytics` flag, the `TIGER_ANALYTICS` env var, or any of the following env vars: `DO_NOT_TRACK`, `NO_TELEMETRY`, `DISABLE_TELEMETRY`.
- Events are not sent if the user is not authenticated.
- We track one event per command execution or MCP tool call.
- Events for CLI commands are named: `Run <command>`. For example, `Run tiger service create`.
- Events for MCP tools are named: `Call <tool_name> tool`. For example, `Call service_create tool`.
- CLI events include any args that were passed (there are currently no commands that pass sensitive information in args)
- CLI events include any flags that were set, but explicitly exclude sensitive flags like `--password`, `--new-password`, etc.
- MCP events include tool arguments as properties, but explicitly exclude sensitive properties like `password`, `query`, etc.
- All events include a `success` property. If `false`, the event will also include an `error` property containing the error message.
- We do not currently track the state of the config or any env vars that were set. We could do this as a follow-up if we want, but I wanted to avoid tracking too much unnecessary information, at least to start with. Tracking just the explicit flags/arguments seems like a good place to start.

Closes [AGE-207](https://linear.app/tigerdata/issue/AGE-207/tiger-cli-introduce-some-kind-of-telemetry-for-usageerror-visibility)

Related PRs:
- https://github.com/timescale/savannah-public/pull/55

### DIFF
--- a/internal/tiger/cmd/root.go
+++ b/internal/tiger/cmd/root.go
@@ -135,7 +135,7 @@ func wrapCommandsWithAnalytics(cmd *cobra.Command) {
 	// Wrap this command's RunE if it exists
 	if cmd.RunE != nil {
 		originalRunE := cmd.RunE
-		cmd.RunE = func(c *cobra.Command, args []string) (err error) {
+		cmd.RunE = func(c *cobra.Command, args []string) (runErr error) {
 			start := time.Now()
 
 			defer func() {
@@ -155,7 +155,7 @@ func wrapCommandsWithAnalytics(cmd *cobra.Command) {
 					analytics.Property("args", args), // NOTE: Safe right now, but might need allow-list in the future if some args end up containing sensitive info
 					analytics.Property("elapsed_seconds", time.Since(start).Seconds()),
 					analytics.FlagSet(c.Flags()),
-					analytics.Error(err),
+					analytics.Error(runErr),
 				)
 			}()
 


internal/tiger/api/client_util.go
```diff
@@ -38,13 +38,8 @@ func getHTTPClient() *http.Client {
 	return sharedHTTPClient
 }
 
-// NewTigerClient creates a new API client with the given API key
-func NewTigerClient(apiKey string) (*ClientWithResponses, error) {
-	cfg, err := config.Load()
-	if err != nil {
-		return nil, fmt.Errorf("failed to load config: %w", err)
-	}
```
nathanjcochran, line 46: We were often loading the config multiple times, because this function loads it internally. I decided to pass it in as an argument instead, since it's usually already been loaded by the caller.

cevian:
We also need to check the following env variables: 
- DO_NOT_TRACK
- NO_TELEMETRY
- DISABLE_TELEMETRY

nathanjcochran:
> We also need to check the following env variables:
> 
>     DO_NOT_TRACK
>     NO_TELEMETRY
>     DISABLE_TELEMETRY

Done here: https://github.com/timescale/tiger-cli/pull/87/commits/e9494bc460029dfd1db3bcf2b5fd13e631512c41

cevian (approved):

PR #91, opened 2025-10-28
fix: preserve cli timeout

adolsalamanca:
Prevent using the API calls context in the wait commands,
as it includes a 30s timeout, which is honored even though
we're creating a child context with a 30m timeout later.

Rely on the cmd.Context() instead, so we preserve timeout behavior we want,
either 30m by default, or the one specified via flags to the cli.


nathanjcochran (commented):
LGTM! Thank you!!

PR #92, opened 2025-10-28
Fix flaky auth tests

nathanjcochran:
This fixes some flaky auth tests that were sometimes in failing in CI (for example, see [here](https://github.com/timescale/tiger-cli/actions/runs/18884195002/job/53895170617)).

The issue manifested as an unexpected EOF from the local OAuth callback server, which was causing the tests to fail [here](https://github.com/timescale/tiger-cli/blob/f0f657928abac4349de742045f1c4a621eac297f/internal/tiger/cmd/auth_test.go#L416-L420) when simulating a request from the user's browser:

```
--- FAIL: TestAuthLogin_OAuth_SingleProject (0.16s)
    auth_test.go:366: Mock browser opening URL: http://127.0.0.1:42739/oauth/authorize?client_id=45e1b16d-e435-4049-97b2-8daad150818c&code_challenge=QtzkvFiFpRa2QZlWnzHuvtn_Ad3qP3hCDl8wuKoXMkY&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A35441%2Fcallback&response_type=code&state=UsgJbg3qKUKKMhKgRKM1zY5wHZuv5Onp
    auth_test.go:419: Mock browser making callback request to: http://localhost:35441/callback?code=test-auth-code&state=UsgJbg3qKUKKMhKgRKM1zY5wHZuv5Onp
    auth_test.go:261: Mock server received token exchange request
    auth_test.go:423: Mock callback request failed: Get "http://localhost:35441/callback?code=test-auth-code&state=UsgJbg3qKUKKMhKgRKM1zY5wHZuv5Onp": EOF
```

I believe it was happening because we were previously doing a non-graceful shutdown of the local HTTP server that handles the callback request. The shutdown was happening after we had already [returned a redirect and returned the result on the resultChan](https://github.com/timescale/tiger-cli/blob/f0f657928abac4349de742045f1c4a621eac297f/internal/tiger/cmd/oauth.go#L207-L212), so the code looked valid. However, because the `http.ResponseWriter` buffers data before flushing it to the underlying TCP connection, and we were doing a hard/non-graceful shutdown, the TCP connection was sometimes being closed before the redirect response had actually been flushed, resulting in an EOF for the caller. I heard no reports of this happening outside of CI, but it's possible it could have happened in the wild too.

This fixes it by doing a graceful shutdown (which waits for any in-progress requests to complete) instead of a hard shutdown.

### DIFF
--- a/internal/tiger/cmd/auth_test.go
+++ b/internal/tiger/cmd/auth_test.go
@@ -15,7 +15,6 @@ import (
 	"regexp"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/timescale/tiger-cli/internal/tiger/config"
 )
@@ -410,10 +409,6 @@ func mockOpenBrowser(t *testing.T) func(string) error {
 
 		// Give the OAuth server a moment to start
 		go func() {
-			// Sleep to ensure the OAuth callback server is listening
-			// This prevents "EOF" errors in CI when the server hasn't started yet
-			time.Sleep(100 * time.Millisecond)
-
 			// Make the OAuth callback request directly
 			callbackURL := fmt.Sprintf("%s?code=test-auth-code&state=%s", redirectURI, state)
 			t.Logf("Mock browser making callback request to: %s", callbackURL)


internal/tiger/cmd/auth_test.go
```diff
@@ -410,10 +409,6 @@ func mockOpenBrowser(t *testing.T) func(string) error {
 
 		// Give the OAuth server a moment to start
 		go func() {
-			// Sleep to ensure the OAuth callback server is listening
-			// This prevents "EOF" errors in CI when the server hasn't started yet
-			time.Sleep(100 * time.Millisecond)
```
nathanjcochran, line 415: This was just a confused attempt to fix the issue previously, but isn't actually needed. Because we open the `net.Listener` synchronously [here](https://github.com/timescale/tiger-cli/blob/f0f657928abac4349de742045f1c4a621eac297f/internal/tiger/cmd/oauth.go#L124-L128), before we open the browser [here](https://github.com/timescale/tiger-cli/blob/f0f657928abac4349de742045f1c4a621eac297f/internal/tiger/cmd/oauth.go#L89-L91), it should never be the case that the server isn't ready to handle incoming connections by the time this function executes.

internal/tiger/cmd/oauth.go
```diff
@@ -162,26 +161,11 @@ func (l *oauthLogin) startOAuthServer(expectedState, codeVerifier string) (*oaut
...
-			return err
-		}
-		return nil
-	}
-
-	return errors.Join(
-		cls(s.server),
-		cls(s.listener),
-	)
-}
```
nathanjcochran, line 183: I decided to get rid of this function entirely and just call `http.Server.Shutdown()` directly above. The shutdown method already closes the listener internally, so we shouldn't need to do it ourselves, like it was previously being done here.

PR #93, opened 2025-10-28
Add `service_fork` MCP tool

nathanjcochran:
Adds a new `service_fork` MCP tool, which offers the same functionality as `tiger service fork`.

### DIFF
--- a/internal/tiger/password/connection.go
+++ b/internal/tiger/password/connection.go
@@ -48,11 +48,11 @@ func GetConnectionDetails(service api.Service, opts ConnectionDetailsOptions) (*
 		endpoint = service.Endpoint
 	}
 
-	if endpoint.Host == nil || *endpoint.Host == "" {
+	if endpoint == nil || endpoint.Host == nil || *endpoint.Host == "" {
 		return nil, fmt.Errorf("endpoint host not available")
 	}
 
-	if endpoint.Port == nil || *endpoint.Port == 0 {
+	if endpoint == nil || endpoint.Port == nil || *endpoint.Port == 0 {
 		return nil, fmt.Errorf("endpoint port not available")
 	}
 


internal/tiger/mcp/service_tools.go
```diff
@@ -273,11 +334,41 @@ WARNING: Creates billable resources.`,
...
+	// service_fork
+	mcp.AddTool(s.mcpServer, &mcp.Tool{
+		Name:  "service_fork",
+		Title: "Fork Database Service",
+		Description: `Fork an existing database service to create a new independent copy.
+
+You must specify a fork strategy:
+- 'NOW': Fork at the current database state (creates new snapshot or uses WAL replay)
+- 'LAST_SNAPSHOT': Fork at the last existing snapshot (faster fork)
+- 'PITR': Fork at a specific point in time (point-in-time recovery)
```
murrayju: This is probably redundant, I bet you could drop the second "point in time" with minimal impact

internal/tiger/mcp/service_tools.go
```diff
@@ -273,11 +334,41 @@ WARNING: Creates billable resources.`,
...
+
+By default:
+- Name will be auto-generated as '{source-service-name}-fork'
+- CPU and memory will be inherited from the source service
+- The forked service will be set as the default service
+
+Default behavior: Returns immediately while service provisions in background (recommended).
+Setting wait=true will block for several minutes until ready - only use if user explicitly needs immediate access.
```
murrayju: This wording (after the `-`) is confusing to me. I'd say something more along the lines of "only use if your next steps depends on database readiness to proceed"

internal/tiger/mcp/service_tools.go
```diff
@@ -510,6 +601,144 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
...
+			logging.Debug("MCP: Password storage failed", zap.Error(err))
+		} else {
+			logging.Debug("MCP: Password saved successfully", zap.String("method", result.Method))
+		}
+	}
+
+	// Include password in ServiceDetail if requested
+	if input.WithPassword {
+		output.Service.Password = util.Deref(service.InitialPassword)
+	}
```
murrayju, line 699: I think you could drop this, use details.Password below.

murrayju (approved):
Looks reasonable. I didn't test, can do tomorrow if you'd like.

A couple minor vibe-based thoughts.

internal/tiger/mcp/service_tools.go
nathanjcochran: Yeah, looks like Claude copied it from the [`tiger service fork --help` text](https://github.com/timescale/tiger-cli/blob/46c31510ced9eb71d618b02e17646f413fb6edf8/internal/tiger/cmd/service.go#L1065), where it's likewise redundant but perhaps make a little more sense because the flag is `--to-timestamp`, rather than `fork-strategy=PITR` (so there's nothing else indicating that it's technically using "point-in-time recovery"/"PITR" under the hood, which is a [legit Postgres concept](https://www.postgresql.org/docs/current/continuous-archiving.html) that's probably worth mentioning explicitly).

I removed the redundancy here and replaced it with a hint that the PITR strategy requires the `target_time` parameter, which was mentioned in one other place, but I think is probably worth being extra clear about: https://github.com/timescale/tiger-cli/pull/93/commits/5788da75f0222f1cc989999f8cd640a782254c8a.

Writing tool definitions is tough because it's so hard to measure the effect of small changes like this, but we can always improve more later.

internal/tiger/mcp/service_tools.go
nathanjcochran: Agreed. Looks like this one was copied from the [`service_create` tool description](https://github.com/timescale/tiger-cli/blob/5788da75f0222f1cc989999f8cd640a782254c8a/internal/tiger/mcp/service_tools.go#L329).

Updated both here: https://github.com/timescale/tiger-cli/pull/93/commits/b4009a43d333f6b9d85c30667c57f0a69eb4e1e3

internal/tiger/mcp/service_tools.go
nathanjcochran: Only reason I'm hesitant to do that is because `GetConnectionDetails` _can_ technically fail if the host or port isn't returned from the backend (see [here](https://github.com/timescale/tiger-cli/blob/b4009a43d333f6b9d85c30667c57f0a69eb4e1e3/internal/tiger/password/connection.go#L51-L57)), in which case it won't return the connection details (including the password). So doing it this way is slightly safer, and ensures users _always_ get the password back, so long as they specify `with_password=true` and the initial password is present in the API response.

PR #95, opened 2025-11-02
Add daily execution of all tests

josesahad:
Wires up the CI pipeline to run the integration tests in addition to the unit tests. Also configures it to run the tests at 8am daily, and to make them runnable via a manual workflow dispatch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,6 @@ jobs:
         env:
           TIGER_PUBLIC_KEY_INTEGRATION: ${{ secrets.TIGER_PUBLIC_KEY_INTEGRATION }}
           TIGER_SECRET_KEY_INTEGRATION: ${{ secrets.TIGER_SECRET_KEY_INTEGRATION }}
-          TIGER_PROJECT_ID_INTEGRATION: ${{ secrets.TIGER_PROJECT_ID_INTEGRATION }}
         run: go test -race -coverprofile=coverage.out -coverpkg=./... ./...
 
       - name: Generate coverage report


.github/workflows/test.yml
```diff
@@ -50,7 +53,26 @@ jobs:
...
       - name: Test
-        run: go test -race -coverprofile=coverage.out -coverpkg=./... ./...
+        env:
+          TIGER_PUBLIC_KEY_INTEGRATION: ${{ secrets.TIGER_PUBLIC_KEY_INTEGRATION }}
+          TIGER_SECRET_KEY_INTEGRATION: ${{ secrets.TIGER_SECRET_KEY_INTEGRATION }}
+          TIGER_PROJECT_ID_INTEGRATION: ${{ secrets.TIGER_PROJECT_ID_INTEGRATION }}
+        run: |
+          # Run all tests with race detection and coverage
+          # Integration tests will skip automatically if secrets are not set
+          go test -race -coverprofile=coverage.out -coverpkg=./... ./...
```
cevian: But the secrets will be set from the env: section above. So this will run all integration tests. I don't think you need any of the code below


This is the skip logic btw 
```
	publicKey := os.Getenv("TIGER_PUBLIC_KEY_INTEGRATION")
	secretKey := os.Getenv("TIGER_SECRET_KEY_INTEGRATION")
	projectID := os.Getenv("TIGER_PROJECT_ID_INTEGRATION")


	if publicKey == "" || secretKey == "" || projectID == "" {
		t.Skip("Skipping service not found test: TIGER_PUBLIC_KEY_INTEGRATION, TIGER_SECRET_KEY_INTEGRATION, and TIGER_PROJECT_ID_INTEGRATION must be set")
	}```

so just setting the env is enough. We don't need the .env file or script

File: .github/workflows/test.yml [hunk 2]
```diff
@@ -50,7 +53,26 @@ jobs:
         run: go vet ./...
 
       - name: Test
-        run: go test -race -coverprofile=coverage.out -coverpkg=./... ./...
+        env:
+          TIGER_PUBLIC_KEY_INTEGRATION: ${{ secrets.TIGER_PUBLIC_KEY_INTEGRATION }}
+          TIGER_SECRET_KEY_INTEGRATION: ${{ secrets.TIGER_SECRET_KEY_INTEGRATION }}
+          TIGER_PROJECT_ID_INTEGRATION: ${{ secrets.TIGER_PROJECT_ID_INTEGRATION }}
```
nathanjcochran: ~This repo is public, and this workflow runs on pull requests, including pull requests from forks (e.g. from external contributors). As a result, these new "secrets" are not really secret, and can be exfiltrated by anyone who creates a fork of the repo.~

~Are these credentials for an fully isolated test project that we wouldn't be worried about someone gaining access to and potentially abusing? If not, we should think of a different approach for running the integration tests.~

nathanjcochran: Actually, nevermind. I forgot that the [docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow) say:

> With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.

So I think this will actually work fine, it's just that the integration tests won't run when tests are triggered from a fork PR.

.github/workflows/test.yml
nathanjcochran: I went ahead and removed the unnecessary code: https://github.com/timescale/tiger-cli/pull/95/commits/d1f68aae0655b9a5df0c94f11480c9a3d7f15b04

.github/workflows/test.yml
```diff
@@ -50,7 +53,14 @@ jobs:
...
```
nathanjcochran: Have these secrets actually been created yet? I'm not seeing them listed in the [repository secrets](https://github.com/timescale/tiger-cli/settings/secrets/actions). Should probably do that before merging.

nathanjcochran: Done.

PR #103, opened 2025-11-07
Add service start/stop commands 

nathanjcochran:
This was branched off of https://github.com/timescale/tiger-cli/pull/16. See that PR for a complete description of the changes.

This PR contains more-or-less the same changes, except that a few things have been updated to adhere to our most current patterns (such as doing all error-handling via `exitWithErrorFromStatusCode` instead of handling individual status codes, using the spinner component to show updates when waiting for the service to be started/stopped, etc). I also addressed the code review feedback in that PR and merged in `main`.

One other change is that I decided to consolidate the `waitForServiceReady`, `waitForServicePaused`, and `waitForServiceDeletion` functions into a single configurable function that can wait for an arbitrary service state/status. There are still duplicate functions in the `mcp` package, but I intend to consolidate those in a follow-up PR as well.

CC @billy-the-fish for Tiger CLI docs update.

Closes [AGE-244](https://linear.app/tigerdata/issue/AGE-244/tiger-cli-finish-service-startstop-implementation).

### DIFF
--- a/internal/tiger/cmd/integration_test.go
+++ b/internal/tiger/cmd/integration_test.go
@@ -1501,3 +1501,477 @@ func TestAuthenticationErrorsIntegration(t *testing.T) {
 		})
 	}
 }
+
+// TestServiceForkIntegration tests forking a service with --now strategy and validates data is correctly copied
+func TestServiceForkIntegration(t *testing.T) {
+	config.SetTestServiceName(t)
+	// Check for required environment variables
+	publicKey := os.Getenv("TIGER_PUBLIC_KEY_INTEGRATION")
+	secretKey := os.Getenv("TIGER_SECRET_KEY_INTEGRATION")
+	projectID := os.Getenv("TIGER_PROJECT_ID_INTEGRATION")
+	if publicKey == "" || secretKey == "" || projectID == "" {
+		t.Skip("Skipping integration test: TIGER_PUBLIC_KEY_INTEGRATION, TIGER_SECRET_KEY_INTEGRATION, and TIGER_PROJECT_ID_INTEGRATION must be set")
+	}
+
+	// Set up isolated test environment with temporary config directory
+	tmpDir := setupIntegrationTest(t)
+	t.Logf("Using temporary config directory: %s", tmpDir)
+
+	// Generate unique names to avoid conflicts
+	timestamp := time.Now().Unix()
+	sourceServiceName := fmt.Sprintf("integration-fork-source-%d", timestamp)
+	tableName := fmt.Sprintf("fork_test_data_%d", timestamp)
+
+	var sourceServiceID string
+	var forkedServiceID string
+
+	// Always logout at the end to clean up credentials
+	defer func() {
+		t.Logf("Cleaning up authentication")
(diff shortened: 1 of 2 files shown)


internal/tiger/cmd/service.go
```diff
@@ -986,58 +957,218 @@ Examples:
...
-				continue
+			// Wait for service to become ready
+			fmt.Fprintf(statusOutput, "⏳ Waiting for service to start (wait timeout: %v)...\n", startWaitTimeout)
+			if err := waitForService(cmd.Context(), waitForServiceArgs{
+				client:    client,
+				projectID: projectID,
+				serviceID: serviceID,
+				handler: &statusWaitHandler{
+					targetStatus: "READY",
+					service:      &service,
```
nathanjcochran, line 1043: It's worth mentioning that there's an upstream bug that causes the status to be incorrectly reported as `QUEUED` at first, because that's what's coming back from the REST API in the initial endpoint response (i.e. it's what the `Status` field is set to on the `service` that gets passed in here). Once we start polling the "get service" endpoint, it's displayed correctly. I hope to fix that soon - see [AGE-243](https://linear.app/tigerdata/issue/AGE-243/tiger-cli-return-service-status-from-gateway-createforktoggle) for more info.

internal/tiger/cmd/integration_test.go
```diff
@@ -841,6 +841,182 @@ func TestServiceLifecycleIntegration(t *testing.T) {
...
+		var status string
+		if service.Status != nil {
+			status = string(*service.Status)
+		}
+
+		t.Logf("Service status after stop: %s", status)
+
+		// The status should be PAUSED for stopped services
+		if status != "PAUSED" {
+			t.Logf("Warning: Expected service status to be PAUSED, got %s. This might be expected depending on timing or service state.", status)
```
Askir: This might be expected? That seems like a weird error :D

internal/tiger/cmd/integration_test.go
```diff
@@ -841,6 +841,182 @@ func TestServiceLifecycleIntegration(t *testing.T) {
...
+		var status string
+		if service.Status != nil {
+			status = string(*service.Status)
+		}
+
+		t.Logf("Service status after start: %s", status)
+
+		// The status should be READY for started services
+		if status != "READY" {
+			t.Logf("Warning: Expected service status to be READY, got %s. This might be expected depending on timing or service state.", status)
```
Askir: Again a bit of an odd error message . I think we just remove the this might be expected part no?

internal/tiger/cmd/integration_test.go
nathanjcochran: Good call, fixed here: https://github.com/timescale/tiger-cli/pull/103/commits/a71be8dadc000e210a5c6fb32aaa0fbe190a0781

internal/tiger/cmd/integration_test.go
nathanjcochran: Fixed here: https://github.com/timescale/tiger-cli/pull/103/commits/a71be8dadc000e210a5c6fb32aaa0fbe190a0781

PR #105, opened 2025-11-13
Update homebrew cask caveats

nathanjcochran:
Updates the homebrew cask caveats and install script output to mention `tiger mcp install`.

Also fixes a bug where a multiline string provided for the `caveats` field in the `.goreleaser.yaml` would generate a Ruby cask file that looks like this:

```ruby
  caveats do
    "Tiger CLI has been installed successfully!"
    ""
    "Get started with:"
    "  tiger auth login"
    ""
    "For help:"
    "  tiger --help"
  end
```

The problem there is that only the final string (i.e. `"  tiger --help"`) is getting returned from that method (via Ruby's implicit return syntax). As a result, when installing the cask, only the final line is printed out:

```
==> Caveats
  tiger --help
```

This ultimately seems to be a bug in GoReleaser itself (see the templating code [here](https://github.com/goreleaser/goreleaser/blob/c3bfb5a5b75bd2225c5237b686853d3e85b81c28/internal/pipe/cask/templates/cask.rb#L88-L94)). For now, I worked around it by using `/n` for newline characters instead of using a normal YAML multiline string, which generates the following Ruby output:

```ruby
  caveats do
    "Tiger CLI has been installed successfully!\n\nGet started with:\n  tiger auth login\n  tiger mcp install\n\nFor help:\n  tiger help"
  end
```

This _should_ look like the following when it's printed out upon installing the cask (though it's admittedly somewhat hard to test without doing a real release):

```
Tiger CLI has been installed successfully!

Get started with:
  tiger auth login
  tiger mcp install

For help:
  tiger help
```

Note that I also updated it to recommend `tiger help` instead of `tiger --help`, since they produce identical output, but `tiger help` is a little more terse.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -453,8 +453,9 @@ main() {
     # Show usage information
     log_success "Get started with:"
     log_success "    ${BINARY_NAME} auth login"
+    log_success "    ${BINARY_NAME} mcp install"
     log_success "For help:"
-    log_success "    ${BINARY_NAME} --help"
+    log_success "    ${BINARY_NAME} help"
     log_success "Happy coding with Tiger CLI! 🐅"
 }
 


.goreleaser.yaml
```diff
@@ -129,7 +129,7 @@ homebrew_casks:
     description: |-
       Tiger CLI - Tiger Cloud Platform CLI
       Command-line interface for managing Tiger Cloud platform resources
-    binary: tiger
+    binaries: [ tiger ]
```
nathanjcochran, line 132: Addresses a GoReleaser deprecation warning. See https://goreleaser.com/deprecations/#homebrew_casksbinary.

PR #143, opened 2026-04-06
feat: migrate service logs to cursor-based pagination

AdrianLC:
## Summary

- Rewrites `FetchServiceLogs` to use cursor-based pagination — `lastCursor` from each response is passed back as `cursor` on the next request, removing the Elasticsearch 10,000-hit window limit that silently capped offset pagination
- `FetchServiceLogs` now returns `[]api.ServiceLogEntry` (structured) instead of `[]string`; each entry carries `timestamp`, `message`, and `severity` from Elasticsearch
- Adds `--since` flag (lower-bound timestamp) to `tiger service logs` alongside the existing `--until`
- Adds `Since` field to `ServiceLogsInput` in the MCP `service_logs` tool; MCP output returns structured `entries` with timestamp, message, and severity (no redundant flat `logs` array)
- `openapi.yaml` updated: adds `since`, `cursor` params; `ServiceLogEntry` schema; `entries` field on `ServiceLogs`; `page` param marked deprecated; regenerated `types.go` and `client.go`

## Merge order

Depends on savannah-public adding `entries` and `lastCursor` to the response (timescale/savannah-public#88), which depends on timescale/timescale-graphql-client#18. Merge those first.

### DIFF
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -1593,8 +1593,8 @@ Examples:
 					color.NoColor = false
 				}
 
-				for _, log := range logs {
-					colorizedLog := colorizeLogLine(log, shouldColorize)
+				for _, entry := range logs {
+					colorizedLog := colorizeLogLine(entry.Message, shouldColorize)
 					fmt.Fprintln(outputWriter, colorizedLog)
 				}
 			}
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -1495,7 +1495,7 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 // buildServiceLogsCmd creates the logs command for viewing service logs
 func buildServiceLogsCmd() *cobra.Command {
 	var tail int
-	var from time.Time
+	var since time.Time
 	var until time.Time
 	var node int
 	var output string
@@ -1520,7 +1520,7 @@ Examples:
   tiger service logs svc-12345
 
   # View logs within a time range
-  tiger service logs --from "2024-01-15T09:00:00Z" --until "2024-01-15T10:00:00Z"
+  tiger service logs --since "2024-01-15T09:00:00Z" --until "2024-01-15T10:00:00Z"
 
   # View logs for a specific node (for services with HA replicas)
   tiger service logs --node 1
@@ -1550,9 +1550,9 @@ Examples:
 			cmd.SilenceUsage = true
 
 			// Prepare parameters
-			var fromPtr *time.Time
-			if !from.IsZero() {
-				fromPtr = &from
+			var sincePtr *time.Time
+			if !since.IsZero() {
+				sincePtr = &since
 			}
 
 			var untilPtr *time.Time
@@ -1571,7 +1571,7 @@ Examples:
 			ctx, cancel := context.WithTimeout(cmd.Context(), time.Minute)
 			defer cancel()
 
-			logs, err := common.FetchServiceLogs(ctx, cfg, serviceID, tail, fromPtr, untilPtr, nodePtr)
+			logs, err := common.FetchServiceLogs(ctx, cfg, serviceID, tail, sincePtr, untilPtr, nodePtr)
 			if err != nil {
 				return err
 			}
@@ -1609,7 +1609,7 @@ Examples:
 
 	// Add flags
 	cmd.Flags().IntVar(&tail, "tail", 100, "Number of log lines to show")
-	cmd.Flags().TimeVar(&from, "from", time.Time{}, []string{time.RFC3339}, "Fetch logs after this timestamp (RFC3339 format, e.g., 2024-01-15T09:00:00Z)")
+	cmd.Flags().TimeVar(&since, "since", time.Time{}, []string{time.RFC3339}, "Fetch logs after this timestamp (RFC3339 format, e.g., 2024-01-15T09:00:00Z)")
 	cmd.Flags().TimeVar(&until, "until", time.Time{}, []string{time.RFC3339}, "Fetch logs before this timestamp (RFC3339 format, e.g., 2024-01-15T10:00:00Z)")
 	cmd.Flags().IntVar(&node, "node", 0, "Specific service node to fetch logs from (for services with HA replicas, 0 is valid)")
 	cmd.Flags().VarP((*outputFlag)(&output), "output", "o", "Output format (text, json, yaml)")
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -1495,6 +1495,7 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 // buildServiceLogsCmd creates the logs command for viewing service logs
 func buildServiceLogsCmd() *cobra.Command {
 	var tail int
+	var from time.Time
 	var until time.Time
 	var node int
 	var output string
@@ -1506,7 +1507,7 @@ func buildServiceLogsCmd() *cobra.Command {
 		Long: `View logs for a database service.
 
 Fetches and displays logs from the specified service. By default, shows the last
-100 log entries. Supports filtering by time.
+100 log entries. Supports filtering by time range.
 
 The service ID can be provided as an argument or will use the default service
 from your configuration.
@@ -1518,8 +1519,8 @@ Examples:
   # View logs for specific service
   tiger service logs svc-12345
 
-  # View logs before a specific time
-  tiger service logs --until "2024-01-15T10:00:00Z"
+  # View logs within a time range
+  tiger service logs --from "2024-01-15T09:00:00Z" --until "2024-01-15T10:00:00Z"
 
   # View logs for a specific node (for services with HA replicas)
   tiger service logs --node 1
@@ -1549,6 +1550,11 @@ Examples:
 			cmd.SilenceUsage = true
 
 			// Prepare parameters
+			var fromPtr *time.Time
+			if !from.IsZero() {
+				fromPtr = &from
+			}
+
 			var untilPtr *time.Time
 			if !until.IsZero() {
 				untilPtr = &until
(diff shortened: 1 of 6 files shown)


openapi.yaml
```diff
@@ -1331,7 +1361,15 @@ components:
...
-          description: Array of log entries (up to 500 entries per page, may be empty)
+          description: Array of log message strings. Preserved for backwards compatibility.
+        entries:
+          type: array
+          items:
+            $ref: '#/components/schemas/ServiceLogEntry'
+          description: Structured log entries with timestamp and severity metadata. Only present on the cursor-based path.
+        lastCursor:
+          type: string
+          description: Opaque cursor for the next page of results. Present when more log entries exist older than the last entry in this response. Absent when there are no further results.
```
Askir, line 1373: If you do the rename you'll have to change it here too.

openapi.yaml
Askir, line 604: Maybe remove this one here already? That should help with avoiding anyone using it in the future since this repo is public, it also means some of the old generated code is deleted.

internal/tiger/common/logs.go
```diff
@@ -54,29 +53,38 @@ func FetchServiceLogs(
...
+// to constructing minimal entries from the plain logs strings (legacy path).
+func toEntries(logs *api.ServiceLogs) []api.ServiceLogEntry {
+	if logs.Entries != nil {
+		return *logs.Entries
+	}
+	entries := make([]api.ServiceLogEntry, len(logs.Logs))
+	for i, msg := range logs.Logs {
+		entries[i] = api.ServiceLogEntry{Message: msg}
+	}
+	return entries
```
Askir: Do we need the fallback here once the API has changed and we no longer use the `page` param?

internal/tiger/mcp/service_tools.go
```diff
@@ -1158,21 +1172,33 @@ func (s *Server) handleServiceLogs(ctx context.Context, req *mcp.CallToolRequest
...
+			Timestamp: e.Timestamp,
+			Message:   e.Message,
+			Severity:  e.Severity,
+		}
+	}
+
 	output := ServiceLogsOutput{
-		Logs: logs,
+		Logs:    logs,
+		Entries: structured,
```
Askir: It seems a bit duplicate to return both the logs string arary as well as the structured entries. I think it is okay to just just remove the logs here, otherwise this bloats the context window a lot if the LLM calls this function even once.

Also we do the type mapping to the MCP type usually in utils.go I think, rather than inline.

internal/tiger/mcp/service_tools.go
Askir: This could maybe be slightly tweaked to mention from/until?

internal/tiger/mcp/service_tools.go
```diff
@@ -383,6 +384,9 @@ func (ServiceLogsInput) Schema() *jsonschema.Schema {
 	schema.Properties["tail"].Minimum = util.Ptr(1.0)
 	schema.Properties["tail"].Examples = []any{50, 100, 1000}
 
+	schema.Properties["from"].Description = "Fetch logs after this timestamp (RFC3339 format, e.g., '2024-01-15T09:00:00Z'). Lower bound for the log search window."
+	schema.Properties["from"].Examples = []any{"2024-01-15T09:00:00Z", "2025-01-16T08:00:00Z"}
+
```
Askir, line 389: This should probably mention the until param and how to use them together.

internal/tiger/cmd/service.go
```diff
@@ -1587,8 +1593,12 @@ Examples:
...
 				}
 
-				for _, log := range logs {
-					colorizedLog := colorizeLogLine(log, shouldColorize)
+				for _, entry := range logs {
+					line := entry.Message
+					if !entry.Timestamp.IsZero() {
+						line = entry.Timestamp.UTC().Format("2006-01-02 15:04:05 UTC") + " " + line
+					}
+					colorizedLog := colorizeLogLine(line, shouldColorize)
```
Askir, line 1601: Can we change this to use the passed down `severity` field instead of doing string comparison? That should get it consistent with the frontend too.

Askir (approved):
I'll approve already, but I think there is a few things that we should fix before doing the release. Mainly some MCP cleanup,

I didn't add a comment for this but I think there are some mocks for the logs api responses that have the old format still, I think we want to update those to the new one.

I also think you can remove all the fallback logic here, and just cleanly switch to the new format.

openapi.yaml
nathanjcochran: I agree in theory, but I feel like it's easier to keep the files in sync. If we remove `page` here, but not upstream, then people can't just copy the files to sync them in the future (and Claude will constantly be trying to put the `page` parameter back). I mentioned elsewhere that I think it's possible to mark a parameter/field as deprecated with `deprecated: true`, so that might be the best middle-ground here.

internal/tiger/common/logs.go
nathanjcochran: I agree, I don't think we need this. If we deploy the updated server code first, there shouldn't be any situation where the new CLI is hitting the old server endpoint that only returns the `logs` field.

internal/tiger/mcp/service_tools.go
nathanjcochran: I agree that we don't want duplicate content in the MCP response. But on that note, do we need to return the structured entries at all? Doesn't the `message` field already contain the severity and timestamp (or am I misunderstanding)? In other words, aren't `timestamp` and `severity` already kind of duplicative with the `message`? I think LLMs are pretty good at understanding logs in their normal text format (they've seen lots of them in the training data), so I'm not sure it's really helpful for us to split it out into separate fields anyways. And just returning the log messages as a string array massively cuts down on the size of the response (no JSON objects with nested keys for each log message, etc.)

Also, I think it's fine for some of the conversions to happen inline. Especially if the conversion functions aren't re-used (I think most of the ones in `utils.go` are there because they're reused across multiple MCP tool implementations).

internal/tiger/cmd/service.go
```diff
@@ -1599,6 +1609,7 @@ Examples:
 
 	// Add flags
 	cmd.Flags().IntVar(&tail, "tail", 100, "Number of log lines to show")
+	cmd.Flags().TimeVar(&from, "from", time.Time{}, []string{time.RFC3339}, "Fetch logs after this timestamp (RFC3339 format, e.g., 2024-01-15T09:00:00Z)")
```
nathanjcochran: As mentioned in my comment on the other PR, I think maybe this should be `--since` instead of `--from`.

internal/tiger/mcp/service_tools.go
nathanjcochran: Hmm. I don't know if it needs to mention `until`. That parameter has its own description, and I think it's clear enough for an LLM to understand how it would work with `from`.

My main comment is that "Lower bound for the log search window." seems unnecessary. "Fetch logs after this timestamp" seems clear enough. But I do think it would be good to include information about what happens if this parameter is not provided (to make it better match the `until` description).

internal/tiger/cmd/service.go
```diff
@@ -1495,6 +1495,7 @@ Note: You can specify both CPU and memory together, or specify only one (the oth
 // buildServiceLogsCmd creates the logs command for viewing service logs
 func buildServiceLogsCmd() *cobra.Command {
 	var tail int
+	var from time.Time
```
AdrianLC: rename to since to match rename in. public-api

internal/tiger/cmd/service.go
AdrianLC: yes but IMO let's do a separate PR

internal/tiger/common/logs.go
AdrianLC: ok good point. I was concerned in case of rollbacks and such but it makes no sense for a CLI release so I'll get rid of this

internal/tiger/mcp/service_tools.go
AdrianLC: If it's not provided we are limited only by page sizes up to 50_000 logs or more. So when the loop finishes with the tail parameter

internal/tiger/mcp/service_tools.go
AdrianLC: The message won't contain the Timestamp which is a separate field now. It still contains the severity but it's likely to change if we finish the move to full structured json logs. 

I'll keep only the new structure then.

aprimakina (approved):

internal/tiger/mcp/service_tools.go
nathanjcochran: > The message won't contain the Timestamp which is a separate field now. 

Hmm. Okay. I think keeping just the new structure is definitely better than keeping both, but I do still think this might be sub-optimal for LLMs.

I really think the "best" format for LLMs is probably closer to the native text format. Partly because that's just what LLMs have seen most often in the training data, but also because it's significantly fewer tokens, due to the fact that the content is _just_ the log message text instead of a bunch of JSON objects, each with a `timestamp`, `message`, and `severity` field (which takes up a lot of tokens when duplicated for each log message, and probably doesn't really help the LLM understand the content in any meaningful way).

So imo, it might be better to just prepend the timestamp (like you're doing for the CLI output) and return an array of strings.

> It still contains the severity but it's likely to change if we finish the move to full structured json logs.

Wait, wouldn't this be a backwards-compatibility issue? If you removed the severity from the `message` field, the way the logs are currently being formatted for the sake of the CLI output would break for existing versions of Tiger CLI, in that sense that the severity would no longer be output in each message (since we aren't prepending that field to the message before outputting it, like we do with the timestamp). We should probably figure out our plan for this now, if that's actually something you think you're going to do in the near future...

nathanjcochran: Since the MCP tool call output format can be changed after the fact without affecting backwards-compatibility (we don't need the MCP output to be backwards-compatible, since it's just read by an LLM), I'll go ahead and merge this now, and we can continue to think about it and potentially change it later.

AdrianLC: Currently we are not planning to strip the severity out of messages. If we did we'd prepend it here first for compatibility, like we did with the Timestamp. But it's not planned yet

I can easily open another PR to use the single phrase output for MCP to save tokens. Let's do that along with the colorizing change in https://github.com/timescale/tiger-cli/pull/144/ before creating a new release

PR #144, opened 2026-04-09
feat(logs): colorize log lines using structured severity field

AdrianLC:
## Summary

- Replaces the regex-based log level colorizer with one that uses the `severity` field from the structured `ServiceLogEntry`
- Only the severity token (e.g. `ERROR:`) is colorized within the line — not the full line — preserving the original behaviour that's easier on the eyes across different terminal color schemes
- Using the API-provided severity field eliminates false matches: the old regex would colorize any word matching a severity name anywhere in the message body (e.g. a query mentioning `WARNING` or `ERROR`), whereas now only the actual severity of that log entry is colorized
- Severity levels follow the PostgreSQL spec: https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS

## Merge order

Depends on timescale/tiger-cli#143 (which introduces the structured `ServiceLogEntry` with the `severity` field). Merge that first.

### DIFF
--- a/internal/tiger/cmd/service.go
+++ b/internal/tiger/cmd/service.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"regexp"
 	"strings"
 	"time"
 
@@ -1598,8 +1597,7 @@ Examples:
 					if !entry.Timestamp.IsZero() {
 						line = entry.Timestamp.UTC().Format("2006-01-02 15:04:05 UTC") + " " + line
 					}
-					colorizedLog := colorizeLogLine(line, shouldColorize)
-					fmt.Fprintln(outputWriter, colorizedLog)
+					fmt.Fprintln(outputWriter, colorizeLogEntry(line, entry.Severity, shouldColorize))
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


internal/tiger/cmd/service.go
```diff
@@ -1617,47 +1615,30 @@ Examples:
...
 
-// logLevelRegex matches PostgreSQL log levels in log lines
-// Pattern: word boundary + log level + colon
-var logLevelRegex = regexp.MustCompile(`\b(DEBUG|INFO|NOTICE|WARNING|LOG|ERROR|FATAL|PANIC|DETAIL|HINT|QUERY|CONTEXT|LOCATION):`)
-
-// colorizeLogLine applies color to the log level in a PostgreSQL log line
-// If colorEnabled is false, returns the line unchanged
-func colorizeLogLine(line string, colorEnabled bool) string {
+// colorizeLogEntry applies a color to the entire log line based on the
+// structured severity field. If colorEnabled is false, returns the line unchanged.
```
nathanjcochran: I intentionally did _not_ colorize the full log line in the original implementation. Some of the colors are just too hard on the eyes when the full log line is colored, depending on the user's terminal color scheme (which we don't really have control over). Coloring only the log level/severity makes it less of a problem, imo (i.e. even if `INFO` is hard to read, the rest of the log line isn't, so it's not a big deal). I think we should keep the original behavior of only coloring the severity.

nathanjcochran: But honestly, given that, I'm not even sure it's worth updating this to use the severity field now, since we have to extract the severity from the message anyways to colorize it.

Though maybe passing the severity in is a more deterministic way of figuring out what the "correct" severity is for that line, and therefore which string to search for and colorize? So I guess that would still be an improvement (though minor)? Your call.

AdrianLC: I changed it to color only the severity word while still matching with the API provided info.
It's a smaller improvement but at least we can avoid coloring when a word happens by coincidence in the message.

I also removed a bunch of words that are not really part of the log severities. Those happen but are part of the same message and are included inside in multiline messages now.

nathanjcochran (approved):

PR #145, opened 2026-04-09
feat(mcp): return logs as pre-formatted strings to reduce token usage

AdrianLC:
## Summary

Improves on the structured log entries introduced in #143. That PR returned an `entries` array of JSON objects, each with `timestamp`, `message`, and `severity` fields. This PR replaces that with a flat `logs []string` where each line is an RFC3339 timestamp prepended to the log message.

The severity is already embedded in the message text so there is no information loss, but the token count per MCP tool call drops significantly — a JSON object with three named fields per entry becomes a single string.

### DIFF
--- a/internal/tiger/mcp/service_tools.go
+++ b/internal/tiger/mcp/service_tools.go
@@ -395,14 +395,7 @@ func (ServiceLogsInput) Schema() *jsonschema.Schema {
 
 // ServiceLogsOutput represents output for service_logs
 type ServiceLogsOutput struct {
-	Entries []serviceLogEntryOutput `json:"entries" jsonschema:"Structured log entries with timestamp, message, and severity, ordered from oldest to newest"`
-}
-
-// serviceLogEntryOutput is the MCP output shape for a single log entry.
-type serviceLogEntryOutput struct {
-	Timestamp time.Time `json:"timestamp"`
-	Message   string    `json:"message"`
-	Severity  string    `json:"severity"`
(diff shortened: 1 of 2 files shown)


nathanjcochran (approved):
LGTM! Thanks for making this change!

internal/tiger/mcp/service_tools.go
```diff
@@ -1184,14 +1177,14 @@ func (s *Server) handleServiceLogs(ctx context.Context, req *mcp.CallToolRequest
...
 
-	structured := make([]serviceLogEntryOutput, len(entries))
+	logs := make([]string, len(entries))
 	for i, e := range entries {
-		structured[i] = serviceLogEntryOutput{
-			Timestamp: e.Timestamp,
-			Message:   e.Message,
-			Severity:  e.Severity,
+		if !e.Timestamp.IsZero() {
+			logs[i] = e.Timestamp.UTC().Format(time.RFC3339) + " " + e.Message
```
nathanjcochran: Why are we using a different timestamp format here compared to [in the CLI command](https://github.com/timescale/tiger-cli/blob/61ddc7c6b4b6e68f7ac80071c9309fd22abb3f30/internal/tiger/cmd/service.go#L1599) and [API](https://github.com/timescale/savannah-public/blob/81ec0a85969f265dd327b3b35ae561285b2098c7/internal/handlers/services.go#L789)? Shouldn't we use the same format everywhere, for consistency?

AdrianLC: Just Claude doing it wrong. 

I changed it but made the CLI use local timezone I think that's better in that case?
Let me know if you disagree though and we can change it to UTC too

nathanjcochran (approved):

PR #146, opened 2026-04-10
Fix integration tests for structured logs and fork service limits

Askir:
## Summary
- **ServiceLogs test**: Updated JSON unmarshalling from `[]string` to `[]struct{message, severity, timestamp}` to match the structured log entries introduced in the cursor-based pagination migration (#143).
- **Fork tests**: Replaced `ForkService_LastSnapshot_NoBackupsYet` (which expected `--last-snapshot` to fail on new services) with `ForkService_LastSnapshot_EarlyFork` that expects success and deletes the fork afterwards, freeing the service slot for subsequent fork tests.


nathanjcochran (approved):
LGTM! Thank you!!

PR #151, opened 2026-05-07
add read-only mode to refuse write tool calls

aprimakina:
Adds a `read_only` config option that:
  - Refuses write/destructive MCP tools (service_create, _fork, _start, _stop, _resize, _update_password) with a clear error message.
  - Injects `default_transaction_read_only=on` into the DB connection GUCs so `db_execute_query` and `tiger db connect` can't perform writes.

Closes https://linear.app/tigerdata/issue/AGE-422/finish-tiger-cli-mcp-read-only-permissions-design


internal/tiger/mcp/errors.go
```diff
@@ -0,0 +1,46 @@
...
+// gate itself stays correct because handlers reload config per call.
+func buildServerInstructions(cfg *config.Config) string {
+	if cfg == nil || !cfg.ReadOnly {
+		return ""
+	}
+	return "READ-ONLY MODE IS ENABLED. The following Tiger MCP tools will refuse to run: " +
+		strings.Join(readOnlyGatedTools, ", ") + ". " +
+		"Before asking the user to provide inputs for any of these operations, " +
+		"tell them read-only mode is on and ask whether to disable it " +
+		"(run: `tiger config unset read_only`) or skip the operation."
```
nathanjcochran: I'm not convinced it's a good idea to tell the agent the command to use to disable read-only mode like this.

specs/spec_mcp.md
```diff
@@ -59,6 +59,24 @@ Alternatively, for manual configuration, the Tiger MCP server can be added to yo
...
+
+Tools gated by read-only mode:
+- `service_create`
+- `service_fork`
+- `service_start`
+- `service_stop`
+- `service_resize`
+- `service_update_password`
+
+`db_execute_query` is intentionally not gated; database-level read-only enforcement is a separate concern (see the `tsdb_admin.read_only_role` mechanism documented under `tiger db create role --read-only`).
```
nathanjcochran: I don't think this is right. I think we definitely _do_ want the read-only mode to apply to database queries. That's why we went through the effort of creating the [read-only connection string parameter](https://github.com/timescale/protected_roles/pull/19) that locks the connection into read-only transaction mode. See [how Ghost is doing it](https://github.com/timescale/ghost/blob/dbebf54f131ddf354cef526a43aeaef02073ab9f/internal/common/connection.go#L35-L41) for reference.

Relying on `tiger db create role --read-only` is really weak because you have to _ask_ the agent to use the read-only role you created - there's no way to really lock it down to a particular role (the role is just a parameter to the `db_execute_query` MCP tool).

internal/tiger/mcp/errors.go
aprimakina: What is your concern? Are you suggesting saying something more generic, like "disable the read-only mode or skip the operation"?

specs/spec_mcp.md
aprimakina: Sure, the changes in the current branch are just the 1st step. I plan to add the read-only mode for the DB operations in a separate PR.

internal/tiger/mcp/errors.go
nathanjcochran: My concern is that if you tell the agent how to disable read-only mode, it could just execute that bash command itself (if the user gave it permission to, or it's running in yolo mode) to break out and gain write access, which kind of defeats the point. Granted, that's technically _always_ possible, I just don't think we should make it so easy by telling agents exactly which command to run. So yeah, I think "disable the read-only mode or skip the operation" would be better.

aprimakina: I've updated the message in https://github.com/timescale/tiger-cli/pull/151/changes/56cf611cb0b398bce824b3d90965f66915ec4c10, though the agent can still easily pick up the required command. I don't think we can really hide it.

```
⏺ Read-only mode is currently enabled, which prevents service_create from running.                                                                                                                         
                                                                                                                                                                                                         
  Would you like to:                                                                                                                                                                                       
  1. Disable read-only mode and proceed — run tiger config unset read_only in your terminal
  2. Skip the operation 
```

internal/tiger/cmd/config.go
```diff
@@ -212,6 +212,9 @@ func outputTable(w io.Writer, cfg *config.ConfigOutput) error {
 	if cfg.PasswordStorage != nil {
 		table.Append("password_storage", *cfg.PasswordStorage)
 	}
+	if cfg.ReadOnly != nil {
+		table.Append("read_only", fmt.Sprintf("%t", *cfg.ReadOnly))
```
Askir, line 216: thoughts about specifying that this is mcp read only? I feel like setting a flag on the CLI that's called read-only might confuse people if it only impact the mcp?

Askir (approved):
Looks good generally. But I wouldn't wanna release this without the sql read-only feature I think. I'm a bit scared of someone enabling read only and then expecting to be safe with the MCP but in reality they aren't. So I'd merge those in at least close together as we are otherwise blocking the release process.

PR #154, opened 2026-05-11
test: sweep stale services

aprimakina:
- New helper `sweepStaleIntegrationServices` deletes any service named `integration-*` older than 1 hour.                                
  - Called once per CI run, near the end of `TestServiceLifecycleIntegration` (<s>before its `Logout` subtest, while creds are still live</s> after `Login`).

`go test`'s outer 15m alarm, SIGKILL, or job cancellation kills the test process before deferred cleanups can run, leaving orphan services on the account. This sweeper catches those orphans on the next run. The 1-hour threshold protects in-flight services from concurrent runs.

### DIFF
--- a/internal/tiger/cmd/integration_test.go
+++ b/internal/tiger/cmd/integration_test.go
@@ -93,15 +93,20 @@ func executeIntegrationCommand(ctx context.Context, args ...string) (string, err
 	return buf.String(), err
 }
 
+// testServicePrefix is the name prefix used by every service created by the
+// integration tests. The sweeper filters by this prefix so it only deletes
+// services this test suite owns.
+const testServicePrefix = "integration-"
+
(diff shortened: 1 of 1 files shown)


internal/tiger/cmd/integration_test.go
```diff
@@ -161,6 +215,10 @@ func TestServiceLifecycleIntegration(t *testing.T) {
 		t.Logf("Login successful")
 	})
 
+	// Sweep orphans from prior runs while we're authenticated and before
+	// the test does anything that could fail. Runs once per CI run.
+	sweepStaleIntegrationServices(t, time.Hour)
```
Askir, line 220: PR description says at the end of the integration run but this looks like the start (which also makes more sense imo :D) 

So nothing to change just wanted to flag that the description is confusing.

Askir (approved):

PR #155, opened 2026-05-11
feat: read only mode guc injection

aprimakina:
When read-only mode is enabled, the `db_execute_query` MCP tool and the new `--read-only` CLI flag open the database connection with Tiger Cloud's `tsdb_admin.read_only_connection=true`. The GUC is immutable for the session, so writes and DDL are rejected by the server regardless of the SQL the caller (or LLM) sends.

### DIFF
--- a/internal/tiger/mcp/db_tools.go
+++ b/internal/tiger/mcp/db_tools.go
@@ -116,6 +116,8 @@ Connects to a PostgreSQL database service in Tiger Cloud and executes the provid
 
 Multi-statement queries (semicolon-separated) are supported when no parameters are provided. All result sets are returned. By default, statements execute in an implicit transaction that automatically commits on success or rolls back on error. Explicit transactions (opened with BEGIN) must be explicitly committed with COMMIT, or they roll back when the connection closes.
 
+When read-only mode is enabled in the Tiger CLI configuration, the database connection is opened with an immutable read-only GUC, so writes and DDL will be rejected by the server regardless of the query contents.
+
 WARNING: Can execute any SQL statement including INSERT, UPDATE, DELETE, and DDL commands. Always review queries before execution.`,
 		InputSchema:  DBExecuteQueryInput{}.Schema(),
 		OutputSchema: DBExecuteQueryOutput{}.Schema(),
@@ -146,6 +148,7 @@ func (s *Server) handleDBExecuteQuery(ctx context.Context, req *mcp.CallToolRequ
 		zap.Duration("timeout", timeout),
 		zap.String("role", input.Role),
 		zap.Bool("pooled", input.Pooled),
+		zap.Bool("read_only", cfg.ReadOnly),
 	)
 
 	// Get service details to construct connection string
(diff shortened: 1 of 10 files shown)


internal/tiger/mcp/db_tools.go
```diff
@@ -116,6 +116,8 @@ Connects to a PostgreSQL database service in Tiger Cloud and executes the provid
 
 Multi-statement queries (semicolon-separated) are supported when no parameters are provided. All result sets are returned. By default, statements execute in an implicit transaction that automatically commits on success or rolls back on error. Explicit transactions (opened with BEGIN) must be explicitly committed with COMMIT, or they roll back when the connection closes.
 
+When read-only mode is enabled in the Tiger CLI configuration, the database connection is opened with an immutable read-only GUC, so writes and DDL will be rejected by the server regardless of the query contents.
```
Askir: I think it's again important to mention here that this is for the MCP only. If e.g. claude code uses `bash: tiger db connect` it bypasses the readonly mode. Claude probably needs to know that and so should the user.

Askir (approved):
looks good!

internal/tiger/mcp/db_tools.go
nathanjcochran: I mean, this is an MCP tool description. Doesn't that make it clear?

Though personally, I'm skeptical that we even need to tell the agent about this. It isn't supposed to modify the read-only config option itself, so why does it need to know about it? I'd be concerned about giving it too many clues about how to disable read-only mode, since we _really_ don't want it to try to do that.

If we just want to make sure it doesn't get confused if there's an error due to read-only mode being set, I think we could at least trim this down and save some tokens (i.e. why does it need to mention that it's a GUC? That seems like an implementation detail the agent doesn't need to know).

internal/tiger/mcp/errors.go
```diff
@@ -41,5 +41,7 @@ func buildServerInstructions(cfg *config.Config) string {
 	return "READ-ONLY MODE IS ENABLED. The following Tiger MCP tools will refuse to run: " +
 		strings.Join(readOnlyGatedTools, ", ") + ". " +
 		"Before asking the user to provide inputs for any of these operations, " +
-		"tell them read-only mode is on."
+		"tell them read-only mode is on. " +
+		"db_execute_query is still available, but the database connection is opened with " +
+		"an immutable read-only GUC, so writes and DDL will be rejected by the server."
```
nathanjcochran: Fwiw, the MCP server instructions should probably include more than just info about read-only mode being enabled. For example, this is what Ghost's instructions say: https://github.com/timescale/ghost/blob/a6c83e824303eeac4605950ce0fc3fdc98c819c2/internal/mcp/server.go#L35-L37.

The main benefit of this is that some agents (e.g. Claude Code) will hide tool descriptions from the context if the user has too many MCP tools installed, and instead provide a "tool search" tool to find available tools. But the agent only knows to search for tools based on the MCP server instructions (which _are_ still in the context). So the server instructions should ideally provide a good high-level summary of what the MCP is capable of, so that agents know to search for relevant tools.

Also, is `errors.go` really the right place for this function? It doesn't have anything to do with errors. Maybe it should just go in `server.go`?

nathanjcochran (approved):
Left a couple minor comments but overall LGTM!

internal/tiger/mcp/db_tools.go
nathanjcochran: Actually, it looks like this is already covered by the update to the MCP server instructions you added in `errors.go`. So I'm not sure this is necessary at all.

aprimakina: Ok, removing 👍

PR #156, opened 2026-05-18
Read only cli commands

aprimakina:
Extend read-only mode enforcement to CLI commands and DB connections
  
Previously, read-only mode only blocked MCP tool calls. This PR extends the gate to:
  - CLI destructive commands — service create/fork/start/stop/resize/update-password/delete now call `common.CheckReadOnly` before any API request
  - DB connections — db connection-string and db connect now OR the `--read-only` flag with `cfg.ReadOnly`, so the global config forces read-only PostgreSQL sessions even without the flag

@ggodeke requested the change, we prefer a more locked-down experience in the tiger CLI compared to ghost

### DIFF
--- a/internal/tiger/cmd/db.go
+++ b/internal/tiger/cmd/db.go
@@ -145,11 +145,16 @@ from your configuration. This command will launch an interactive psql session
 with the appropriate connection parameters.
 
 Authentication is handled automatically using:
-1. Stored password (keyring, ~/.pgpass, or none based on --password-storage setting)  
+1. Stored password (keyring, ~/.pgpass, or none based on --password-storage setting)
 2. PGPASSWORD environment variable
 3. If authentication fails, offers interactive options:
    - Enter password manually (will be saved for future use)
    - Reset password (update or generates a new password via the API)
+
+Use --read-only to open the psql session in Tiger Cloud's immutable read-only
+mode (writes and DDL are rejected by the server). The global read_only config
+option (or TIGER_READ_ONLY=true) also forces this behavior, so sessions started
+while read-only mode is on are always read-only.
 
 Examples:
   # Connect to default service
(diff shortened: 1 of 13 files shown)


internal/tiger/common/errors.go
```diff
@@ -1,6 +1,23 @@
...
+)
+
+// ErrReadOnly is returned when a destructive operation is attempted while
+// read-only mode is enabled in the user's config.
+var ErrReadOnly = errors.New("this operation is not allowed in read-only mode")
+
+// CheckReadOnly returns ErrReadOnly if read-only mode is enabled. Callers
+// should invoke this before any destructive API call.
+func CheckReadOnly(cfg *config.Config) error {
+	if cfg != nil && cfg.ReadOnly {
```
nathanjcochran: In what case would `cfg` be `nil`? It looks like the prior version of this function (which was moved here from the `./internal/tiger/mcp` package) always assumed `cfg` was non-`nil` (which makes sense to me - I believe we usually error-out if we can't load the config). Just want to make sure we don't end up silently ignoring an error that caused the config to not load, and reporting that read-only mode is not enabled (when it actually might be).

internal/tiger/cmd/db.go
```diff
@@ -61,6 +61,9 @@ Use --with-password to include the password directly in the connection string.
 
 Use --read-only to emit a connection string that opens the session in Tiger
 Cloud's immutable read-only mode (writes and DDL are rejected by the server).
+The global read_only config option (or TIGER_READ_ONLY=true) also forces this
+behavior, so connection strings produced while read-only mode is on always
+open read-only sessions.
```
nathanjcochran, line 66: Should we include a similar note in the `tiger db connect` description/help text? That command's help text doesn't include much about read-only mode at all, except for one example of using the `--read-only` flag.

nathanjcochran (approved):
Left two minor comments, but overall LGTM! ✅

internal/tiger/common/errors.go
aprimakina: You're right. Every caller passes `cfg.Config` from a successful `common.LoadConfig`, which already errors out on load failure, so we never reach this with a nil cfg. Dropping the check 👍

PR #158, opened 2026-05-20
feat(auth): support OAuth Bearer login alongside PAT

aprimakina:
Add OAuth (PKCE) authentication to the CLI and consolidate how auth info is
  fetched, while keeping PAT / API-key auth fully supported.

  ## Changes

  - **OAuth login**: `tiger auth login` now defaults to an interactive browser flow. PAT auth still works via `--public-key/--secret-key` flags or `TIGER_PUBLIC_KEY` / `TIGER_SECRET_KEY`.
  - **Bearer client with auto-refresh**: OAuth sessions use a Bearer-authenticated client that transparently refreshes expired access tokens and persists the rotated token.
  - **`auth logout`**: revokes the refresh token server-side for OAuth sessions.
  - Removed the unused GraphQL-based PAT-creation path.


internal/tiger/common/client.go
```diff
@@ -102,13 +114,16 @@ func ValidateAPIKey(ctx context.Context, cfg *config.Config, client *api.ClientW
...
-	// Identify the user with analytics
-	a := analytics.New(cfg, client, authInfo.ApiKey.Project.Id)
+	a := analytics.New(cfg, client, apiKey.Project.Id)
 	a.Identify(
-		analytics.Property("userId", authInfo.ApiKey.IssuingUser.Id),
-		analytics.Property("email", string(authInfo.ApiKey.IssuingUser.Email)),
-		analytics.Property("planType", authInfo.ApiKey.Project.PlanType),
+		analytics.Property("userId", apiKey.IssuingUser.Id),
+		analytics.Property("email", string(apiKey.IssuingUser.Email)),
+		analytics.Property("planType", apiKey.Project.PlanType),
```
Askir, line 114: Do we ever set these in the new oAuth based flow? Otherwise our analytics events will be lacking these details for oauth users, which would be a bit sad and break nice dashboards.

internal/tiger/api/client_util.go
```diff
@@ -37,36 +39,126 @@ func getHTTPClient() *http.Client {
...
+		if v > 0 {
+			token.Expiry = time.Unix(int64(v), 0)
+		}
+	case int64:
+		if v > 0 {
+			token.Expiry = time.Unix(v, 0)
+		}
+	}
+}
+
```
Askir: Any thoughts about changing this (or atleast adding the `expires_in` field) on the backend instead if this is the oauth standard? :D

internal/tiger/api/client_util.go
```diff
@@ -37,36 +39,126 @@ func getHTTPClient() *http.Client {
...
+			AuthStyle: oauth2.AuthStyleInParams,
+		},
+	}
+
+	var src oauth2.TokenSource = oauthCfg.TokenSource(context.Background(), token)
+	if persist != nil {
+		src = &persistingTokenSource{base: src, persist: persist, last: token.AccessToken}
+	}
+
+	httpClient := oauth2.NewClient(context.Background(), src)
```
Askir: This should probably also use the getHttpClient from above, which defines some sane defaults?

internal/tiger/cmd/oauth.go
```diff
@@ -232,26 +218,32 @@ func openBrowserImpl(url string) error {
...
+	client, err := api.NewTigerClientWithToken(l.cfg, token, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create API client: %w", err)
+	}
+	resp, err := client.GetProjectsWithResponse(ctx)
 	if err != nil {
 		return "", fmt.Errorf("failed to get user projects: %w", err)
 	}
+	if resp.JSON200 == nil {
+		return "", fmt.Errorf("failed to get user projects: %s", resp.Status())
```
Askir: There is a helper that surfaces the backends errors directly into the CLI: `common.ExitWithErrorFromStatusCode`
We use that across the rest of the code base, this makes it possible to put a good error message where it actually appears (upstream in a service) and it is forwarded all the way to the CLIs output

internal/tiger/config/credentials.go
```diff
@@ -125,77 +145,77 @@ func storeToFile(credentials string) error {
...
-		return "", "", fmt.Errorf("project ID not found in stored credentials")
+	if len(data) == 0 {
+		return "", ErrNotLoggedIn
 	}
-
-	return creds.APIKey, creds.ProjectID, nil
+	return string(data), nil
 }
 
-// RemoveCredentials removes stored credentials from keyring and file fallback
```
Askir, line 186: I'm a bit confused over these removed comments, not like it is super important but they aren't wrong are they? :D

internal/tiger/common/client.go
```diff
@@ -23,6 +27,16 @@ var (
...
+func defaultGetStoredCredentials() (*config.Credentials, error) {
+	// Honor a PAT-shaped test override when present; production leaves it nil.
+	if GetCredentials != nil {
+		if apiKey, projectID, err := GetCredentials(); err == nil && apiKey != "" {
+			return &config.Credentials{APIKey: apiKey, ProjectID: projectID}, nil
+		}
+	}
+	return config.GetStoredCredentials()
+}
+
```
Askir: It'd be nice if we could clean this up at least in a follow up and change all the tests to override the correct method with maybe even an oauth token.

internal/tiger/cmd/oauth.go
```diff
@@ -232,26 +218,32 @@ func openBrowserImpl(url string) error {
...
+	projects := *resp.JSON200
 
 	switch len(projects) {
 	case 0:
 		return "", fmt.Errorf("user has no accessible projects")
 	case 1:
-		return projects[0].ID, nil
+		return projects[0].Id, nil
 	default:
 		return selectProjectInteractively(projects, l.out)
```
Askir, line 244: This is the only way we allow setting a new projectId right? In theory with oauth you could change projects without having to go through the login command again. Not sure if it's worth allowing this via an extra command, a cli ENV var or a new flag on each command e.g. `--projectId`? Fine for now but maybe a convenience thing especially as we think about multi project orgs. @nathanjcochran any opinion here?

Askir (approved):
I think it looks good generally but maybe fix the analytics values before we merge :D

internal/tiger/config/credentials.go
aprimakina: The comments are not wrong; I'll put them back :)

internal/tiger/common/client.go
aprimakina: Great catch, we didn't. Identify was only being called from `ValidateAPIKey`, which is the PAT-only path. Should be fixed with the next commit, except for `plan_type`. Do you think we need it? I haven't noticed any plan-segmented dashboards under https://us.posthog.com/project/104050/dashboard/636540

internal/tiger/common/client.go
aprimakina: https://github.com/timescale/tiger-cli/pull/158/changes/030d9f0ef0e62eddb5444934c65e5c2437fc717c

internal/tiger/common/client.go
aprimakina: https://github.com/timescale/tiger-cli/pull/158/changes/030d9f0ef0e62eddb5444934c65e5c2437fc717c

internal/tiger/api/client_util.go
aprimakina: https://github.com/timescale/savannah-gateway/pull/1749

aprimakina: https://github.com/timescale/tiger-cli/pull/158/changes/f7ae542b719fe9e587cd358fe904c3f92028ad74

internal/tiger/cmd/oauth.go
nathanjcochran: Yeah, imo, we should probably have a command for switching projects (e.g. `tiger project switch`?), which shows this same interactive menu by default (or takes a project ID as an argument), and stores the user's choice in the keyring, like it does now when you log in. That way, users could switch projects without logging in again (assuming they authenticated with OAuth - the command should probably show an error/warning if you try to run it with API key auth). Could also support other `tiger project` sub-commands in the future, like `tiger project list`.

I think we could also bring back the global `--project-id` flag (and a corresponding `TIGER_PROJECT_ID` env var), which could be used to override the saved project ID on a per-command basis without changing what's saved in the keyring. Note that the `--project-id` flag and `TIGER_PROJECT_ID` env var were part of the original CLI design, but were removed in [this PR](https://github.com/timescale/tiger-cli/pull/74) because we couldn't support switching projects while using API keys/client credentials for auth (which are inherently project-scoped).

PR #163, opened 2026-06-08
feat(upgrade): add `tiger upgrade` self-update command

aprimakina:
## What

Adds a `tiger upgrade` self-update command and reworks the update notifier to check on every invocation, reconciled so the notifier points users at the path that works for their install method.

- `tiger upgrade`: replaces the running binary with the latest release: check `latest.txt` → download the platform archive → verify SHA-256 → extract → atomic in-place swap.

- Breaking config change: `version_check_interval` → `version_check` (bool, default `true`). A load-time migration preserves intent for existing configs (`interval == 0` stays disabled); explicit `version_check` / `TIGER_VERSION_CHECK` wins. `config set version_check_interval` now errors.

### DIFF
--- a/internal/tiger/cmd/root.go
+++ b/internal/tiger/cmd/root.go
@@ -16,6 +16,7 @@ import (
 	"github.com/timescale/tiger-cli/internal/tiger/common"
 	"github.com/timescale/tiger-cli/internal/tiger/config"
 	"github.com/timescale/tiger-cli/internal/tiger/logging"
+	"github.com/timescale/tiger-cli/internal/tiger/util"
 	"github.com/timescale/tiger-cli/internal/tiger/version"
 )
 
@@ -27,6 +28,11 @@ func buildRootCmd(ctx context.Context) (*cobra.Command, error) {
 	var passwordStorage string
 	var skipUpdateCheck bool
 	var colorFlag bool
+
+	// versionCheckCh receives the result of the background update check started
+	// in PersistentPreRunE and drained in PersistentPostRunE. nil when no check
+	// was launched (disabled, non-interactive, CI, or --skip-update-check).
+	var versionCheckCh chan *version.CheckResult
 
 	cmd := &cobra.Command{
 		Use:   "tiger",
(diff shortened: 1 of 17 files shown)


internal/tiger/cmd/root.go
```diff
@@ -80,6 +86,27 @@ tiger auth login
...
 			}
 
+			// Kick off a background check for a newer release so the network
+			// fetch overlaps with the command's actual work; the result is
+			// printed in PersistentPostRunE. Gated to interactive, non-CI
+			// terminals. `version --check` runs its own synchronous check.
+			isVersionCheckCmd := cmd.Name() == "version" && cmd.Flag("check") != nil && cmd.Flag("check").Changed
+			if cfg.VersionCheck && !skipUpdateCheck && !isVersionCheckCmd &&
+				!util.IsCI() && util.IsTerminal(cmd.ErrOrStderr()) {
+				versionCheckCh = make(chan *version.CheckResult, 1)
```
Askir, line 98: I think these checks also pass during the `tiger upgrade` command which is probably a bit counter intuitive since it'll post that there is a new version available when upgrading? Or is there a gate somewhere that I am missing?

internal/tiger/cmd/upgrade.go
```diff
@@ -0,0 +1,547 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runUpgrade(cmd, requestedVersion, force)
+		},
+	}
+
+	cmd.Flags().StringVar(&requestedVersion, "version", "", "specific version to install (e.g. v1.2.3). Defaults to latest.")
+	if err := cmd.Flags().MarkHidden("version"); err != nil {
+		panic(err)
+	}
+	cmd.Flags().BoolVar(&force, "force", false, "reinstall even if the current version already matches, or the binary was installed via a package manager")
```
Askir, line 71: Is the "replace despite package manager used" truly intended? Do the package managers not trip over it if you change the binaries out underneath them? 😅 I guess its very niche probably a non-issue.

internal/tiger/cmd/upgrade.go
```diff
@@ -0,0 +1,547 @@
...
+	}
+	if err := stagedFile.Chmod(0o755); err != nil {
+		return err
+	}
+	if err := stagedFile.Close(); err != nil {
+		return fmt.Errorf("failed to close staged binary %s: %w", stagedPath, err)
+	}
+
+	if runtime.GOOS == "windows" {
+		cleanupStaleOldBinaries(currentBinaryPath)
```
Askir, line 506: This is only called here right? So we always keep the old one until the next upgrade? I guess that's not necessarily bad as a backup anyway. Just good to know.

Askir (approved):
Looks good to me!! Since you build a `--version` flag and also allow `--force` we could have an integration test I think, where you'd actually run this agains the live CDN in github actions? Not sure if worth it, but I guess breaking the upgrade command would be quite ugly since it impacts everyone so maybe it is.

internal/tiger/cmd/upgrade.go
aprimakina: Right, and it's forced rather than chosen: this branch is Windows-only, and Windows locks a running executable, so the process can't delete its own .exe, only rename it. The earliest opportunity to remove the .old.<pid> file is a future invocation, which is what this call does. On Linux/macOS there's no leftover at all; the rename overwrites in place.

internal/tiger/cmd/root.go
aprimakina: Fixed ✅

internal/tiger/cmd/upgrade.go
aprimakina: By default, it is refused, with the right package-manager command shown instead. `--force` is a hidden escape hatch (it warns before proceeding), mainly for when the tap/repo lags behind a release or the install-method detection guesses wrong.

And the package managers don't trip over it: the next brew upgrade / apt install just overwrites our binary with its own copy, so the drift self-heals.

The whole mechanism is inspired by ghost's upgrade command, which gates package-manager installs the same way.

PR #164, opened 2026-06-12
feat(config): enable read-only mode by default for new installations

aprimakina:
Flip the read_only default to true so fresh installs refuse mutating operations (CLI service commands, MCP write tools) and open database sessions read-only until the user opts in with `tiger config set read_only false`.

Existing installations are grandfathered: older CLI versions only wrote read_only to the config file when set explicitly, so a config file without the key predates the default flip and `MigrateReadOnly` (same in-memory-shim pattern as `MigrateVersionCheck`) resolves it to false. To keep that detection sound, every config file written by Set/Unset/Reset now records read_only explicitly; unsetting or resetting restores the current default (true).


aprimakina:
We decided to put the idea on hold to avoid disrupting the onboarding flow.

PR #165, opened 2026-06-16
feat(db): offer read replica connection on `tiger db connect`

aprimakina:
When tiger db connect / psql runs in an interactive terminal and the service has active read replicas, prompt the user to connect to one instead of the primary. With no replicas, non-TTY stdin, or --no-replica-prompt, it connects to the primary directly as before.

  Notes
  - Replicas share the primary's credentials (password storage/recovery still operates on the primary).
  - --pooled is strict for the primary (hard error if no pooler) and best-effort for replicas (warns, falls back to direct).
  - Refactors `GetConnectionDetails`/`GetReplicaConnectionDetails` onto a shared `buildConnectionDetails` helper.


internal/tiger/cmd/read_replica.go
```diff
@@ -0,0 +1,233 @@
...
+	case targetCancel:
+		return nil, nil
+	case targetPrimary:
+		return returnPrimary()
+	}
+
+	replica := choice.replica
+
+	// --pooled is best-effort on replicas: warn and connect directly if the
+	// chosen replica has no pooler.
```
Askir, line 84: Is there a particular reason for why we do it this way for replicas? Sorry not super familiar with connection poolers in general.

Askir (approved):
Looks good over all just a bit confused over the pooler behaviour.

internal/tiger/cmd/read_replica.go
aprimakina: For the primary, --pooled is the explicit request, so we fail hard if it's unavailable (unchanged behavior). For a replica, it's just a modifier on the connect action you already chose interactively, so rather than blocking the connection, we warn and fall back to a direct connection.

PR #166, opened 2026-06-18
feat(db): add schema introspection via `tiger db schema` and `db_sche…

aprimakina:
Introduces schema introspection for a service's database, exposed both as a CLI command (`tiger db schema`) and an MCP tool (`db_schema`). It returns tables (regular, partitioned, and foreign), views, materialized views, enum types, functions, procedures, indexes, triggers, and TimescaleDB hypertable / continuous aggregate metadata, rendered as readable plain text for both humans and an AI agent's context window.

Only objects the connecting role can access are returned, and the connection is always opened in immutable read-only mode — so the MCP tool is marked `ReadOnlyHint: true` and is safe to expose even when `db_execute_query` is gated.

  **What's included**

  - tiger db schema [service-id] — new CLI command with flags:
    - --schema — restrict output to a single schema
    - --internal — include system schemas (pg_*, information_schema, TimescaleDB internals) and extension-owned objects
    - --definitions — include view SELECTs and function/procedure bodies
    - --comments — include COMMENT ON text
  - db_schema MCP tool — same engine and options, with a rich input schema (descriptions, examples, defaults) and `ReadOnlyHint: true`.

  **Implementation**

  - The introspection engine (schema.go / schema_format.go) is ported from the ghost CLI, with a new pure entry point `FetchSchemaFromConn`(conn, ident, opts) that takes a live pgx connection.
  - Tiger-specific glue in common/:
    - `CheckServiceReady` — readiness guard that allows-lists READY against DeployStatus
    - `ConnectToService` — shared service → *pgx.Conn helper (db_execute_query was refactored onto it, removing duplicated connection setup)
    - `FetchServiceSchema` — readiness check → read-only connect → introspect; shared by the CLI command and the MCP tool
  - By default only user-facing schemas/objects are shown; definitions and comments are omitted unless requested, since they can be large and may embed implementation details.


aprimakina:
Motivation to have the new command:
  - Correctness. Schema introspection is tricky (index sort order, trigger bitmasks, cross-schema FKs, hidden leaf partitions, TimescaleDB metadata), so encoding the correct queries once returns a complete, accurate picture every time instead of an agent getting details wrong on the fly.
  - Access control. `db_schema` shows only objects the connecting role can access — filtering out catalog internals and platform noise — and is read-only by design, so it's safe to expose even when `db_execute_query` is locked down.
  - Token cost and round-trips. Improvising means many back-and-forth calls that burn context and time, while `db_schema` does it in one call, returning compact, pre-formatted plain text (not JSON, which packs less efficiently into a context window).
  - Determinism for the model. A named tool with a fixed description is far easier to invoke reliably than hoping the model recalls the right catalog query each session — the difference between a clear API and a blank SQL prompt.
  - The honest trade-off. A dedicated tool is rigid, but if a user needs something, it doesn't expose the agent falls back to `db_execute_query`, so the two compose — curated default plus escape hatch.

internal/tiger/mcp/db_tools.go
```diff
@@ -127,6 +127,119 @@ WARNING: Can execute any SQL statement including INSERT, UPDATE, DELETE, and DDL
...
+		zap.String("role", input.Role),
+		zap.Bool("pooled", input.Pooled),
+	)
+
+	serviceResp, err := cfg.Client.GetServiceWithResponse(ctx, cfg.ProjectID, input.ServiceID)
+	if err != nil {
+		return nil, DBSchemaOutput{}, fmt.Errorf("failed to get service details: %w", err)
+	}
+	if serviceResp.StatusCode() != http.StatusOK {
+		return nil, DBSchemaOutput{}, serviceResp.JSON4XX
```
Askir: This only returns a useful error if its a 4xx status code if its a 5xx it would be nil I think. Not sure if worth improving.

Askir (approved):
Only checked the plumbing since the introspection logic comes from ghost, but looks good to me 👍

internal/tiger/mcp/db_tools.go
aprimakina: Fixed

PR #167, opened 2026-06-19
feat(mcp): cap db_execute_query result size to protect agent context

aprimakina:
LLMs using `db_execute_query` tend to SELECT raw rows and process locally, blowing up the context window and running slowly. Cap the data the tool returns and steer agents toward doing the work in SQL.

  - Row cap per result set via the new `mcp_max_rows` config (default 100). Not a per-call parameter — the config is the single knob, so the model can't override it. `LIMIT` in the query remains the preferred tool when the agent knows how many rows it wants.
  - 256 KiB safety net across the response, guarding against a few very wide rows that the row cap alone would miss. Always keep at least one row, so an oversized first row never yields an empty result.
  - On truncation, the query is not cancelled — we stop collecting but drain the result set, so writes and later statements still run and `rows_affected` reports the true total. The response sets `truncated: true` and returns a `notice` steering the agent to aggregate / filter / paginate in SQL instead of re-running the query.
  - Server instructions + tool description push aggregation, filtering, sorting, and joins into SQL instead of fetching raw rows.

MCP-only; CLI query paths are unchanged.

### DIFF
--- a/internal/tiger/mcp/db_tools.go
+++ b/internal/tiger/mcp/db_tools.go
@@ -19,37 +19,22 @@ import (
 )
 
 const (
-	// mcpMaxRowsCeiling is the hard upper bound on rows returned per result set,
-	// regardless of the configured or per-call max_rows.
-	mcpMaxRowsCeiling = 10000
-
-	// mcpMaxResponseBytes caps the total serialized row data across a response,
-	// guarding the model's context against a few very wide rows that the row cap
-	// alone would miss. Not user-configurable.
+	// mcpMaxResponseBytes caps total serialized row data per response, catching a
+	// few very wide rows the row cap alone would miss. Not user-configurable.
 	mcpMaxResponseBytes = 256 * 1024
 )
 
-// resolveMaxRows returns the effective per-result-set row cap: the per-call
-// value if > 0, else the configured value, else the default, clamped to
-// mcpMaxRowsCeiling. It also sanitizes non-positive config-file/env values
-// (which bypass `tiger config set` validation) to the default.
-func resolveMaxRows(configured, requested int) int {
-	n := configured
-	if requested > 0 {
-		n = requested
+// resolveMaxRows returns the row cap from mcp_max_rows, falling back to the
+// default for non-positive config-file/env values (which skip set validation).
+func resolveMaxRows(configured int) int {
+	if configured <= 0 {
+		return config.DefaultMCPMaxRows
 	}
-	if n <= 0 {
-		n = config.DefaultMCPMaxRows
-	}
-	if n > mcpMaxRowsCeiling {
-		n = mcpMaxRowsCeiling
-	}
-	return n
+	return configured
 }
 
-// approxRowSize estimates the serialized size, in bytes, of a single row's
-// values. It is used to enforce the overall response byte budget. The estimate
-// mirrors how the row is ultimately serialized to JSON for the client.
+// approxRowSize estimates a row's serialized size in bytes for the byte budget,
+// mirroring how it is ultimately marshaled to JSON for the client.
 func approxRowSize(values []any) int {
 	if b, err := json.Marshal(values); err == nil {
 		return len(b)
@@ -61,7 +46,7 @@ func approxRowSize(values []any) int {
 // truncationNotice builds the actionable guidance returned to the model when a
 // response is truncated.
 func truncationNotice(maxRows int) string {
-	return fmt.Sprintf("Results were truncated to limit the amount of data returned (max_rows=%d per result set, plus an overall response size cap). More rows exist. Do the work in the database instead of re-running this query: aggregate (GROUP BY, COUNT, SUM, AVG), filter (WHERE), or paginate (LIMIT/OFFSET). To pull more raw rows in a single call, set max_rows (up to %d).", maxRows, mcpMaxRowsCeiling)
+	return fmt.Sprintf("Results were truncated to limit the amount of data returned (the configured mcp_max_rows=%d per result set, plus an overall response size cap). More rows exist. Do the work in the database instead of re-running this query: aggregate (GROUP BY, COUNT, SUM, AVG), filter (WHERE), or paginate (LIMIT/OFFSET).", maxRows)
 }
 
 // DBExecuteQueryInput represents input for db_execute_query
@@ -72,7 +57,6 @@ type DBExecuteQueryInput struct {
 	TimeoutSeconds int      `json:"timeout_seconds,omitempty"`
 	Role           string   `json:"role,omitempty"`
 	Pooled         bool     `json:"pooled,omitempty"`
-	MaxRows        int      `json:"max_rows,omitempty"`
 }
 
 func (DBExecuteQueryInput) Schema() *jsonschema.Schema {
@@ -99,12 +83,6 @@ func (DBExecuteQueryInput) Schema() *jsonschema.Schema {
 	schema.Properties["pooled"].Description = "Use connection pooling (if available)"
 	schema.Properties["pooled"].Default = util.Must(json.Marshal(false))
 	schema.Properties["pooled"].Examples = []any{false, true}
-
-	// No schema Default: it would be injected by the SDK and shadow the
-	// configured mcp_max_rows fallback when max_rows is omitted.
-	schema.Properties["max_rows"].Description = fmt.Sprintf("Maximum number of rows to return per result set. When the query produces more, the result set is truncated (the response indicates this) and the in-flight query is aborted to avoid streaming and buffering data the model won't use. Defaults to the configured mcp_max_rows (%d). Hard-capped at %d. Prefer aggregating or filtering in SQL over raising this.", config.DefaultMCPMaxRows, mcpMaxRowsCeiling)
-	schema.Properties["max_rows"].Minimum = util.Ptr(1.0)
-	schema.Properties["max_rows"].Examples = []any{100, 500}
 
 	return schema
 }
@@ -153,10 +131,10 @@ func (DBExecuteQueryOutput) Schema() *jsonschema.Schema {
 	resultSetSchema.Properties["rows"].Description = "Result rows as arrays of values. Omitted for commands that don't return rows (INSERT, UPDATE, DELETE, etc.)"
 	resultSetSchema.Properties["rows"].Examples = []any{[][]any{{1, "alice", "2024-01-01"}, {2, "bob", "2024-01-02"}}}
 
-	resultSetSchema.Properties["rows_affected"].Description = "Number of rows affected. For SELECT, this is the number of rows returned (which equals the number of rows in this response; when truncated is true, more rows existed but were not returned). For INSERT/UPDATE/DELETE, this is the number of rows modified. Returns 0 for statements that don't return or modify rows (e.g. CREATE TABLE)."
+	resultSetSchema.Properties["rows_affected"].Description = "Number of rows affected. For SELECT, this is the total number of rows the query produced; when truncated is true this exceeds the number of rows actually returned in this response. For INSERT/UPDATE/DELETE, this is the number of rows modified. Returns 0 for statements that don't return or modify rows (e.g. CREATE TABLE)."
 	resultSetSchema.Properties["rows_affected"].Examples = []any{5, 42, 1000}
 
-	resultSetSchema.Properties["truncated"].Description = "True when this result set was capped (by max_rows or the overall response size limit) and additional rows exist that were not returned. Refine the query in SQL to get the data you need."
+	resultSetSchema.Properties["truncated"].Description = "True when this result set was capped (by the configured mcp_max_rows row limit or the overall response size limit) and additional rows exist that were not returned. Refine the query in SQL to get the data you need."
 
 	schema.Properties["execution_time"].Description = "Execution time as a human-readable duration string"
 	schema.Properties["execution_time"].Examples = []any{"123ms", "1.5s", "45.2µs"}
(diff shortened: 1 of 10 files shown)


internal/tiger/mcp/db_tools.go
```diff
@@ -297,25 +384,53 @@ func processResultSet(conn *pgx.Conn, rows pgx.Rows) (ResultSet, error) {
...
+
 		resultRows = append(resultRows, values)
 	}
 
-	// Check for errors during iteration
-	if err := rows.Err(); err != nil {
+	if truncated {
+		// Cancel before the deferred rows.Close() so pgx aborts the query
+		// instead of draining the remaining rows.
+		cancel()
```
Askir: Does this make sense? If the LLM provide a multi statement string, this will cancel everything and roll it back according to claude. That means If there is two selects the second one might never fire and if it mixes updates and deletes in those will be rolled back.

Should we maybe just do the truncation client side? So first processing everything and load it into memory but then truncate the individual result sets after execution?

Askir (approved):
Not 100% sure about the cancellation logic it seems a bit weird that we would roll everything back just because you hit the row limit.

Also Claude told me this :D
> mcp_max_rows is missing from `tiger config show` table output

internal/tiger/mcp/db_tools.go
```diff
@@ -52,6 +100,12 @@ func (DBExecuteQueryInput) Schema() *jsonschema.Schema {
 	schema.Properties["pooled"].Default = util.Must(json.Marshal(false))
 	schema.Properties["pooled"].Examples = []any{false, true}
 
+	// No schema Default: it would be injected by the SDK and shadow the
+	// configured mcp_max_rows fallback when max_rows is omitted.
+	schema.Properties["max_rows"].Description = fmt.Sprintf("Maximum number of rows to return per result set. When the query produces more, the result set is truncated (the response indicates this) and the in-flight query is aborted to avoid streaming and buffering data the model won't use. Defaults to the configured mcp_max_rows (%d). Hard-capped at %d. Prefer aggregating or filtering in SQL over raising this.", config.DefaultMCPMaxRows, mcpMaxRowsCeiling)
```
nathanjcochran: This is a lot of text for a single field description, and it feels somewhat duplicative with the notice that gets returned when results are truncated. I'd try to make this a little more concise, if possible. Each field description uses up context, and we have lots of tools/fields, so we need to be careful or our MCP will end up eating a ton of context.

internal/tiger/mcp/db_tools.go
```diff
@@ -116,6 +179,10 @@ Connects to a PostgreSQL database service in Tiger Cloud and executes the provid
...
 
+DO THE WORK IN THE DATABASE. PostgreSQL is far more efficient at processing data than fetching raw rows and computing in your context. Push computation into SQL: aggregate with GROUP BY / COUNT / SUM / AVG / MIN / MAX, filter with WHERE, sort and take the top N with ORDER BY ... LIMIT, and join/transform server-side. Avoid SELECT * on large tables; project only the columns you need.
+
+RESULTS ARE CAPPED. Each result set returns at most max_rows rows (default 100, configurable via mcp_max_rows), and the total response size is bounded. If a result set is truncated, the response sets "truncated": true and includes a "notice" — refine the query in SQL (aggregate, filter, or paginate with LIMIT/OFFSET) rather than re-running it to pull everything. Only raise max_rows when you genuinely need more raw rows in a single call.
```
nathanjcochran: Again, all of this content feels very duplicative to me. I don't think we need to beat a dead horse. We should aim to explain this all just once, and in as few words as possible, imo.

internal/tiger/mcp/db_tools.go
nathanjcochran: I agree with Jascha. I don't think we should cancel the query here, just drain the rows. I think calling [rows.Close()](https://pkg.go.dev/github.com/jackc/pgx/v5#Rows) is probably sufficient for this result set, and then also [br.Close()](https://pkg.go.dev/github.com/jackc/pgx/v5#BatchResults) to close the entire batch (i.e. additional result sets).

internal/tiger/mcp/db_tools.go
```diff
@@ -297,25 +384,53 @@ func processResultSet(conn *pgx.Conn, rows pgx.Rows) (ResultSet, error) {
...
+		cancel()
+	} else if err := rows.Err(); err != nil {
 		return ResultSet{}, err
 	}
 
 	commandTag := rows.CommandTag()
+	rowsAffected := commandTag.RowsAffected()
+	if truncated {
+		// The command tag is unreliable after an abort; report rows returned.
+		rowsAffected = int64(len(resultRows))
```
nathanjcochran: I think another big benefit of not cancelling the query is that we could return the _true_ number of rows that the query returned here (as returned in the command tag), even when the results are truncated. I think that's important info, personally. It could help the LLM decide how to proceed.

internal/tiger/mcp/db_tools.go
```diff
@@ -13,10 +13,57 @@ import (
...
 	"github.com/timescale/tiger-cli/internal/tiger/common"
+	"github.com/timescale/tiger-cli/internal/tiger/config"
 	"github.com/timescale/tiger-cli/internal/tiger/logging"
 	"github.com/timescale/tiger-cli/internal/tiger/util"
 )
 
+const (
+	// mcpMaxRowsCeiling is the hard upper bound on rows returned per result set,
+	// regardless of the configured or per-call max_rows.
+	mcpMaxRowsCeiling = 10000
```
nathanjcochran: I'm personally not in favor of including a hard upper limit on the number of rows that can be returned. I think tools should give good defaults, but allow users to override them and do what they want at the end of the day. If a user wants to return more than 10,000 rows, they should be allowed to configure `mcp_max_rows` with a higher limit (or even disable the limit entirely, imo). I don't think we should prevent users from making their own decisions with respect to how they fetch and process their own data 🤷‍♂️.

internal/tiger/mcp/db_tools.go
```diff
@@ -25,6 +72,7 @@ type DBExecuteQueryInput struct {
 	TimeoutSeconds int      `json:"timeout_seconds,omitempty"`
 	Role           string   `json:"role,omitempty"`
 	Pooled         bool     `json:"pooled,omitempty"`
+	MaxRows        int      `json:"max_rows,omitempty"`
```
nathanjcochran: I'm skeptical that we really need this parameter at all. LLMs can already limit the number of rows returned from a query by adding a `LIMIT` (and I think they're typically pretty good at doing this). Do we really need to expose a different knob to control basically the same thing (but with different semantics)? Won't this just confuse LLMs? Ultimately, a `LIMIT` is always the better choice, since it allows the query planner to optimize the query differently and avoiding sending rows over the wire in the first place. By exposing this parameter, it feels like we've nudging the LLM to use the parameter instead of a `LIMIT`, even though it's not going to be as efficient.

internal/tiger/mcp/db_tools.go
```diff
@@ -13,10 +13,57 @@ import (
...
+
+// resolveMaxRows returns the effective per-result-set row cap: the per-call
+// value if > 0, else the configured value, else the default, clamped to
+// mcpMaxRowsCeiling. It also sanitizes non-positive config-file/env values
+// (which bypass `tiger config set` validation) to the default.
+func resolveMaxRows(configured, requested int) int {
+	n := configured
+	if requested > 0 {
+		n = requested
+	}
```
nathanjcochran: If I'm reading this correctly, it looks like the LLM can override the configured limit (i.e. the `mcp_max_rows` config setting) by sending it's own limit via the `max_rows` tool parameter. Is that really what we want? What's the point of having a global config setting if the LLM can just disregard it?

See my other comment about whether we really need the `max_rows` parameter at all (I'm skeptical).

nathanjcochran (commented):
I'm a bit skeptical of the need for this feature. For one thing, LLMs can already limit the number of results returned by including a `LIMIT` in the query itself. But also, many agent harnesses (e.g. Claude Code) will automatically send tool call results to a file if they're too large (instead of putting them directly into the context), and then the agent can use tools like `grep` or `jq` to search through the results in the file. So I'm not really convinced we need to add yet _another_ way to limit/control the size of tool call results.

I guess I'm not wholly opposed to a configurable user limit (i.e. the `mcp_max_rows` config setting), but I do think the `max_rows` parameter and the hard upper cap are probably unnecessarily complicating things and limiting what users can do. See my comments below about that. Just my two cents.

internal/tiger/mcp/db_tools.go
aprimakina: Fixed.

internal/tiger/mcp/db_tools.go
aprimakina: Removed the hard upper limit ✅

internal/tiger/mcp/db_tools.go
aprimakina: Nice idea 👍  Since we no longer cancel, the command tag is valid, so `rows_affected` now reports the true total even when truncated.

internal/tiger/mcp/db_tools.go
aprimakina: Trimmed the text 👍

internal/tiger/mcp/db_tools.go
aprimakina: Resolved by dropping the param, `mcp_max_rows` config is now authoritative, and the model can't override it.

aprimakina:
> I'm a bit skeptical of the need for this feature. For one thing, LLMs can already limit the number of results returned by including a LIMIT in the query itself. But also, many agent harnesses (e.g. Claude Code) will automatically send tool call results to a file if they're too large (instead of putting them directly into the context), and then the agent can use tools like grep or jq to search through the results in the file. So I'm not really convinced we need to add yet another way to limit/control the size of tool call results.

> I guess I'm not wholly opposed to a configurable user limit (i.e. the mcp_max_rows config setting), but I do think the max_rows parameter and the hard upper cap are probably unnecessarily complicating things and limiting what users can do. See my comments below about that. Just my two cents.

Fair points — dropped both the `max_rows` param and the hard ceiling, so `mcp_max_rows` config is now the only knob. On keeping a default cap: LIMIT covers the case where the model knows the count, but not every query includes one, and the file-spill behavior is Claude Code-specific — other MCP clients may put the full result straight into context. The byte cap also handles a case LIMIT can't handle: wide rows (e.g., large jsonb/text) that blow up context even at low row counts. So it stays as a fixed safety net rather than another knob.

internal/tiger/mcp/db_tools.go
```diff
@@ -116,6 +157,8 @@ Connects to a PostgreSQL database service in Tiger Cloud and executes the provid
 
 Multi-statement queries (semicolon-separated) are supported when no parameters are provided. All result sets are returned. By default, statements execute in an implicit transaction that automatically commits on success or rolls back on error. Explicit transactions (opened with BEGIN) must be explicitly committed with COMMIT, or they roll back when the connection closes.
 
+Process data in the database, not in your context: aggregate, filter, sort/limit, and join in SQL rather than fetching raw rows. Results are capped per result set (default 100 rows, configurable via mcp_max_rows) and by total size; a truncated response sets "truncated": true with a "notice" on how to refine the query.
```
nathanjcochran: I still think this is somewhat duplicative with the notice text, `notice` field description, and `truncated` field description.

internal/tiger/mcp/db_tools.go
```diff
@@ -392,25 +358,49 @@ func processResultSet(conn *pgx.Conn, rows pgx.Rows) (ResultSet, error) {
...
 	}
 
-	// Check for errors during iteration
-	if err := rows.Err(); err != nil {
+	if truncated {
+		// Drain so the command tag (and true row count) is available.
+		rows.Close()
+	} else if err := rows.Err(); err != nil {
 		return ResultSet{}, err
 	}
```
nathanjcochran: I _think_ it's probably safe to call `rows.Close` unconditionally, fwiw (not just when `truncated == true`). I think it's normally closed automatically when the `rows.Next()` loop ends, but the docs say:

> It is safe call Close after rows is already closed.

So calling it unconditionally is probably fine too, and maybe a little less code. I'd also recommend checking `rows.Err()` unconditionally, since a query could fail even after we've scanned a bunch of rows, and we'd only know if we check `rows.Err()`.

internal/tiger/mcp/server.go
```diff
@@ -37,7 +37,9 @@ type Server struct {
 // gate itself stays correct because handlers reload config per call.
 func buildServerInstructions(cfg *config.Config) string {
 	base := "Tiger MCP provides tools for creating, managing, and querying Tiger Cloud database services (managed TimescaleDB/PostgreSQL). " +
-		"Use it to provision and fork services, start/stop/resize instances, rotate credentials, fetch service logs, execute SQL queries, and search Tiger documentation."
+		"Use it to provision and fork services, start/stop/resize instances, rotate credentials, fetch service logs, execute SQL queries, and search Tiger documentation. " +
+		"When working with data, do the work in the database: push aggregation, filtering, sorting, and joins into SQL rather than fetching raw rows and computing locally. " +
+		"db_execute_query caps the rows and total size it returns; if a result is truncated, refine the query (aggregate, filter, or paginate with LIMIT/OFFSET) instead of re-running it to pull everything."
```
nathanjcochran: I don't think this needs to be in the top-level server instructions at all. I'd keep this in the `db_execute_query` tool description, since it's really just about that one tool.

nathanjcochran (approved):
Left a few minor comments but overall LGTM!

PR #174, opened 2026-07-02
feat: send CLI User-Agent on the OAuth token exchange

aprimakina:
The login token exchange used oauth2's default HTTP client, so it sent no `tiger-cli` User-Agent — the gateway recorded the CLI session's `device_name` as the Go default. This routes the exchange through a client that sets the CLI User-Agent (enriched with OS/arch, `tiger-cli/<version> (os/arch)`) via a shared `config.UserAgent()` helper reused by the API client.

Completes the CLI side of session device-info recording:
- **tiger-cli** (this PR) — sends the User-Agent on the token exchange.
- timescale/savannah-gateway#1837 — forwards `x-forwarded-user-agent` / `x-forwarded-for` to the User service.
- timescale/savannah-users#560 — records `device_name` / `device_ip` on the session.


Askir (approved):
looks simple enough 👍

PR #175, opened 2026-07-02
Add metrics commands as experimental

RobAtticus:
These commands can be used to get the full series of metric data points for a service. We introduce a TIGER_EXPERIMENTAL env var to enable this since the REST endpoint is currently marked as preview. After some testing, we can graduate this.


internal/tiger/util/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	To      time.Time         `json:"to"`
+	Count   int               `json:"count"`
+	Min     float64           `json:"min"`
+	MinTime time.Time         `json:"min_time"`
+	Max     float64           `json:"max"`
+	MaxTime time.Time         `json:"max_time"`
+	Avg     float64           `json:"avg"`
+	P50     float64           `json:"p50"`
+	P95     float64           `json:"p95"`
+}
```
AdrianLC: I would just return the data points and let the agent do its own calculations. 

We already have the AggFn parameter and everything, it's just not exposed in the gateway yet so if we want aggregations I would expose that instead of doing it here.

Doing average and percentiles over last value aggregations (what the API does currently) is not accurate because there is already a big lose of data points since we are only taking last_value from the samples.

AdrianLC: After discussing with @RobAtticus we agreed to remove this. I pushed a commit doing that

AdrianLC (approved):
LGTM

Aside from removing the mode=summary I made some minor adjustments to match recent changes in the API.

For context on why we are removing `mode=summary` the metric data points at the DB layer are currently aggregated with LAST_VALUE within the bucket. Doing Avg or Percentiles over Last values is wrong and doest not result in accurate values.
The right way to solve this is to another API endpoint that can summarize or aggregate correctly.

RobAtticus:
Claude flagged a few minor things that I pushed in a latest commit. It had one more one bigger refactor to remove some of the code duplication, but we can see how this evolves a bit first before doing that.

I'll leave this for you to merge in your AM since I'm about to sign off for the day

AdrianLC (approved):

PR #177, opened 2026-07-15
feat(db): accept a read replica ID in DB operation commands

aprimakina:
The tiger db connection commands (connection-string, connect/psql, test-connection, schema) and the MCP db_execute_query / db_schema tools now accept a read replica set ID wherever a service ID is accepted. A replica ID connects to that replica's endpoint using the parent primary's credentials (replicas share them); a primary ID behaves as before.

The API resolves a replica ID via the normal service-get endpoint, returning the replica's endpoint plus a forked_from link to its parent — so credentials (and password reset) resolve against the parent.


internal/tiger/cmd/db.go
```diff
@@ -92,29 +93,25 @@ Examples:
 				return err
 			}
 
-			service, err := getServiceDetailsFunc(cmd, cfg, args)
+			target, err := resolveConnectionTarget(cmd, cfg, args)
```
Askir: We have `resolveConnectionTaget` and `resolveConnectTarget` now doing wildly different things. But reading as if someone typo'd.
Maybe we could clarify these two? One is the connection chooser and returns ConnectionDetails as the result, the other one is the lookup of the ID

internal/tiger/common/replica.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ConnectionTarget describes where to connect and whose credentials to use.
+// Connect supplies the endpoint/pooler, Credential the password — the same
+// service for a primary; for a read replica, Connect is the replica and
+// Credential is the parent primary whose credentials it shares.
+type ConnectionTarget struct {
+	Connect    api.Service
+	Credential api.Service
+	IsReplica  bool
+}
```
Askir, line 19: I find this poorly named, it took me a long time while reading to understand that Credential is only the CredentialService/ the service where we resolve the credentials from rather than the actual credentials.

I think it would make sense to at least call the two `ConnectionService` and `CredentialService`. Sorry for being a bit pedantic about naming, here. I try not to do that usually but I really struggled understanding what's happening here 😅

Askir (approved):
Naming confused me a lot here but I think otherwise its good.

What happens if you use `update-password`, `save-password` and `tiger db create role` with a read replica? For the record I think its fine if update-password and create role fail (its a read replica after all so read only) but it should probably have a good error. For save-password I genuinely am not sure if it should override the parents password or error out.

internal/tiger/common/replica.go
aprimakina: All good, totally makes sense https://github.com/timescale/tiger-cli/pull/177/changes/c32b197a5858f3eca1705ab8eb192a29959c439e

internal/tiger/cmd/db.go
aprimakina: https://github.com/timescale/tiger-cli/pull/177/changes/89993fbc3c58130ec2fa69d9d870c453902132f5

aprimakina:
> What happens if you use update-password, save-password and tiger db create role with a read replica? For the record I think its fine if update-password and create role fail (its a read replica after all so read only) but it should probably have a good error. For save-password I genuinely am not sure if it should override the parents password or error out.

That was exactly the changes I was still working on (https://github.com/timescale/tiger-cli/pull/177/changes/94aadd70e0deeaf5bc71f1bf5c1c12e809ea19a5), `update-password` and `db create role` now reject a read replica ID up front, before any connection, with a pointer to the primary.

For `save-password` I went with redirect rather than error. Since a read replica shares the primary's credentials, save-password <replicaID> stores the password under the parent, which is exactly where db connect/test-connection read it. Erroring would've been consistent with the other two. Open to switching it to reject if you'd prefer consistency!

PR #179, opened 2026-07-27
refactor: one file per command and per MCP tool; flatten internal/tiger

nathanjcochran:
Pure file/package reorganization to make the repo easier to navigate. Three changes:

1. Moved everything from `internal/tiger/` up into `internal/`. The extra level of hierarchy wasn't buying us anything.

2. Split the CLI command files so each command gets its own file in `internal/cmd/`, named to match the command in snake_case (`tiger service create` → `service_create.go`). `service.go` was 2000+ lines and `db.go` and `mcp.go` weren't far behind. Group commands keep their own file and hold the helpers shared across their subcommands. Test files follow the same layout, with package-wide test scaffolding in `main_test.go`.

3. Split the MCP tool files the same way — one file per tool, each laid out as input/output schemas, then a `new*Tool()` constructor, then the handler. Tool registration moved into `server.go`, so adding a tool is one new file plus one `addTool` line.

No behavior changes. The only code that isn't a straight move is the MCP tool registration: the inline `&mcp.Tool{...}` literals became `new*Tool()` functions, and `registerServiceTools`/`registerDatabaseTools` are now just lists of `addTool` calls. Everything else is byte-identical to before, verified by diffing every top-level declaration's source text against main.

CLAUDE.md and docs/development.md are updated to describe the new conventions.


Askir (approved):
Only really looked at the CLI part but that generally looks good to me 👍 
I just wonder if we can clear up a bit where the shared helpers actually lie I find it a bit confusing that e.g. `checkStdinIsTTY` and `readPasswordFromTerminal` are still in db.go.

I asked claude about other unclear cases and it suggested a rule set like this:

<details>
<summary>Claude output</summary>

1. **Used by one command** → that command's file.
2. **Used by several commands in one group** → the group file (`db.go`, `service.go`).
3. **Used across groups, must stay in `package cmd`** (test-override vars, cobra types) → a named package-level file: `completion.go`, `flag.go`, or a new `terminal.go`.
4. **Used across groups, no dependency on `cmd`** → `internal/util` or `internal/common`.
5. **Used by both CLI and MCP** → `internal/common` (already the documented rule; this just makes it the same rule rather than a separate one).
Claude Stop

</details>

And it says these functions would have to be moved then:

<details>
<summary>Claude output</summary>

- `checkStdinIsTTY`, `readPasswordFromTerminal` — in `db.go`, used by `service_update_password.go`, `read_replica.go`, `password_recovery.go` → rule 3, new `terminal.go`, joined by `readString` (currently squatting in `root.go`).
- `generateSecurePassword` — in `db.go`, nothing db-specific, used by the `service update-password --auto-generate` path → rule 4, `internal/util`.
- `serviceIDCompletion` — in `service.go`, used by both groups → rule 3, `completion.go`, which exists for exactly this.
- `updateNumberBuffer` — in `oauth.go`, borrowed by `mcp_install.go` → rule 3, alongside the other shared TUI bits.
- `password_recovery.go` — currently two flows in one file. The rotation trio (`promptAndResetPassword` → `resetServicePassword` → `updateAndSaveServicePassword`) is the genuinely shared kernel and is duplicated in `internal/mcp/service_update_password.go` → rule 5, `internal/common`. Everything else (bubbletea menu, `testConnectionWithPassword`, `isAuthenticationError`, `testSaveAndLaunchPsqlWithPassword`) is reachable only from `db connect` → rule 1, fold into `db_connect.go`, which also resolves the current `password_recovery.go` → `launchPsql` back-reference. File stops existing.
</details>

This seems sensible to me, but curious what you think? Not sure if we want to fold it in here if you agree.

Also:
<details>
<summary>Claude output</summary>

> - CLAUDE.md:480 and :994 still say .goreleaser.yml; this PR renamed it to .goreleaser.yaml.
> - docs/CODE_REVIEW.md (11 refs) and docs/DECISIONS.md:45 still point at internal/tiger/…. Point-in-time artifacts, so arguably delete rather than fix — but they're the only stale paths left in the repo.

</details>

nathanjcochran:
Yes, totally agree that we could (and should) go further with the restructuring. This was just kind of intended to be a big initial push in the right direction, with hopefully more smaller (and easier to review) PRs to follow 🤞.

PR #181, opened 2026-08-03
refactor: move shared helpers to their correct files

nathanjcochran:
Follow-up to #179, addressing [Askir's review feedback](https://github.com/timescale/tiger-cli/pull/179#pullrequestreview-4846275764) about where shared helpers should live.

The main outcome is a written-down rule for helper placement in CLAUDE.md — place a helper by who calls it, working down the list until one matches:

1. One command → that command's file.
2. Several commands in one group → the group file.
3. Across groups → a named package-level file.
4. A genuine standalone utility, with no notion of a command → `internal/util`.
5. Used by both CLI and MCP → `internal/common`.

Applying it:

- Cross-group helpers in `cmd` now live in `*_helper.go` files, so anything without that suffix is a command: new `terminal_helper.go` and `password_helper.go`, plus renamed `completion_helper.go` and `flag_helper.go`.
- All shell completion functions now live in `completion_helper.go` (`serviceIDCompletion`, `configOptionCompletion`, `mcpGetCompletion`), matching how ghost does it.
- `generateSecurePassword` moved to `internal/util` as `util.GenerateSecurePassword`.
- `read_replica.go` and `password_recovery.go` folded into `db_connect.go`, and `oauth.go` into `auth_login.go` — each was reachable from only one command.
- The `service list` output chain moved out of `service.go` into `service_list.go`, and MCP's `ServiceInfo`/`convertToServiceInfo` out of `utils.go` into `service_list.go`.

Also fixed two stale `.goreleaser.yml` references in CLAUDE.md.

As with #179 this is a pure reorganization — every moved declaration is byte-identical to before, verified by diffing declaration text against main. The exceptions are exporting `GenerateSecurePassword` plus its two call sites, and splitting one `var` block so the terminal helpers could move independently.


internal/cmd/password_helper.go
```diff
@@ -0,0 +1,81 @@
+package cmd
+
```
Askir, line 2: any opinion about giving files like this one that do not contain any actual cli command a different name so they are more easily identifiable (like the _test for the test files)?

E.g. password_util.go or password_helper.go?

Askir (approved):
lgtm! Thank you!

internal/cmd/password_helper.go
nathanjcochran: Good call. Done here: https://github.com/timescale/tiger-cli/pull/181/commits/9174df6aa997a24c4696d562f22623544cbbde59.

I also realized that `internal/cmd/oauth.go` could be folded into `internal/cmd/auth_login.go`, since the OAuth flow isn't used anywhere else. So that just left:

- `completion.go` -> `completion_helper.go`
- `flag.go` -> `flag_helper.go`
- `password.go` -> `password_helper.go`
- `terminal.go` -> `terminal_helper.go`

PR #184, opened 2026-08-05
fix(service): don't report READY before the endpoint accepts connections

gonzaloserrano:
`service create` waits on the control-plane `status` field. READY gates on in-cluster signals (leader pod serving, post-deploy runners), never on the customer-facing endpoint, so callers connecting immediately get `ECONNREFUSED` while the allocated port, DNS and LB catch up. `internal/common/ready.go` already documents READY as "accepting connections".

Prod repro in `timescale/migration-smoke-test` ([run 30977861902](https://github.com/timescale/migration-smoke-test/actions/runs/30977861902)): READY 20.1s after create, refused 91ms later. Twice in 40 runs. `start` and `fork` share the pattern.

`WaitForConnectable` polls the endpoint after the status wait, capped at 2 minutes. Any Postgres protocol error counts as serving: `28P01` proves the server is up, just refusing our credentials, so no password is needed. `57P03` is the exception and keeps waiting.

Best-effort: unverifiable endpoints warn rather than fail; a VPC-only service is healthy but unreachable from here.

A server-side gate is worth doing separately, but cannot replace this: readiness is a property of the path between a client and the service, not of the service alone.

For review: do `start`/`fork` belong here, and should the budget be a flag? Call sites covered only by compilation (create tests use `--no-wait`).


gonzaloserrano:
@nathanjcochran PTAL, in case you think this is a real issue (I’m not familiar with the codebase or API semantics) I’ll mark it as ready for review.

nathanjcochran:
I'm fairly surprised by this. I'm pretty confident that the deployer only sets the status to `READY` after verifying the database is reachable via the same external DNS name that customers use. I asked Claude to look into this, and this is what it said:

> There is already a server-side gate. To reach READY, [the leader pod must be serving *and* every post-deploy runner must return nil](https://github.com/timescale/savannah-deployer/blob/2752d5d9fe15806ba4aa5ebeb75484fbc1198bb1/internal/controller/leader_status.go#L137-L178), and one of those runners is [`externalDNSChecker`](https://github.com/timescale/savannah-deployer/blob/2752d5d9fe15806ba4aa5ebeb75484fbc1198bb1/cmd/savannah-deployer/main.go#L384-L399) — it [opens a real libpq connection](https://github.com/timescale/savannah-deployer/blob/2752d5d9fe15806ba4aa5ebeb75484fbc1198bb1/internal/deployer/status/recovery.go#L62-L99) to [the external DNS name on the allocated port](https://github.com/timescale/savannah-deployer/blob/2752d5d9fe15806ba4aa5ebeb75484fbc1198bb1/pkg/dbutils/connector.go#L142-L159), the [same hostname and port the API hands the customer](https://github.com/timescale/savannah-deployer/blob/2752d5d9fe15806ba4aa5ebeb75484fbc1198bb1/internal/server/connectivity.go#L63-L69).

So I'm not sure that the premise of this PR is correct. If there _is_ a bug, I don't think it's because "READY gates on in-cluster signals (leader pod serving, post-deploy runners), never on the customer-facing endpoint, so callers connecting immediately get ECONNREFUSED while the allocated port, DNS and LB catch up." It should already be the case that the deployer waits until all of those things are ready before setting the status to `READY` 🤔. So I think something else must be going on (and I unfortunately don't have the relevant expertise to dig into the deployer and figure out what it is).

Additionally, I'm a little hesitant to add an extra wait condition in the CLI, beyond simply waiting for `READY`. The `READY` status is the signal that Tiger Console uses to determine if a service is ready, and I'd like to maintain parity with the console in that regard, so that all interfaces are consistent and users don't get confused about the status of their service. Furthermore, services in a VPC will _never_ be reachable from the CLI (unless the CLI happens to be running within the VPC), and I don't want the default behavior for VPC-connected services to be that the operation waits until it eventually times out. For that reason and others like it, I'd prefer to keep the current semantics, where we trust the deployer to tell us if a service is ready, rather than trying to infer it ourselves.

Furthermore, this fix doesn't seem to address any of the MCP tools, which have the same waiting behavior. If we _did_ add a new wait condition, I'd want to do it across the board, so that the same logic holds for both MCP tools and CLI commands.

But again, I think there's probably something else going wrong in your failing CI runs that we should dig into first. For whatever reason, it seems the deployer was able to connect to the database fine (and therefore set the status to `READY`) even though the CI runner could not. I think we need to figure out why that is before we make any changes to Tiger CLI.

gonzaloserrano:
Hey, indeed, it’s not correct. Sorry for the noise.

I'm debugging the analysis and it’s wrong in several ways.

I’m checking if its somehow related to DNS.

gonzaloserrano:
Raised https://github.com/timescale/savannah-deployer/issues/2213

PR #185, opened 2026-08-05
Switch logging from zap to slog, remove debug logging

nathanjcochran:
Replaces the `internal/logging` zap wrapper with the standard library's `slog`, adopting the patterns Ghost uses.

Logging is now reserved for the MCP server, which is the only long-running backend process the CLI hosts. `tiger mcp start` builds a logger over the command's stderr and hands it to `mcp.NewServer`, which uses it for its own output and also passes it to the MCP SDK (which uses it for its own internal logging). Everywhere else — CLI commands, analytics, MCP install, the signal handler — writes to stdout/stderr directly or returns an error. Log statements outside the MCP server (which produced structured log lines interspersed with the normal CLI output, which looked pretty bad imo) were removed rather than converted. None of them seemed very useful anyways, except perhaps to the developer originally building the feature (but developers can always put in their own print statements as needed during development - I don't think they need to remain in the committed code).

Within the MCP server the logging is deliberately sparse, matching Ghost: we mostly log failures that would otherwise be swallowed, not per-step tracing. Startup emits almost nothing, so the logs are mostly silent until something goes wrong.

Because there is no longer a log level to configure, the `debug` config option, `TIGER_DEBUG`, and `--debug` are gone. This is technically a backwards-incompatible change, but I doubt many actual end-users were using that feature, so I think it's probably safe to remove. But definitely let me know if you disagree.


Askir (approved):
Looks good to me, I don't think the --debug change is problematic. And we are below 1.x so I think small "breaking" changes like this are definitely okay if we put them into the changelog.

PR #186, opened 2026-08-06
Route all CLI output through the command's stdout/stderr

nathanjcochran:
Adopts Ghost's output conventions. Commands now print with cobra's `cmd.Print*` / `cmd.PrintErr*` helpers instead of `fmt.Print*` or spelled-out `fmt.Fprintf(cmd.OutOrStdout(), …)`, and the root command wires its writers to stdout/stderr explicitly — cobra's `Print*` silently falls back to stderr otherwise (which is why `tiger upgrade` was sending all its progress to the wrong stream). Helpers inside `internal/cmd` take the command and print through it; code in `internal/common` and `internal/version` still takes an `io.Writer` so it stays free of cobra.

The payoff is that every byte a command emits now flows through writers we control, so stream choice is auditable and all output is capturable in tests (I intend to refactor our testing patterns in a later PR, and this helps set the stage for it). Additionally, using the `cmd.Print*` helpers is just cleaner, less verbose, and easier to read overall (imo). Finally, the `cmd.Print*` helpers do not return an error (unlike `fmt.Fprint*`), which is helpful in case we ever want to add [errcheck](https://github.com/kisielk/errcheck) to the CI pipeline (but we're still a long way from that, currently 😅).

This PR also codifies the rule for what goes where: stdout carries the actual command result/payload, whereas stderr carries warnings, progress updates, and anything interactive. Anything that previously did not adhere to that rule has been updated.

Bug fixes and behavior changes that came out of it:

- `tiger mcp install`'s success banner and `tiger auth login`'s credential prompts were writing straight to the process's stdout, bypassing the command entirely and therefore invisible to tests.
- `tiger db test-connection` printed its own diagnostic and then let cobra print the raw error a second time, plus the full usage block. It now reports once.
- `tiger upgrade`'s progress output moves from stderr to stdout.
- Interactive prompts in `auth login` and `db save-password` move to stderr, so they stay visible when stdout is redirected.
- The whole `auth login` OAuth flow — auth URL, browser warnings, and the project picker — moves to stderr. Redirecting stdout previously hid the picker entirely while it sat waiting for keystrokes.

The conventions are written up in CLAUDE.md.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1048,7 +1048,7 @@ if len(args) < 1 {
 
 // Interactive confirmation unless --confirm
 if !confirmFlag {
-    fmt.Fprintf(cmd.ErrOrStderr(), "Type the service ID '%s' to confirm: ", serviceID)
+    cmd.PrintErrf("Type the service ID '%s' to confirm: ", serviceID)
     var confirmation string
     fmt.Scanln(&confirmation)
     if confirmation != serviceID {
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -505,8 +505,8 @@ capabilities — the docs proxy being disabled, and each write tool skipped in
 read-only mode — so a client can see why a tool it expected is missing.
 
 Everything outside `internal/mcp` writes to stdout/stderr directly rather than
-logging. Don't add log statements to CLI commands — print to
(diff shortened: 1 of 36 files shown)


Askir (approved):
I think CI/CD is red due to a flake. Looks good to me though!

PR #187, opened 2026-08-06
Read stdin through the command, with shared read helpers

nathanjcochran:
Follow-up to #186, applying the same idea to input. Commands read through `cmd.InOrStdin()` rather than `os.Stdin`, and the ad-hoc terminal helpers in `internal/cmd` are replaced by `util.ReadLine` / `util.ReadPassword` / `util.ReadAll`, ported from Ghost.

The benefit is one place that knows how to read a line, a password, or piped input, instead of each command hand-rolling `bufio` and `term` calls. All three helpers run the blocking read on a goroutine and select on the context, so Ctrl-C unblocks a waiting prompt instead of hanging until the user hits enter. Tests drive the interactive paths by stubbing `util.IsTerminal` and `util.ReadPassword` centrally, rather than every command carrying its own override hooks.

The interactive menus needed the same treatment. BubbleTea silently defaults to `os.Stdin` when `WithInput` is omitted — and there's no compile error to catch it — so all four pickers (project, read replica, password recovery, MCP client) were reading real stdin regardless of what the command was given. They now get both streams explicitly, which is why `oauthLogin` carries the `*cobra.Command` instead of a writer.

Finally, every prompt that blocks on a real-time answer is now gated on whether stdin is a terminal, and names the flag to use when it isn't. Three were missing that check and would fail on an opaque EOF, or hang, when run in CI.

Bug fixes and behavior changes:

- The password prompt never saved or restored terminal state, so cancelling one could leave the shell in raw mode with echo off. Affects `db save-password`, `db connect`'s password recovery, `service update-password`, and `auth login`.
- `tiger mcp install` with no argument now tells you to name a client, instead of failing somewhere inside BubbleTea.
- `tiger service delete` without `--confirm` now points at that flag rather than dying on `failed to read confirmation: EOF`. Note this also stops a piped confirmation (`echo svc-x | tiger service delete svc-x`) from working — `--confirm` is the automation path for a destructive command.
- `tiger auth login` with multiple projects and no TTY points at `--public-key` / `--secret-key` instead of erroring out of the picker.

The conventions are written up in CLAUDE.md.


internal/util/read.go
```diff
@@ -0,0 +1,92 @@
...
+	return readString(ctx, func() (string, error) {
+		val, err := term.ReadPassword(int(f.Fd()))
+		if err != nil {
+			return "", err
+		}
+		return string(val), nil
+	})
+}
+
+func ReadAll(ctx context.Context, stdin io.Reader) (string, error) {
```
Askir, line 42: This seems to have no callers, I guess it's still fine to add since you ported it over. Just calling that out.

internal/util/read.go
```diff
@@ -0,0 +1,92 @@
...
+		err error
+	}
+
+	resultCh := make(chan result, 1)
+	go func() {
+		defer close(resultCh)
+
+		val, err := readFn()
+		if err != nil {
+			resultCh <- result{err: err}
```
Askir, line 75: This breaks for piped input apparently if it doesn't end with a `\n`, idk if that's a real use case though so probably fine.

Askir (approved):

internal/util/read.go
nathanjcochran: Yeah, I know. I decided to include it because we'll need it when we create the new command for executing SQL directly (without `psql`), since we'll want that to accept multi-line input via stdin.

internal/util/read.go
nathanjcochran: Yeah, I don't think that's a real use case for `ReadLine()`. I think all of the usages right now are gated with `util.IsTerminal(cmd.InOrStdin())` anyways, so I don't think it's even possible to call it with piped input currently. If that changes in the future, we can revisit.

And fwiw, I think `ReadAll()` _does_ work fine with piped input (and it's more intended for that use case).

PR #190, opened 2026-08-12
Add oapi-codegen config files

nathanjcochran:
Tiger CLI drove oapi-codegen with inline flags, and was lacking a couple of nice options that ghost sets. This PR adds oapi-codegen config files matching ghost's (`internal/api/types.yaml` and `internal/api/client.yaml`), and switches the `go:generate` lines to use them.

The two options that matter are `name-normalizer: ToCamelCaseWithInitialisms`, which capitalizes initialisms [the way Go does](https://go.dev/wiki/CodeReviewComments#initialisms) (`ServiceID` rather than `ServiceId`, `CPUMillis` rather than `CpuMillis`, etc.), and `always-prefix-enum-values`, which prefixes enum constants with their type (`api.DeployStatusREADY` rather than `api.READY`) so values from different enums can't collide and their purpose is more obvious/clear.

Most of the diff is the regenerated files plus the renames propagated through their callers. Behavior is unchanged: with case folded, the regenerated `client.go` is identical to the old one and `types.go` differs only in the enum prefixes.

Also fixes the spec path in the `go:generate` lines, which pointed one level above the repo root (probably because of the change to get rid of an extra level of folder hierarchy in #179).


Askir (approved):
lgtm!

PR #192, opened 2026-08-19
Sync openapi.yaml

areina:
Update the spec to get the latest changes. There shouldn't be behavioral changes.

- update the spec and regenerate the API client
- some service fields are now required, changed related Go code from pointers to plain values
- fixed a wrong path in the generate command
- kept the error "details" field by hand, the API still sends it


nathanjcochran (approved):
LGTM! Thanks!!

PR #193, opened 2026-08-19
Add service backups command and MCP tool 

areina:
- lists the backups for a service, preview only
- table doesn't display the label, json and yaml keep it
- sizes use docker/go-units

Example:
```
> TIGER_EXPERIMENTAL=1 tiger service backups <service-id>

  ┌──────────────────┬─────────────┬──────────┬──────────┬──────────────────────────────────────────────┐
  │     STARTED      │    TYPE     │ DURATION │   SIZE   │                   REGIONS                    │
  ├──────────────────┼─────────────┼──────────┼──────────┼──────────────────────────────────────────────┤
  │ 2026-08-19 09:30 │ FULL        │ 11m12s   │ 4.5GiB   │ us-east-1 (FINISHED), eu-central-1 (RUNNING) │
  │ 2026-08-19 14:00 │ INCREMENTAL │ 45s      │ 50.83MiB │ us-east-1 (FINISHED), eu-central-1 (FAILED)  │
  └──────────────────┴─────────────┴──────────┴──────────┴──────────────────────────────────────────────┘


```


internal/cmd/service_backups.go
```diff
@@ -0,0 +1,152 @@
...
+// TIGER_EXPERIMENTAL in buildServiceCmd.
+func buildServiceBackupsCmd(app *common.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "backups [service-id]",
+		Aliases: []string{"backup"},
+		Short:   "List backups for a service",
+		Long: `List the full and incremental backups taken for a database service.
+
+Backups run automatically on a schedule; there is no command to create or delete
+one. To restore data, create a recovery fork with tiger service fork.
```
nathanjcochran, line 29: Question: do we think we'll _ever_ need to support other backup-related commands (e.g. for creating or deleting them?). If we think that's a possibility, it might make more sense for this to be `tiger service backup list`, to make room for other subcommands in the future (e.g. `tiger service backup create`). But if that's not something we think we'll ever support, then I agree that `tiger service backups` makes more sense (it matches `tiger service logs`, for example).

In any case, I think the current approach is fine for now (especially because this is an experimental command), but I just figured I'd pose the question in case it influences the design at all.

internal/cmd/service_backups.go
```diff
@@ -0,0 +1,152 @@
...
+				cmd.SilenceUsage = true
+				return err
+			}
+
+			serviceID, err := getServiceID(cfg, args)
+			if err != nil {
+				return err
+			}
+
+			cmd.SilenceUsage = true
```
nathanjcochran: I just merged a PR (https://github.com/timescale/tiger-cli/pull/194) that sets `SilenceUsage: true` on the top-level `cobra.Command` struct for every command, instead of setting it inline like this. That ensures that the help text isn't output if the user mistypes a flag or argument (since it buries the actual error). I think we should update this to follow the new pattern.

internal/cmd/service_backups.go
```diff
@@ -0,0 +1,152 @@
...
+
+// outputBackupsTable omits the label: it repeats STARTED and TYPE, and no command
+// takes it as input. It stays in the json and yaml output.
+func outputBackupsTable(backups []api.Backup, output io.Writer) error {
+	table := tablewriter.NewWriter(output)
+	table.Header("STARTED", "TYPE", "DURATION", "SIZE", "REGIONS")
+
+	for _, backup := range backups {
+		table.Append(
+			backup.StartedAt.Format("2006-01-02 15:04"),
```
nathanjcochran: This displays no timezone info. Is the assumption that it's UTC? Will users know that? Should it be explained in the command description? Or maybe this format should include a timezone? 🤔

nathanjcochran (approved):
Left a few small comments/questions, but LGTM! Tested it and it works well ✅

internal/cmd/service_backups.go
areina: That's a very good point. I don't think it will be possible to create/delete backups, but I'm already adding endpoints to enable cross-region backups and also to set the retention. Using `backup` instead of `backups` is a better options. Thanks for catching this, will update it.

internal/cmd/service_backups.go
areina: I will update this PR, thanks for the heads up.

internal/cmd/service_backups.go
areina: Good point. Do we have any convention about how to display timestamps in the CLI? It would be worth to have one and be consistent between different commands.

nathanjcochran: I am not aware of any existing convention.

It looks like the format you used here matches the format of the `tiger service list` output (see [here](https://github.com/timescale/tiger-cli/blob/f93dac59e2241445c9b49af0fdf6063b0cf21da4/internal/cmd/service_list.go#L118)). However, the `tiger service create`/`fork`/`get` output includes the timezone (see [here](https://github.com/timescale/tiger-cli/blob/f93dac59e2241445c9b49af0fdf6063b0cf21da4/internal/cmd/service.go#L163)). So we're currently not being consistent.

I think we'd ideally adopt a consistent convention across the board, but that's probably out-of-scope for this PR. Feel free to pick whichever approach you think makes the most sense for now, and we can consolidate later on in a follow-up PR.

internal/cmd/service_backups.go
```diff
@@ -0,0 +1,152 @@
...
+
+			serviceID, err := getServiceID(cfg, args)
+			if err != nil {
+				return err
+			}
+
+			cmd.SilenceUsage = true
+
+			ctx, cancel := context.WithTimeout(cmd.Context(), 30*time.Second)
+			defer cancel()
```
nathanjcochran: These `context.WithTimeout` calls were also redundant with a global timeout set in `internal/api/client_utils.go`, and I just merged a PR (https://github.com/timescale/tiger-cli/pull/198) that gets rid of them everywhere. So you can probably get rid of these `context.WithTimeout` calls here and in the MCP tool handler too.

PR #195, opened 2026-08-20
Extend read_only config into all/prod/off modes

aprimakina:
`read_only` was a bool: writes were refused either everywhere or nowhere. It now takes three modes, so an agent can be let loose on development services while production stays protected:

-   all:   every service (equivalent to the old `true`)
-   prod:  only services tagged PROD
-   off:   nothing (equivalent to the old `false`, and still the default)

The booleans keep working - `read_only: true` in a config file, or TIGER_READ_ONLY=true, still mean `all`. The default stays `off`; flipping it to `prod` is a separate decision, since it would change behavior for production services - notably `tiger db connect`, which would start opening read-only sessions.

Because `prod` protects only services tagged PROD, every gate needs its target's environment tag. common.CheckReadOnly(cfg, tag) is the verdict, and CheckReadOnlyByServiceID fetches the service when only its ID is known - one extra API call, and only under `prod`. Both require a tag, so a gate cannot silently skip prod mode.

Notes on the individual surfaces:

- MCP registration cannot express `prod`, so the write tools stay registered (they still work on DEV services) and refuse per call instead. The server instructions gained a prod variant so a refusal does not look like a bug, and ServiceDetail now reports each service's environment tag so the model can check before offering to modify one.
- Database sessions reach the same verdict as the control-plane gate: a mode that refuses writes to a service also opens its connections read-only. That covers the connection strings embedded in service get/list output, not just the db commands.
- A replica set is judged on its own environment tag rather than its primary's, on both planes, matching how the API tags replica sets individually.
- `db create role` is gated too, and the `db connect` password-recovery menu no longer offers to rotate the password of a protected service.

## First-login prompt

`read_only` defaults to `off`, so the protection above only reaches users who go looking for it. Flipping the default would change behaviour for every existing user — notably `tiger db connect`, which would start opening read-only sessions against `PROD` services, with Postgres reporting the refusal so nothing points at the setting.

A first interactive `tiger auth login` now offers it instead:

```
Protect services tagged PROD from writes? [Y/n]:
```

- Triggered only when no credentials existed beforehand, so existing users are never asked and no default changes. The answer is written to the config file, so `config show` reflects a choice rather than a version-dependent default.
- Both answers are recorded, so it asks exactly once. Declining stores `read_only: off` rather than leaving the key absent — absence can't record a negative answer, so a logout and second login would otherwise ask again. The guard tests key *presence* rather than the resolved value, since `off` is also the default.
- Skipped when stdin isn't a terminal, so scripted and CI logins neither block nor change behaviour. Unlike the CLI's other TTY gates it skips silently rather than erroring, since nothing is being refused.

### DIFF
--- a/internal/cmd/auth_login.go
+++ b/internal/cmd/auth_login.go
@@ -31,8 +31,21 @@ const nextStepsMessage = `
 • Install MCP server for your favorite AI coding tool: tiger mcp install
 • List existing services: tiger service list
 • Create a new service: tiger service create
-• Enable read-only mode: tiger config set read_only true
 `
+
+// readOnlyNextStep tells the user how to set read_only by hand. Printing it to
+// someone who just picked a mode from the menu would read as if their answer
+// didn't take, so it's only for logins that left the choice unmade.
+const readOnlyNextStep = "• Protect services from writes: tiger config set read_only prod (or all, off)\n"
+
+// nextSteps is the post-login message, picking up readOnlyNextStep when the
+// config file holds no read_only value.
+func nextSteps(readOnlySet bool) string {
+	if readOnlySet {
+		return nextStepsMessage
+	}
+	return nextStepsMessage + readOnlyNextStep
+}
 
 var (
 	// openBrowser can be overridden for testing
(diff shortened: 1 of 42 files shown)


internal/cmd/auth_login.go
```diff
@@ -158,11 +163,54 @@ Examples:
 	return cmd
 }
 
-func finishLogin(cmd *cobra.Command, projectID string) {
+func finishLogin(cmd *cobra.Command, cfg *config.Config, projectID string, firstLogin bool) {
 	cmd.Printf("Successfully logged in (project: %s)\n", projectID)
+
+	if firstLogin {
+		offerProdProtection(cmd, cfg)
+	}
```
nathanjcochran: I'm not sure we need to limit this to just the first login. I think it's probably fine if we show this to pre-existing users the first time they run `tiger auth login` after this change, since many of them might not know about this feature yet. Imo, the logic below (which doesn't prompt if they already have `read_only` config option set in their config) is probably sufficient on its own, without _also_ including this "first login" check.

Additionally, I'm wondering if it would make sense to show the prompt again any time the user logs into a different project, or switches projects? Since a user's preference for whether to protect prod services might depend at least in part on the project they're in? 🤔

internal/cmd/auth_login.go
```diff
@@ -158,11 +163,54 @@ Examples:
...
+	answer, err := util.ReadLine(cmd.Context(), cmd.InOrStdin())
+	if err != nil {
+		// Interrupted or EOF - not an answer, so record nothing and ask again.
+		return
+	}
+
+	mode := config.ReadOnlyProd
+	if strings.HasPrefix(strings.ToLower(answer), "n") {
+		mode = config.ReadOnlyOff
+	}
```
nathanjcochran: I think it would be better to check for exact strings here (e.g. `y`/`yes`, or `n`/`no`), rather than assuming that any string that begins with `n` means "no" and all other strings mean "yes". Imo, if the user doesn't enter `y`/`yes`/`n`/`no` (or an empty string, which should default to `yes` given the capital `Y` in `Y/n`), I think we should print a warning and either prompt again, or simply return early without saving the config value. I think it's probably better to require a clear and explicit response, rather than to assume other strings have meaning. See how Ghost does it, for reference.

Alternatively, we could show an interactive menu with all three read-only options, rather than a yes/no decision that only maps to two of the three possible options. That honestly might be the best approach, and it would also help teach the user about all of the different read-only modes available.

internal/common/service.go
```diff
@@ -4,13 +4,32 @@ import (
...
 // Matches front-end logic for generating a random service name
 func GenerateServiceName() string {
 	return fmt.Sprintf("db-%d", 10000+rand.Intn(90000))
 }
 
+// ServiceEnvironmentTag returns a service's environment tag. A fetched service
+// carries it as a free-form string under metadata, not the enum the create and
+// fork requests take. Only PROD is recognized: absent, empty and unknown values
+// all read as DEV, because metadata is optional in the response and treating
+// unknowns as protected would have read_only=prod block untagged services.
```
nathanjcochran: Is it _really_ the case that the environment tag is optional? Like, is it actually possible to create a service with no environment tag? Or is it just marked as optional currently due to being misrepresented in the REST API layer? Because if it's actually always present, we should consider updating the REST API's `openapi.yaml` spec to mark it as a required field, which would simplify this logic a bit.

We should also probably considering making the metadata environment field use the same enum as the endpoint for setting the environment tag, for consistency.

If we did both of those things, we could probably get rid of this function altogether (doesn't have to happen as part of this PR, but would be good not to lose track of it).

internal/config/config.go
```diff
@@ -38,6 +39,66 @@ const (
...
+// BlocksAll reports whether the mode blocks writes to every service. It's the
+// only verdict reachable without knowing the target, so it gates the surfaces
+// that have no service in hand.
+func (m ReadOnlyMode) BlocksAll() bool {
+	return m == ReadOnlyAll
+}
+
+// ParseReadOnlyMode converts a user-supplied read_only value into a ReadOnlyMode,
+// still accepting the booleans this option took before the modes existed.
+func ParseReadOnlyMode(value string) (ReadOnlyMode, error) {
```
nathanjcochran: Does this function really need to be public/exported? I think it's only called from within this package, right? In that case, it should probably be private/unexported (i.e. `parseReadOnlyMode`).

internal/common/errors.go
```diff
@@ -1,24 +1,73 @@
...
+		return fmt.Errorf("service %s: %w", serviceID, err)
+	}
 	return nil
 }
 
+// ForcesReadOnlySession reports whether read-only mode requires a database
+// session against this service to open in Tiger Cloud's immutable read-only mode.
+func ForcesReadOnlySession(cfg *config.Config, service api.Service) bool {
+	return CheckReadOnly(cfg, ServiceEnvironmentTag(service)) != nil
+}
```
nathanjcochran: I'm not sure `errors.go` is the best place for these functions. Perhaps they should go in a new `read_only.go` file instead, or something like that?

internal/common/replica.go
```diff
@@ -19,6 +19,13 @@ type ConnectionTarget struct {
 	IsReplica         bool
 }
 
+// ReadOnlySession reports whether read-only mode requires this target's session to
+// open in Tiger Cloud's immutable read-only mode. Only ConnectionService decides,
+// so a replica is judged on its own environment tag rather than its primary's.
+func (t *ConnectionTarget) ReadOnlySession(cfg *config.Config) bool {
+	return ForcesReadOnlySession(cfg, t.ConnectionService)
+}
+
```
nathanjcochran: I feel like there are too many different helper functions for checking read-only mode:
- `CheckReadOnly`
- `CheckReadOnlyByServiceID`
- `ForcesReadOnlySession`
- `ReadOnlySession`

I think that ideally, there would just be a single canonical function for checking read-only mode. When reviewing this, it was confusing to see different functions being called in different places for more-or-less the same purpose (especially because of the differences in naming, which made me think some of the functions have different behavior, when they really don't).

`ReadOnlySession` and `ForcesReadOnlySession` , in particular, are _very_ thin wrappers around `CheckReadOnly`, and could probably be eliminated in favor of just calling `CheckReadOnly` directly (it would still be just a single line of code at the call sites). Imo, that would make the code for checking read-only mode a lot more uniform (and would also make it easier to find all of the read-only mode checks across the codebase, since you could just look for one or two functions with variations of the same name - i.e. `CheckReadOnly*`). I think `CheckReadOnlyByServiceID` still probably makes sense to keep as a helper function, since it does the extra work of fetching the service beforehand, but the others don't feel like they're pulling their weight, and are probably just making things more confusing imo.

internal/mcp/server.go
```diff
@@ -64,17 +66,32 @@ func addTool[In, Out any](s *Server, readOnly bool, t *mcp.Tool, h mcp.ToolHandl
...
-	intro := "Tiger MCP provides tools for managing and querying Tiger Cloud database services (managed TimescaleDB/PostgreSQL). "
+	const (
+		intro        = "Tiger MCP provides tools for managing and querying Tiger Cloud database services (managed TimescaleDB/PostgreSQL). "
+		capabilities = "Use it to provision and fork services, start/stop/resize instances, rotate credentials, fetch service logs, execute SQL queries, and search Tiger documentation."
+	)
+
+	mode := config.ReadOnlyOff
+	if cfg != nil {
+		mode = cfg.ReadOnly
+	}
```
nathanjcochran: Can the passed-in `*config.Config` ever actually be `nil`? I don't think we need to be defensive against things that can never actually happen (passing `nil` to this function instead of an actual config would be a programming bug imo, and it would therefore be acceptable to panic). I think it would be fine to just reference `cfg.ReadOnly` directly.

internal/mcp/service_create.go
```diff
@@ -104,7 +104,9 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
 		return nil, ServiceCreateOutput{}, err
 	}
 
-	if err := common.CheckReadOnly(cfg); err != nil {
+	// Deliberately DEV-only: this tool takes no environment_tag, unlike the CLI's
+	// --environment, so the API always tags what it creates DEV.
+	if err := common.CheckReadOnly(cfg, api.EnvironmentTagDEV); err != nil {
```
nathanjcochran: We should probably consider adding an `environment` parameter to this MCP tool (as well as `service_fork`) to match the CLI command. Not sure why it doesn't take that parameter right now - probably an oversight. Doesn't need to be part of this PR, but might be worth a Linear ticket so we don't lose track of it.

nathanjcochran (approved):
Looks good to me! I left a handful of minor comments that would be nice to address (some of which could be done as follow-up PRs), but I didn't notice any actual bugs.

Thanks for tackling this!

internal/mcp/server.go
aprimakina: Fixed

internal/config/config.go
aprimakina: Fixed.

internal/cmd/auth_login.go
aprimakina: > I'm not sure we need to limit this to just the first login…

Agreed, and done — `firstLogin` is gone. 

> I'm wondering if it would make sense to show the prompt again any time the user logs into a different project, or switches projects?

I'd prefer not to; this sounds like a candidate for a new future - per-project config that we may want to introduce one day.

internal/mcp/service_create.go
aprimakina: https://linear.app/tigerdata/issue/PE-663/add-environment-parameter-to-service-create-and-service-fork-mcp-tools

internal/common/service.go
aprimakina: The REST layer sets it unconditionally; however, it may return UNSPECIFIED (could be backward compatibility for very old services).
https://linear.app/tigerdata/issue/PE-665/make-service-metadataenvironment-required-and-typed

PR #197, opened 2026-08-24
Add tiger project command and --project-id login flag

aprimakina:
tiger project <project-id> switches the active project for an OAuth login: it verifies access via /projects, rewrites the stored credentials, clears the stale service_id default, and rebuilds the API client so a later token refresh persists the new project. API-key logins (stored or via env vars) are refused with the auth exit code, since a key is scoped to a single project.

tiger auth login --project-id picks the project up front: the OAuth flow skips the interactive picker (and validates access), while the API-key flow errors on a mismatch with the key's own project. Any login that doesn't land on the known previous project also clears the service_id default, sharing clearStaleDefaultService with tiger project; when the effective value comes from --service-id or TIGER_SERVICE_ID, the command warns instead of claiming it was cleared.

Supporting changes from review:
- main.go extracts common.ExitCodeError with errors.As so tiger's exit codes survive fmt.Errorf wrapping, while foreign ExitCode() carriers (psql) still count only unwrapped
- common.EnvAPIKey is the single owner of the env-credential precedence rule
- project fetching and access checks are shared between login and project via project_helper.go
- positional args can be excluded from analytics via analytics.OmitArgsAnnotation; tiger project uses it since its argument is a project ID
- test setup unsets TIGER_PUBLIC_KEY/TIGER_SECRET_KEY so host env keys can't break the suite; assertExitCode moved to main_test.go and older hand-rolled exit-code assertions converted


cmd/tiger/main.go
```diff
@@ -9,11 +9,18 @@ import (
...
 )
 
 func main() {
 	if err := run(); err != nil {
-		// Check if it's a custom exit code error
+		// A common.ExitCodeError anywhere in the chain sets the exit code.
+		// Foreign carriers (psql's *exec.ExitError) count only unwrapped, so
+		// wrapped subprocess codes can't collide with tiger's exit codes.
+		var codeErr common.ExitCodeError
+		if errors.As(err, &codeErr) {
```
RobAtticus: this is not technically needed since `common.ExitCodeError` implements the anonymous interface in the next block

RobAtticus (approved):

cmd/tiger/main.go
nathanjcochran: But we do need to unwrap the `err`, since it could have been wrapped by a call to `fmt.Errorf("error: %w", err)`, for instance, and the wrapped error would _not_ implement the anonymous interface below. So I think that's what this change is achieving.

But given that, I think we could probably get rid of the anonymous interface check below, since I don't think we have any errors with an `ExitCode()` method that _aren't_ `common.ExitCodeError`.

internal/cmd/project.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/timescale/tiger-cli/internal/analytics"
+	"github.com/timescale/tiger-cli/internal/api"
+	"github.com/timescale/tiger-cli/internal/common"
+)
+
+func buildProjectCmd(app *common.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "project <project-id>",
```
nathanjcochran: I'm not sure `tiger project` should be the command that switches projects. I'd prefer `tiger project use <project-id>` or `tiger project switch <project-id>` (one of those should probably be an alias for the other).

That's because in the future, I imagine we will probably add other project-related subcommands, like `tiger project list`, `tiger project rename`, `tiger project leave`, etc. (see the `ghost space` subcommands for comparison), and I think it would be strange for the top-level command to be the command that switches projects if it also has subcommands. As of right now. Tiger CLI's top-level commands never have behavior of their own if they also have subcommands (they're just used as a namespace/grouping mechanism).

internal/cmd/project.go
```diff
@@ -0,0 +1,104 @@
...
+cleared when you switch away.
+
+Example:
+  tiger project my-project-id`,
+		Args: cobra.ExactArgs(1),
+		// The argument is a project ID, which analytics must not record.
+		Annotations:       map[string]string{analytics.OmitArgsAnnotation: "true"},
+		ValidArgsFunction: cobra.NoFileCompletions,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
```
nathanjcochran: I just merged a PR earlier today (https://github.com/timescale/tiger-cli/pull/194) that sets `SilenceUsage: true` for all commands outside of the RunE function, thereby ensuring that the help text isn't printed if the user calls the command with an invalid flag or argument. I think we should make that same change here, to match the new pattern.

internal/cmd/project.go
```diff
@@ -0,0 +1,104 @@
...
+with that project's keys instead.
+
+The default service (config key service_id) belongs to the project it was set in, so it is
+cleared when you switch away.
+
+Example:
+  tiger project my-project-id`,
+		Args: cobra.ExactArgs(1),
+		// The argument is a project ID, which analytics must not record.
+		Annotations:       map[string]string{analytics.OmitArgsAnnotation: "true"},
```
nathanjcochran: Why can't we record a project ID in the analytics event? I don't think its sensitive information. In fact, we already [record the project ID as an extra field](https://github.com/timescale/tiger-cli/blob/e830526478c7c07d251b297a329ec862bd8f6dc0/internal/analytics/analytics.go#L175-L177) in every single analytics event we track. So I'm not sure why we'd go out of our way to omit it here. I'd vote to get rid of this, along with the `OmitArgsAnnotation` machinery (I'm not sure it's the best way to achieve this anyways).

internal/cmd/project_helper.go
```diff
@@ -0,0 +1,57 @@
...
+func fetchProjects(ctx context.Context, client api.ClientWithResponsesInterface) ([]api.Project, error) {
+	resp, err := client.GetProjectsWithResponse(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list projects: %w", err)
+	}
+	if resp.JSON200 == nil {
+		return nil, common.ExitWithErrorFromStatusCode(resp.StatusCode(), resp.JSON4XX)
+	}
+	return *resp.JSON200, nil
+}
```
nathanjcochran: Is this helper function really pulling its weight?  I'm personally skeptical that we need helper functions that just wrap an API client call + a check for a nil body. I'd just inline this where it's called (matching the existing pattern for API calls everywhere else in the code).

nathanjcochran (changes requested):

internal/cmd/project.go
aprimakina: Fixed ✅

internal/cmd/project_helper.go
aprimakina: Inlined, both call sites now do the GetProjectsWithResponse + JSON200 check directly.

internal/cmd/project.go
aprimakina: Done.

internal/cmd/project.go
aprimakina: Done.

cmd/tiger/main.go
aprimakina: Good idea, the one holdout was psql's bare *exec.ExitError from db connect.

internal/cmd/project_use.go
```diff
@@ -0,0 +1,105 @@
...
+					errors.New("an API key is scoped to a single project. Run 'tiger auth login' without --public-key/--secret-key"))
+			}
+
+			if targetID == currentProjectID {
+				cmd.Printf("Already using project %s\n", targetID)
+				return nil
+			}
+
+			ctx, cancel := context.WithTimeout(cmd.Context(), 30*time.Second)
+			defer cancel()
```
nathanjcochran: These timeouts were redundant with a timeout set in the client itself, so I removed them in this PR: https://github.com/timescale/tiger-cli/pull/198. Probably makes sense to remove this one too (it isn't buying us anything).

internal/cmd/project_use.go
```diff
@@ -0,0 +1,105 @@
...
+with that project's keys instead.
+
+The default service (config key service_id) belongs to the project it was set in, so it is
+cleared when you switch away.
+
+Example:
+  tiger project use my-project-id`,
+		Args:              cobra.ExactArgs(1),
+		SilenceUsage:      true,
+		ValidArgsFunction: cobra.NoFileCompletions,
```
nathanjcochran: We should be able to provide a shell completion function here that completes project IDs. It would work like the existing completion logic for service IDs, in that it would have to fetch the list of projects in order to generate the completions (see [here](https://github.com/timescale/tiger-cli/blob/acad0099c1cd5f9aafba885ac43dbe5ac6974150/internal/cmd/completion_helper.go#L33-L53)). I believe there is already a REST API endpoint for fetching a list of projects, which could be leveraged for this purpose.

nathanjcochran (approved):
Left two minor comments, but otherwise LGTM!

internal/cmd/project_use.go
aprimakina: https://github.com/timescale/tiger-cli/pull/197/changes/073ac41a54335fe1a7b47bcabd05061e12590786

internal/cmd/project_use.go
aprimakina: https://github.com/timescale/tiger-cli/pull/197/changes/073ac41a54335fe1a7b47bcabd05061e12590786

PR #198, opened 2026-08-24
Simplify shared HTTP client and remove redundant timeouts

nathanjcochran:
## Summary

Our custom HTTP client already enforces a 30-second timeout on every outgoing API request, which made a lot of the `context.WithTimeout(ctx, 30*time.Second)` calls sprinkled through commands and MCP tools redundant — they were just re-imposing a limit the client already had. This PR cleans that up and consolidates how we build and share HTTP clients, using Ghost's approach as a reference point throughout.

## What changed

- **Removed redundant timeouts.** Dropped ~20 `context.WithTimeout(ctx, 30*time.Second)` calls that were wrapping single API calls already covered by the client's own timeout.
- **Simplified the shared HTTP client's transport settings.** Two of the three custom settings (`MaxIdleConns`, `IdleConnTimeout`) were just restating Go's own default values, so they weren't doing anything. The third (`MaxIdleConnsPerHost`) was quietly set higher than Go's default (10 vs. the built-in default of 2) despite its comment claiming to "limit" it, but this wasn't really buying us anything anyways (we don't make a lot of concurrent requests to the same host, and definitely don't need to keep 10 idle connections open). We removed all three and now just rely on Go's defaults.
- **Simplified how the shared client is constructed**, dropping the `sync.Once`-based lazy construction in favor of a plain package variable.
- **Consolidated the CLI's User-Agent header** into one place (the shared client), rather than setting it separately in a couple of different spots. `tiger upgrade`'s release/checksum downloader needs a longer timeout than the shared client allows, so it now clones the shared client and overrides just the timeout, rather than building an entirely separate client that would've missed out on the User-Agent.
- **Widened a couple of timeouts that were arbitrarily shorter** than the client's default (10s) with no real reason to be, so they're no longer unnecessarily prone to premature timeouts (and just to keep things consistent throughout for simplicity's sake).
- **Extended the shared client to a couple of other call sites** (the docs MCP proxy connection, the version-update check) that were previously making HTTP requests through their own separate, ad hoc clients.
- **Updated CLAUDE.md** with a short section describing this pattern, so future work follows it automatically.

## Why

Mainly to cut down on repetitive, easy-to-get-wrong boilerplate and make our HTTP client behavior more consistent and predictable across the codebase — one shared client with one well-understood timeout, rather than each call site making its own (often copy-pasted) decision.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -355,6 +355,10 @@ ctx, cancel := context.WithTimeout(cmd.Context(), 5*time.Second)
 defer cancel()
 ```
 
+If a request needs a longer timeout, clone `api.HTTPClient` and override its
+`Timeout` rather than building a new `http.Client` from scratch — this keeps
+the User-Agent (see `upgradeHTTPClient` in `internal/cmd/upgrade.go`).
+
 ### Configuration System
 
 The CLI uses a layered configuration approach:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -340,6 +340,21 @@ Tiger CLI is a Go-based command-line interface for managing Tiger, the modern da
   - Log fetching with pagination (FetchServiceLogs)
 - **Utilities**: `internal/util/` - Small utility functions with minimal dependencies (formatting, validation, password generation)
 
+### HTTP Requests
+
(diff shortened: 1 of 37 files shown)


Askir (approved):
🚀

PR #199, opened 2026-08-24
Add shell completion for flags with predefined values

nathanjcochran:
## Summary
- Adds tab-completion for CLI flags that only accept a fixed set of values (`--output`, `--environment`, `--addons`, `--cpu`/`--memory`, `--password-storage`, and the experimental `service metrics series` `--fn`/`--role` flags).
- `tiger config set <key> <value>` now completes the value based on the key already typed.
- Wherever a flag's valid values were already exposed by an existing function or generated enum (e.g. `common.ValidAddons()`, the CPU/memory allow-list, the generated metrics `Fn` enum), the completion reuses that same source rather than a separate hardcoded list, so it can't drift from validation.


murrayju (approved):
👍 lgtm, didn't test

PR #202, opened 2026-08-25
Refactor command tests into table-driven tests with unified harness

nathanjcochran:
Rewrites the CLI command tests around a single table-driven harness, adopting the patterns used in ghost.

Previously each command group had its own ad-hoc setup/execute helpers (often largely duplicated but slightly different), tests asserted with substring matches, most of them stubbed internal seams rather than exercising the command end to end, and the suite needed a real system keyring (including gnome-keyring setup in CI). Coverage was also uneven — several commands (`service start`, `service stop`, `service logs`, `mcp start`, `version`, `project use`, shell completion) had no or incomplete tests. Completion was actively dangerous to run: `__complete service get` made a live API call against the developer's account.

Now every command has one table-driven test built on a shared harness in `main_test.go`: it runs the real root command with a mocked API client (via a new `ClientFactory` hook on the App), an isolated temp config directory, a fresh in-memory keyring, and a scrubbed environment, and asserts stdout/stderr/errors with exact matches. Test cases are ordered to follow each command's execution flow, so a table reads like the command it tests. For output that is inherently nondeterministic (random OAuth state, OS-dependent error text, huge generated schemas), a case can swap the exact-match string for a matcher (`matchRegexp`/`matchPrefix`/`matchFunc`) — with a comment saying why — so even those commands fit the standard tables instead of needing one-off test functions. The result each case can assert on includes the config the invocation actually resolved, so flag/env/file precedence is tested on what the command saw rather than re-derived. Cases that exercise `--wait` polling run under `testing/synctest`, so they assert realistic default timeouts (30m/10m) while finishing instantly — the whole `internal/cmd` suite runs in ~1.3s. Tests never touch the real keyring, config, env, or network, which lets us drop the keyring setup from the test workflow. The integration tests are unchanged.

Changes to production code that came out of this work:

- `buildRootCmd` also returns the `App` so tests can inject the mock client, and the upgrade command's current-binary lookup became an overridable var so upgrade tests can run the full download/verify/replace flow against fake release servers.
- `tiger config show` no longer reports `config_dir` — it isn't a real config value (it can't be set with `tiger config set`, only via `--config-dir`/`TIGER_CONFIG_DIR`), and it was the one machine-varying value in the output. All three output formats now list properties in the same sorted order (the table and JSON were partially hand-ordered before). Together these make the output deterministic, so the tests assert plain literal expected strings, ghost-style.
- The remaining test-only seams are gone from production: `SetTestServiceName`/`GetServiceName` (tests reset a mock keyring per test instead of namespacing entries, so `internal/config` no longer imports `testing`), `UseTestConfig` (deleted), and `StoreCredentialsToFile` (unexported). A reflection-based drift guard now asserts the five places a config key must be registered all agree.
- `config.ValidConfigOptions()` returns sorted keys, so `tiger config set <TAB>` no longer reshuffles its candidates on every keystroke.
- Small fixes found while reviewing the code under test: `service start`/`service stop` now pass the command context to their API calls (so Ctrl-C cancels the request), a duplicate unreachable nil-response check in `service list` is gone, and two unused test seams (`getServiceDetailsFunc`, `validateAPIKey`) were inlined.

Coverage from unit tests alone (`go test -coverpkg=./... ./...`, integration tests skipped), main vs. this branch:

| Package | main | branch |
|---|---|---|
| **total** | **42.1%** | **50.1%** |
| internal/cmd | 57.9% | 76.9% |
| internal/common | 61.6% | 65.3% |
| internal/mcp | 60.7% | 66.5% |
| internal/util | 43.4% | 63.6% |
| internal/config | 85.4% | 88.4% |
| internal/version | 82.1% | 83.9% |
| internal/analytics | 85.5% | 72.7% |
| internal/api (generated) | 15.3% | 10.6% |

The `internal/api` and `internal/analytics` drops are expected: the mocked client means unit tests no longer execute the generated HTTP plumbing or send real analytics identify/track calls the way the old seam-stubbed tests incidentally did (the integration tests still cover those paths).


Askir (approved):
I only reviewed the pattern itself but this seems like a great change. So imo feel free to merge. As long as the suite is still green I think this is an improvement.

PR #203, opened 2026-08-26
Add tiger project list command

aprimakina:
`tiger project list` shows the projects the caller has access to, marking the active one. It follows tiger service list: NoArgs, and an `ls` alias.

  The /projects response carries only an id and a name per project, so a bare list gives no way to tell which one subsequent commands operate on. `OutputProject` wraps `api.Project` with a locally-derived. Current field, set by comparing each id against the project ID the App already resolved when it built the API client — no extra lookup, so the marked row is by construction the project every other command in the same invocation targets. It renders as a * in the CURRENT column and as "current": true in JSON and YAML, so all three formats agree.

  No MCP counterpart here: tiger project has no MCP surface today, and project use is deliberately excluded from it. A read-only `project_list` tool would be safe to add later, since it switches nothing.

### DIFF
--- a/internal/cmd/project_list.go
+++ b/internal/cmd/project_list.go
@@ -1,0 +1,115 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+
+	"github.com/timescale/tiger-cli/internal/api"
+	"github.com/timescale/tiger-cli/internal/common"
+	"github.com/timescale/tiger-cli/internal/util"
+)
+
+// OutputProject is a project plus the locally-derived fields the API doesn't
+// return.
+type OutputProject struct {
+	api.Project
+
+	// Current reports whether this is the active project, i.e. the one
+	// subsequent commands operate on.
+	Current bool `json:"current"`
+}
+
+// buildProjectListCmd represents the list command under project
+func buildProjectListCmd(app *common.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List all projects",
+		Long: `List the Tiger Cloud projects you have access to.
+
+The active project — the one subsequent commands operate on — is marked in the
+output. Use 'tiger project use' to switch to another one.`,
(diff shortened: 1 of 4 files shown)


internal/cmd/project_list.go
```diff
@@ -0,0 +1,115 @@
...
+			Current: project.ID == currentProjectID,
+		}
+	}
+	return prepared
+}
+
+// outputProjectsTable outputs projects in a formatted table using tablewriter
+func outputProjectsTable(projects []OutputProject, output io.Writer) error {
+	table := tablewriter.NewWriter(output)
+	table.Header("CURRENT", "PROJECT ID", "NAME")
```
nathanjcochran: Having `CURRENT` as a column of its own, when the most that's ever in it is a single `*`, feels like potentially a lot of wasted/blank space in the output:

```
$ tiger project list
┌─────────┬────────────┬──────────────────────────────┐
│ CURRENT │ PROJECT ID │             NAME             │
├─────────┼────────────┼──────────────────────────────┤
│ *       │ lsof1xqjus │ Nate's Timescale Project     │
│         │ mzel21cd40 │ Tiger (Company wide project) │
└─────────┴────────────┴──────────────────────────────┘
```

Do we need an entire column for that? Or could we just put the `*` in the `NAME` column after the name itself, or something along those lines that? That's how Ghost does it, for example:

```
$ ghost space list
ID          NAME              ROLE
enm7pjcwjx  Ghost             owner
j3tj8c6p4m  Ghost             owner
opvsxq6dsw  Ghost             owner
r48ihzsy3m  Ghost             owner
avjkacsefa  Nathan's Space *  owner
```

Though the meaning of the `*` is admittedly a little less clear there... so perhaps having a separate column is better (I don't actually have strong feelings - just posing the question). If we do keep it as a separate column (which I'm starting to come around to, now that I'm writing this), I wonder if the column should be moved to the end, rather than being the first column? I kind of feel like the Project ID and Name columns should come first, and the `*` should come after? What do you think?  🤔

nathanjcochran (approved):
Left one question I'd be curious to hear your thoughts on, but otherwise LGTM!

internal/cmd/project_list.go
aprimakina: Agree, will move it to the end. I also like how ghost exposes the role 👍

PR #205, opened 2026-08-31
Rewrite CLAUDE.md

nathanjcochran:
The CLAUDE.md file had grown long, disorganized, and verbose, which bloats every agent session's context and makes the actual conventions harder to find. This rewrites it from scratch: reorganized to read top-down (overview, architecture, and repo structure first; specific patterns and practices below), with each convention stated once in a short, focused section.

Code examples, historical comparisons, and lengthy justifications are gone — the doc now just states the current patterns and lets the code itself serve as the example, so there's less to drift out of date. The content was cross-checked against the codebase (and Ghost's CLAUDE.md, since the two projects share many patterns), picking up several conventions that were previously undocumented — error handling and exit codes, shell completion requirements, CLI/MCP pairing expectations, release steps, and more. Net result: ~1380 lines down to ~250 without losing any of the actual rules.

Also removes some unnecessary `yaml:` struct tags to match the documented serialization rule (YAML output round-trips through JSON, so only `json:` tags are needed).

Stacked on #202.


areina (approved):
Thanks for taking care of improving this 👍 
By the way, what has worked pretty well for me as well it is moving specific content to its own document (e.g. `docs/release.md`) and just reference it in the CLAUDE.md file. No need to do it now, in case the file grows again and we need to organize it in the future.

PR #206, opened 2026-08-31
Add `tiger db query` command and share query logic with the `db_query` MCP tool

nathanjcochran:
Adds `tiger db query` (aliased as `tiger db sql`), which runs a SQL query against a service and prints the results, without needing psql installed. The query can come from `--command`, from a `.sql` file via `--file`, or from stdin, and results render as a psql-style table by default or as JSON/YAML with `--output`. Multi-statement queries work, and the usual `db` flags (`--role`, `--pooled`, `--read-only`, `--timeout`) behave as they do elsewhere.

Tiger MCP already had a `db_execute_query` tool doing essentially the same job, so the query logic now lives in a single shared implementation in `internal/common`, the way Ghost does it. The MCP tool keeps the things that are specific to serving an agent — the `mcp_max_rows` and response-size caps, and the truncation notice — while the CLI returns everything the query produced. The tool is also renamed to `db_query` to match the command, which follows our convention for CLI/MCP pairs, and it picks up Ghost's `file` input so an agent can point it at a `.sql` file instead of inlining a large script. Renaming a tool is technically breaking, but it isn't a concern here: the tool name is only ever seen by LLM agents, which discover the available tools at runtime and will simply start calling the new name once the user updates Tiger CLI (it isn't a surface anyone programs against, unlike the CLI commands, which can be embedded in scripts and therefore do need to stay backwards-compatible).

The most consequential change to the MCP tool is how row values are represented. We were letting pgx decode each value into a Go type and then marshaling that to JSON, which quietly mangled a lot of types: uuids came back as 16-element byte arrays, intervals and bit strings as structs of pgx internals, macaddr and bytea as base64, bigints past 2^53 lost precision, and a `float8 'Infinity'` failed the whole response with a marshaling error. We now read every value as the text Postgres itself produced, exactly as psql would show it, with SQL NULL preserved as JSON `null`. As a database company we shouldn't be reshaping what the database returns, and this also keeps the code thin — it leans on pgx rather than adding per-type handling of our own.

Two smaller improvements came along the way. Query and schema commands now check that a service is actually running before connecting, so a paused service produces a clear message telling you to start it rather than an opaque connection error, and a rejected password explains how to save or reset one instead of surfacing a bare pgx authentication failure. And because this touched the connection plumbing, it seemed like a good moment to unwind some accumulated indirection there: a handful of small single-caller wrappers around building connection details, opening pgx connections, and resolving a service ID to a connection target have been collapsed or inlined, so the call sites now show what they actually do instead of hopping through several layers to find out.

The integration coverage for querying lives in the service-lifecycle test rather than the database-commands one, since that's the test that provisions its own service — and therefore the only one of the two that runs in CI today.


murrayju (approved):
I didn't test, but I definitely agree with the direction, and the code changes look reasonable to me.

PR #208, opened 2026-09-01
Add service backup region commands and MCP tools

areina:
- Lists, adds and removes the regions a service's backups are copied to (`service backup region list/add/remove`), preview only, gated behind TIGER_EXPERIMENTAL like the existing backup command.
- Nests the existing `service backup` command under a group (`service backup list`) instead of a bare `service backup`, so `region` reads as part of the same resource rather than a same-prefixed sibling command.
- MCP tools mirror the CLI: `service_backup_list` (renamed from `service_backups`), `service_backup_region_list/add/remove`.

Example:

```
> TIGER_EXPERIMENTAL=1 tiger service backup region add <service-id> --region eu-central-1
✅ Backups for service '<service-id>' will now be copied to 'eu-central-1'.

tiger service backup region list <service-id>
┌──────────────┬───────────────────────┐
│   REGION     │         ADDED         │
├──────────────┼───────────────────────┤
│ eu-central-1 │ 2026-09-01 18:14 CEST │
└──────────────┴───────────────────────┘

> TIGER_EXPERIMENTAL=1 tiger service backup region remove <service-id> --region eu-central-1
✅ Backups for service '<service-id>' will no longer be copied to 'eu-central-1'.
```


internal/cmd/service_backup_region_add.go
```diff
@@ -0,0 +1,88 @@
...
+
+			if resp.StatusCode() != http.StatusCreated {
+				return common.ExitWithErrorFromStatusCode(resp.StatusCode(), resp.JSON4XX)
+			}
+
+			if resp.JSON201 == nil {
+				return fmt.Errorf("empty response from API")
+			}
+
+			cmd.PrintErrf("✅ Backups for service '%s' will now be copied to '%s'.\n", serviceID, region)
```
nathanjcochran, line 68: We recently removed all emojis from Tiger CLI output (see #231). I see emojis being used in a few places in this PR - would be good to remove them, to match the new convention.

internal/cmd/service_backup_region_add.go
```diff
@@ -0,0 +1,88 @@
...
+				return nil
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&region, "region", "", "Region to copy backups to")
+	cmd.Flags().VarP(new(outputFlag), "output", "o", "Output format (json, yaml, table)")
+	cmd.RegisterFlagCompletionFunc("output", outputCompletion())
+
+	cmd.MarkFlagRequired("region")
```
nathanjcochran, line 85: Use the [`markFlagRequired`](https://github.com/timescale/tiger-cli/blob/5db8ff4c9f1ef731c2213e8453cb5f13cb17ef48/internal/cmd/flag_helper.go#L14) helper function instead (I think it might have been added after you originally created this PR). It panics on error instead of silently throwing away the error.

I think there are other places in this PR that need this same change.

internal/cmd/service_backup_region_list.go
```diff
@@ -0,0 +1,109 @@
...
+				return fmt.Errorf("empty response from API")
+			}
+			regions := *resp.JSON200
+
+			if len(regions) == 0 {
+				cmd.PrintErrln("No backup regions configured for this service.")
+				return nil
+			}
+
+			return outputBackupRegions(cmd, regions, cfg.Output)
```
nathanjcochran, line 71: If `len(region) == 0` but the user provided `-o json` or `-o yaml`, I think we'd want to output an empty list here, not the "No backup regions configured for this service." message. That way the output can be reliably parsed by `jq` and other tools like that, which is important for it to be used in scripts, etc. So I'd move the `len(regions) == 0` check into `outputBackupRegionsTable`.

internal/cmd/service_backup_region_remove.go
```diff
@@ -0,0 +1,87 @@
...
+
+			if err := common.CheckReadOnlyByServiceID(cmd.Context(), cfg, client, projectID, serviceID); err != nil {
+				return err
+			}
+
+			if !removeConfirm {
+				if !util.IsTerminal(cmd.InOrStdin()) || !util.IsTerminal(cmd.ErrOrStderr()) {
+					return fmt.Errorf("TTY not detected - cannot prompt for confirmation. Use --confirm to skip the prompt")
+				}
+				cmd.PrintErrf("Are you sure you want to stop copying service '%s' backups to '%s'? Existing copies there will be deleted and cannot be recovered.\n", serviceID, region)
```
nathanjcochran, line 56: > Existing copies there will be deleted and cannot be recovered.

So if you remove a backup region, and then re-add it, what happens? I would assume that the existing backups get re-copied from the service's region to the backup region, which more-or-less "recovers" them. The `tiger service backup add` description says "The region is added immediately; existing backups are copied to it in the background." If that's true, I'm not sure we need such strongly worded language here...? Thoughts?

internal/mcp/service_backup_region_add.go
```diff
@@ -0,0 +1,98 @@
...
+	return util.Must(jsonschema.For[ServiceBackupRegionAddOutput](nil))
+}
+
+func newServiceBackupRegionAddTool() *mcp.Tool {
+	return &mcp.Tool{
+		Name:  toolServiceBackupRegionAdd,
+		Title: "Add Service Backup Region",
+		Description: `Start copying a service's backups to another region.
+
+The region is added immediately; existing backups are copied to it in the background.`,
```
nathanjcochran, line 49: Does this incur additional costs for the user? Assuming it does, it might be good to mention that here. Some of our other tools (e.g. `service_create`) have warnings like: `WARNING: Creates billable resources.` (see [here](https://github.com/timescale/tiger-cli/blob/5db8ff4c9f1ef731c2213e8453cb5f13cb17ef48/internal/mcp/service_create.go#L91)).

internal/mcp/service_backup_region_remove.go
```diff
@@ -0,0 +1,89 @@
...
+	s.logger.Info("MCP: Removing service backup region",
+		slog.String("project_id", projectID),
+		slog.String("service_id", input.ServiceID),
+		slog.String("region_code", input.RegionCode),
+	)
+
+	resp, err := client.DeleteBackupRegionWithResponse(ctx, projectID, input.ServiceID, input.RegionCode)
+	if err != nil {
+		return nil, ServiceBackupRegionRemoveOutput{}, fmt.Errorf("failed to remove backup region: %w", err)
+	}
```
nathanjcochran, line 80: In #234, I added a new pattern for asking for human confirmation via [MCP Elicitation](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation) before an agent performs a destructive action on a PROD service via an MCP tool. If we think removing a backup region is risky enough to require a service ID confirmation when done via the `tiger service backup remove` CLI command, I think we should probably leverage the elicitation confirmation pattern for PROD services here, too. Let me know if you have any questions about it, but I think it should be straightforward to just copy the pattern currently being used in the `service_delete` MCP tool.

internal/cmd/service_backup_list.go
```diff
@@ -16,12 +16,12 @@ import (
...
-func buildServiceBackupsCmd(app *common.App) *cobra.Command {
+// buildServiceBackupListCmd creates the backup-list command for listing a
+// service's backups. The endpoint is marked preview upstream, so registration
+// is gated on TIGER_EXPERIMENTAL in buildServiceCmd.
+func buildServiceBackupListCmd(app *common.App) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "backup [service-id]",
+		Use:   "list [service-id]",
 		Short: "List backups for a service",
 		Long: `List the full and incremental backups taken for a database service.
```
nathanjcochran, line 26: I think this command should have an `ls` alias too (like all of the other `list` commands do).

internal/cmd/service_backup_region_remove.go
```diff
@@ -0,0 +1,87 @@
...
+
+// buildServiceBackupRegionRemoveCmd creates the region remove subcommand.
+func buildServiceBackupRegionRemoveCmd(app *common.App) *cobra.Command {
+	var region string
+	var removeConfirm bool
+
+	cmd := &cobra.Command{
+		Use:   "remove [service-id]",
+		Short: "Stop copying a service's backups to a region",
+		Long: `Stop copying a service's backups to a region.
```
nathanjcochran, line 21: I think this command should probably have a `rm` alias (like the other removal/deletion commands do).

internal/cmd/service_backup_region_list.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/timescale/tiger-cli/internal/common"
+	"github.com/timescale/tiger-cli/internal/util"
+)
+
+// buildServiceBackupRegionListCmd creates the region list subcommand.
+func buildServiceBackupRegionListCmd(app *common.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list [service-id]",
+		Short: "List a service's backup regions",
+		Long: `List the additional regions a service's backups are copied to.
```
nathanjcochran, line 22: I think this command should have an `ls` alias as well.

nathanjcochran (commented):
Sorry for taking so long to review this (it fell off my radar for awhile and I just remembered about it today 😬).

Overall, the changes LGTM. I left a handful of comments about things that should be updated though - mostly related to patterns that have changed since this PR was originally created. It also appears to fallen significantly behind `main`, and it looks like there are some merge conflicts to resolve. Once those things are handled, I think this will be fine to merge. Let me know if you have any questions!

PR #213, opened 2026-09-02
Device code login auth flow

aprimakina:
`tiger auth login` only works where the CLI and the browser sit on the same machine: the redirect goes to `http://localhost:<port>/callback`, so on an SSH session, in a container, or on any host without a browser there is nothing to redirect back to — and pasting the auth URL into a browser elsewhere doesn't help, since that URL redirects to a port on the machine running the CLI.

This adds the RFC 8628 device authorization grant as a second way in: the CLI prints a short code, the user enters it in a browser on whatever machine they do have one, and the CLI polls until it's authorized. `--headless` goes straight there; otherwise the redirect flow runs as before and the device code stands in for exactly one condition — the browser couldn't be opened at all.

Needs the server halves: timescale/savannah-gateway#1948 and timescale/savannah-users#573.

### DIFF
--- a/internal/cmd/auth_login.go
+++ b/internal/cmd/auth_login.go
@@ -497,12 +497,12 @@ func (l *oauthLogin) getTokenViaBrowser(ctx context.Context) (*oauth2.Token, err
 	l.cmd.PrintErrf("Auth URL is: %s\n", authURL)
 	l.cmd.PrintErrln("Opening browser for authentication...")
 
-	if err := openBrowser(authURL); err != nil {
-		l.cmd.PrintErrf("Failed to open browser: %s\n", err)
-		return nil, errBrowserOpenFailed
-	}
+	browserErr := openBrowserAsync(authURL)
 
 	select {
+	case err := <-browserErr:
+		l.cmd.PrintErrf("Failed to open browser: %s\n", err)
+		return nil, errBrowserOpenFailed
 	case result := <-server.resultChan:
 		return result.token, result.err
 	case <-time.After(browserAuthTimeout):
@@ -723,6 +723,11 @@ func (c *oauthCallback) sendError(err error) {
 	c.resultChan <- oauthResult{err: err}
 }
 
+// openBrowserImpl waits for the launcher to exit, which is how a browser that
+// can't be opened at all -- xdg-open with no display, say -- reports itself
+// rather than looking like a success. It usually exits as soon as the page is
+// handed off, but some Linux configurations keep it running for the life of the
+// browser, so call it through openBrowserAsync rather than directly.
 func openBrowserImpl(url string) error {
 	var cmd *exec.Cmd
 
@@ -736,7 +741,24 @@ func openBrowserImpl(url string) error {
 		cmd = exec.Command("xdg-open", url)
 	}
 
-	return cmd.Start()
+	return cmd.Run()
+}
+
+// openBrowserAsync opens the URL in a goroutine and returns a channel that
+// receives an error if the launcher fails. It exists so the login can wait for
+// the callback and for a failure to open at the same time, without blocking on
+// a launcher that runs for the life of the browser.
+func openBrowserAsync(url string) <-chan error {
+	errCh := make(chan error, 1)
+	// Read the var before the goroutine starts, so a test restoring it in
+	// t.Cleanup doesn't race this call.
+	open := openBrowser
+	go func() {
+		if err := open(url); err != nil {
+			errCh <- err
+		}
+	}()
+	return errCh
 }
 
 func (l *oauthLogin) selectProjectID(ctx context.Context, client *api.ClientWithResponses) (string, error) {
--- a/internal/cmd/auth_login.go
+++ b/internal/cmd/auth_login.go
@@ -47,6 +47,19 @@ func nextSteps(readOnlySet bool) string {
 	return nextStepsMessage + readOnlyNextStep
 }
 
+// browserAuthTimeout is how long the redirect flow waits for the callback.
+const browserAuthTimeout = 5 * time.Minute
+
+var (
+	// defaultDeviceCodeTTL bounds polling when the gateway omits expires_in.
+	// Overridden in tests.
+	defaultDeviceCodeTTL = 15 * time.Minute
+)
+
+// errBrowserOpenFailed means the redirect flow never started, which is the one
+// condition the device code stands in for.
+var errBrowserOpenFailed = errors.New("failed to open browser")
+
 var (
 	// openBrowser can be overridden for testing
 	openBrowser = openBrowserImpl
@@ -63,6 +76,7 @@ type credentials struct {
 func buildLoginCmd(app *common.App) *cobra.Command {
 	var flags credentials
 	var projectID string
+	var headless bool
 
 	cmd := &cobra.Command{
 		Use:   "login",
@@ -74,6 +88,9 @@ The OAuth flow will:
 - Open your browser for authentication
 - Let you select a project (if you have multiple)
 - Store an OAuth session for the selected project
+
+If the browser cannot be opened, the command prints a short code to enter in a browser on
+any other machine instead. Use --headless to go straight to that flow.
 
 Use --project-id to pick the project up front and skip the interactive selection. After
 logging in, you can switch projects with 'tiger project use'.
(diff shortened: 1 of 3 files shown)


internal/cmd/auth_login.go
Askir, line 741: This apparently has an edge case where xdg-open is installed but no browser or display is available that the device flow would not kick in. Idk how often that case would happen but using cmd.Run() below instead of Start() apparently would fix it as it waits for the exit status which would tell you that e.g. no display is available and the device code would kick in.

Askir (approved):
LGTM, claude found one tiny issue which im not sure is worth fixing.
Other than that lets get someone to test it with how marketing wants to use it soon, in case anything else is broken for them.

internal/cmd/auth_login.go
```diff
@@ -439,28 +491,141 @@ func (l *oauthLogin) getOAuthToken(ctx context.Context) (*oauth2.Token, error) {
 		return nil, fmt.Errorf("failed to create local server: %w", err)
 	}
 	defer func() {
-		if err := server.server.Shutdown(ctx); err != nil {
+		// Shutdown gives up on a canceled context while a connection is still
+		// active -- the callback that just delivered -- so it gets a live one.
+		if err := server.server.Shutdown(context.WithoutCancel(ctx)); err != nil {
```
nathanjcochran: I'm not fully understanding the reason for the change, or the comment that trying to explain it 🤔. If the user hits `ctrl-c` and cancels the context, don't we always want to shut down immediately? And if so, why are we no longer propagating the cancellation to the context that gets passed to `Server.Shutdown()`? Doesn't that just mean that `ctrl-c` could theoretically hang forever now (if a request remained open to the local HTTP server)?

internal/cmd/auth_login.go
```diff
@@ -439,28 +491,141 @@ func (l *oauthLogin) getOAuthToken(ctx context.Context) (*oauth2.Token, error) {
...
 			l.cmd.PrintErrf("Failed to close local server: %s\n", err)
 		}
 	}()
 
 	authURL := server.oauthCfg.AuthCodeURL(state, oauth2.S256ChallengeOption(codeVerifier))
 	l.cmd.PrintErrf("Auth URL is: %s\n", authURL)
 	l.cmd.PrintErrln("Opening browser for authentication...")
+
+	// Navigating to the URL by hand wouldn't help either: it redirects to a
+	// port on this machine.
```
nathanjcochran: I don't think think we need this comment. It feels like one of those comments LLMs add to justify a change from the old behavior, but it won't make as much sense in the future, when the old behavior (namely, printing out the URL and telling the user to manually navigate to it) is no longer present 😅.

internal/cmd/auth_login.go
```diff
@@ -439,28 +491,141 @@ func (l *oauthLogin) getOAuthToken(ctx context.Context) (*oauth2.Token, error) {
...
+	da, err := oauthCfg.DeviceAuth(deviceFlowContext(ctx))
+	if err != nil {
+		return oauth2.Config{}, nil, fmt.Errorf("failed to start device authorization: %w", err)
+	}
+
+	// The gateway floors interval but not expires_in, and x/oauth2 bounds
+	// polling only when it sent one -- so supply the bound ourselves.
+	if da.Expiry.IsZero() {
+		da.Expiry = time.Now().Add(defaultDeviceCodeTTL)
+	}
```
nathanjcochran: I'm not totally following why this is necessary. The comment isn't really helping (what does "floors" mean in this context?).

internal/cmd/auth_login.go
```diff
@@ -439,28 +491,141 @@ func (l *oauthLogin) getOAuthToken(ctx context.Context) (*oauth2.Token, error) {
...
+// pollDeviceToken polls until the authorization is decided: an OAuth error is
+// that verdict, since x/oauth2 polls through authorization_pending and
+// slow_down, and a 5xx ends it as a failure to answer. Anything else is
+// retried until the codes expire, a deadline DeviceAccessToken takes from
+// da.Expiry.
+func (l *oauthLogin) pollDeviceToken(ctx context.Context, oauthCfg oauth2.Config, da *oauth2.DeviceAuthResponse) (*oauth2.Token, error) {
+	ctx = deviceFlowContext(ctx)
+
+	for {
+		token, err := oauthCfg.DeviceAccessToken(ctx, da)
```
nathanjcochran: I'm kind of confused as to why we're calling [oauth2.Config.DeviceAccessToken()](https://pkg.go.dev/golang.org/x/oauth2#Config.DeviceAccessToken) in a loop. That method already handles polling internally. From the docs:

> DeviceAccessToken polls the server to exchange a device code for a token. 

So why are we wrapping it in our _own_ polling loop? Based on the comment above, it seems like maybe it's because we're trying to be resilient to other types of errors, which `oauth2.Config.DeviceAccessToken()` just quits if it encounters? But I'm having a kind of hard time understanding what those error conditions would be, and if it's really worth the extra complexity here to handle them? Wouldn't it be a safer and more maintainable long-term to just rely on the standard-compliant implementation in `oauth2.Config.DeviceAccessToken()`, rather than wrapping it in our own polling loop with its own interpretation of the various types of errors?

For comparison, [Ghost's logic](https://github.com/timescale/ghost/blob/2994dbed210f1165d818dc64a66661d080d05612/internal/common/login.go#L315-L319) is much simpler in this regard. It just trusts `oauth2.Config.DeviceAccessToken()` and doesn't have any of this `for` loop or error-type checking logic.

internal/cmd/auth_login.go
```diff
@@ -480,6 +645,16 @@ type oauthResult struct {
...
 
+// sendResult delivers the first result and drops the rest: the login is
+// decided by the first, and a handler blocked on a full channel would stay
+// open, which is enough to keep Shutdown from ever returning.
+func sendResult(ch chan<- oauthResult, result oauthResult) {
+	select {
+	case ch <- result:
+	default:
+	}
+}
```
nathanjcochran: Is this only necessary because we're no longer propagating the context cancellation through to `http.Server.Shutdown` (see my other comment above about that)? Not sure why this needed to change. Is it addressing a real problem you encountered in the existing `tiger auth login` flow?

internal/cmd/auth_login.go
nathanjcochran: Yeah, Ghost uses `cmd.Run()` for this exact reason. Though it also has an async version, which I think I needed to add in case `cmd.Run()` doesn't return immediately (which can apparently happen in some cases, and would cause `tiger auth login` to hang forever after opening the browser). See [here](https://github.com/timescale/ghost/blob/2994dbed210f1165d818dc64a66661d080d05612/internal/common/browser.go#L9-L53).

I'm happy to fix this in a follow-up PR though. Doesn't have to be part of this work imo.

nathanjcochran (approved):
LGTM overall. I did not test it though. Left a handful of comments/questions, often just because it wasn't totally clear to me why a certain change was needed, and the code comments didn't always explain it very clearly imo. The main thing that I think might just be unnecessary complexity is the polling loop surrounding [oauth2.Config.DeviceAccessToken](https://pkg.go.dev/golang.org/x/oauth2#Config.DeviceAccessToken), since that method already handles polling internally. Though I could just be missing something...?

internal/cmd/auth_login.go
aprimakina: Reverted.

internal/cmd/auth_login.go
aprimakina: > we're trying to be resilient to other types of errors, which oauth2.Config.DeviceAccessToken() just quits if it encounters? 

Yes, this extra loop is to handle a dropped connection or a timeout. I agree it is not worth it, removing.

internal/cmd/auth_login.go
aprimakina: Added a fix to the current PR 👍

internal/cmd/auth_login.go
aprimakina: Reverted. The change doesn't belong in a device-code PR. Also, the hang isn't reproducible.

internal/cmd/auth_login.go
aprimakina: Reworded — DeviceAccessToken only sets a deadline when Expiry is non-zero, so without this a gateway that omits `expires_in` would be polled forever.

PR #223, opened 2026-09-08
Change the primary name of several commands

nathanjcochran:
This changes the primary name of four commands to make them clearer or shorter and easier to type. Every old name stays around as an alias, so existing scripts and muscle memory keep working. What changes is only the name that appears in `--help` output and in shell completions.

- `tiger db connect` becomes `tiger db psql`. The command is essentially a pass-through to psql and requires it to be installed, and now that `tiger db query` runs queries without psql, the name should make that dependency obvious.
- `tiger db test-connection` becomes `tiger db ping`, keeping `test` and `test-connection` as aliases. It's shorter to type, and avoiding hyphenated multi-word subcommands makes the CLI feel cleaner.
- `tiger db connection-string` becomes `tiger db uri`. Same reasoning, and it's no less clear: URI is the Postgres term for the kind of connection string we return.
- `tiger config show` becomes `tiger config list`, keeping `show` and `ls` as aliases. Elsewhere `show` is an alias for `get` and returns a single item, so using it for the full list of settings was inconsistent. `list` reserves `show` for single-item commands and matches `git config list`.

Files, functions, and tests are renamed to match the new primary names, and the README and CLAUDE.md are updated accordingly.

These changes were originally proposed in [this Slack thread](https://iobeam.slack.com/archives/C099S47CSRX/p1786657856053839?thread_ts=1786654004.137259&cid=C099S47CSRX).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,7 +77,7 @@ Every command gets its own file in `internal/cmd/`, named to match the command i
 
 Place a helper by who calls it, working down this list until one matches:
 
-1. **One command** → that command's file, even when the helper is large. `db_connect.go` holds the entire `db connect`/`psql` flow (read-replica selection, password recovery, the psql handoff) and `auth_login.go` the entire OAuth flow — a long file whose contents all serve one command is easier to follow than several short files with scattered entry points.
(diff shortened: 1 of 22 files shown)


murrayju (approved):
❤️

Askir (approved):

PR #224, opened 2026-09-09
Generate CLI reference docs

nathanjcochran:
This adds generated Markdown reference documentation for every CLI command, following the same approach ghost already uses (leveraging [Cobra's built-in doc generation functionality](https://cobra.dev/docs/how-to-guides/clis-for-llms/)). A small `cmd/generate-docs` tool walks the Cobra command tree and writes one page per command to `docs/cli/`, so the reference always reflects the real help text rather than something maintained by hand.

Generation is wired into `go generate ./...`, alongside the existing API client and mock generation. That gives contributors a single command to remember, and it means the existing CI check now fails whenever a command, flag, or help text changes without the docs being regenerated. The CI step also picks up newly generated files, so a new command can't land without its reference page.

To keep the output identical across machines, the default config directory is now shown in help as `~/.config/tiger` and expanded when the config loads, instead of being expanded up front into a machine-specific path.


internal/cmd/root.go
```diff
@@ -19,6 +19,12 @@ import (
 	"github.com/timescale/tiger-cli/internal/version"
 )
 
+// BuildRootCmd exposes the root command for tools like doc generators.
+func BuildRootCmd(ctx context.Context) (*cobra.Command, error) {
```
Askir:
```suggestion
// BuildRootCmd exposes the root command for the docs generator.
func BuildRootCmd(ctx context.Context) (*cobra.Command, error) {
```

internal/config/config.go
```diff
@@ -372,12 +373,7 @@ func GetConfigFile(dir string) string {
 }
 
 func GetDefaultConfigDir() string {
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return "./.config/tiger"
-	}
-
-	return filepath.Join(homeDir, ".config", "tiger")
+	return util.ExpandPath(DefaultConfigDir)
```
Askir: Does the wrapper method still deserve to exist when its one line? 🤔

Askir (approved):
neat, would be good to wire it into tiger-data-docs via a github action that auto creates a PR there if anything changes here I think. Although that repo is private and this one is public, not sure if that causes issues with the github tokens.

docs/cli/tiger.md
```diff
@@ -0,0 +1,38 @@
+## tiger
+
+Tiger CLI - Tiger Cloud Platform command-line interface
+
+### Synopsis
+
+Tiger CLI is a command-line interface for managing Tiger Cloud platform resources.
+Built as a single Go binary, it provides comprehensive tools for managing database services,
+VPCs, replicas, and related infrastructure components.
```
aprimakina: We currently don't support VPC and replica management via CLI.

docs/cli/tiger_auth_login.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+      --headless            Authorize by entering a code in a browser on any machine, instead of waiting for a redirect back to this one
+  -h, --help                help for login
+      --project-id string   Project ID to log in to (skips interactive project selection)
+      --public-key string   Public key for authentication
+      --secret-key string   Secret key for authentication
+```
+
+### Options inherited from parent commands
```
aprimakina, line 65: Do you find repeating the section for each command useful?

docs/cli/tiger_auth_login.md
```diff
@@ -0,0 +1,78 @@
+## tiger auth login
+
+Authenticate with Tiger Cloud API
```
aprimakina, line 3: Maybe we should drop the short description? Its content is mostly duplicated in the Synopsis, and for many commands, they are just the same.

aprimakina (approved):

internal/config/config.go
nathanjcochran: Haha good catch. It was only called once within this file anyways 🤦. Removed here: https://github.com/timescale/tiger-cli/pull/224/commits/a5b0cfa3023aba796730e77fe04a6310ba7dea8a

docs/cli/tiger.md
nathanjcochran: Good catch. I don't think this top-level description had ever been updated since it was originally written, and I agree that it wasn't very good. Improved it here: https://github.com/timescale/tiger-cli/pull/224/commits/d13c5dc46602afe2f1e5d50c7d4ec1c65fbb112c

docs/cli/tiger_auth_login.md
nathanjcochran: Not really (I thought the same thing), but it's not easily customizable. This is leveraging cobra's built-in doc generation functionality (see [here](https://cobra.dev/docs/how-to-guides/clis-for-llms/)), and it doesn't expose the ability to omit or modify sections (it only provides options to prepend content to the start of the file, or alter how links are written). I think we could probably post-process the generated markdown docs to remove this section, but that would require introducing our own markdown parsing logic, and I'd rather not go down that road as part of this initial PR, which is just aiming to get the basic functionality in-place. We can improve on it later.

docs/cli/tiger_auth_login.md
nathanjcochran: See my comment above about cobra not exposing a lot of ways to customize how the docs are generated. I think it's something we can investigate in the future if we get to the point of wiring this up to the real docs repo (either by post-processing the generated markdown, or by re-implementing cobra's doc generation code with our own custom implementation), but for now, I think this is a decent starting point.

PR #226, opened 2026-09-10
Add a `tiger feedback` command

aprimakina:
Adds `tiger feedback`, which relays feedback, a bug report, or a support request to the Tiger Data team through the gateway's new `/feedback` endpoint. Modeled on ghost's `feedback` command.

```
tiger feedback "I can't connect to my service after resuming it"
echo "Great tool!" | tiger feedback
tiger feedback   # reads from stdin, prompting on a terminal
```

The submitter's email is resolved from the caller's credentials and the client version and OS from the `User-Agent`, so neither is sent in the body. The command opens no support case and returns no ticket.


internal/cmd/feedback.go
```diff
@@ -0,0 +1,85 @@
...
+		Use:   "feedback [message]",
+		Short: "Submit feedback, a bug report, or a support request",
+		Long: `Submit feedback, a bug report, or a support request to the Tiger Data team.
+
+The message is sent with the email address of your account, so the team can
+follow up, and with the CLI version and operating system. Pass the message as
+an argument, or omit it to read from stdin.
+
+This does not open a support case and returns no ticket to track. For anything
+that needs a tracked response, contact support directly.`,
```
nathanjcochran: Should we include a link to the support page in the console here, or something like that? So users know how to file a formal support ticket?

internal/mcp/feedback.go
```diff
@@ -0,0 +1,110 @@
...
+	return schema
+}
+
+func newFeedbackTool() *mcp.Tool {
+	return &mcp.Tool{
+		Name:  "feedback",
+		Title: "Submit Feedback",
+		Description: `Submit feedback, a bug report, or a support request to the Tiger Data team.
+
+The message reaches a person, so confirm the wording with the user before sending. This opens no support case and returns no ticket to track.`,
```
nathanjcochran: This says that it doesn't open a support case or return a ticket to track, but it doesn't give any indication that users _can_ create real support tickets elsewhere, or how to do it. Should be include something to that effect, like we do in the `tiger feedback` help text (i.e. "For anything needs a tracked response, contact support directly.")? Maybe include a link to the support page in the console?

nathanjcochran (approved):
Left a couple very minor comments, but overall LGTM!

internal/cmd/feedback.go
aprimakina: https://github.com/timescale/tiger-cli/pull/226/changes/788d57e5015e93810698f92c6337712981a7af16

internal/mcp/feedback.go
aprimakina: https://github.com/timescale/tiger-cli/pull/226/changes/788d57e5015e93810698f92c6337712981a7af16

PR #228, opened 2026-09-11
Add environment parameter to service_create and service_fork MCP tools

aprimakina:
## Summary

`tiger service create` and `tiger service fork` both accept `--environment` (DEV/PROD), but the matching MCP tools took no such parameter and hardcoded `api.EnvironmentTagDEV` in their read-only gate. Raised in review of the read-only-modes PR (#195) as likely an oversight rather than a deliberate difference.

Both tools now take an optional `environment` parameter:

- Named `environment` to match both the CLI flag and the `environment` field already in these tools' output.
- Typed `api.EnvironmentTag`, enumerated `DEV`/`PROD` with `DEV` as the schema default, so the SDK rejects anything else before the handler runs — the CLI validates against `validEnvironmentTags` the same way.
- Optional, so omitting it keeps today's behavior exactly: a DEV service.

The read-only gate now checks the requested tag rather than a hardcoded one. Under `read_only=prod`, `service_create` with `environment: PROD` is refused, matching the CLI and the rule that `prod` mode won't create a service it would then refuse to stop or delete. Forking a PROD source into a DEV fork stays allowed, since that reads production without changing it.

The `prod` variant of the server instructions covers that new refusal too — it previously only described PROD *services* being protected, so a refusal with no existing service involved would have looked like a bug.


aprimakina:
Closes https://linear.app/tigerdata/issue/PE-663/add-environment-parameter-to-service-create-and-service-fork-mcp-tools

internal/mcp/service_create.go
```diff
@@ -103,9 +109,15 @@ func (s *Server) handleServiceCreate(ctx context.Context, req *mcp.CallToolReque
...
 	}
 
-	// Deliberately DEV-only: this tool takes no environment_tag, so the API always
-	// tags what it creates DEV.
-	if err := common.CheckReadOnly(cfg, api.EnvironmentTagDEV); err != nil {
+	// Default to DEV when unspecified, matching `tiger service create`.
+	environmentTag := input.Environment
+	if environmentTag == "" {
+		environmentTag = api.EnvironmentTagDEV
+	}
```
nathanjcochran: I don't think you actually need to do this. Because you defined the default as `api.EnvironmentTagDEV` in the JSON schema properties above, the `jsonschema` library should populate the field automatically with the default if the caller omits it. So I believe this code is redundant (this is a common mistake that LLMs often make when building out these MCP tools - I've seen it many times now 😅). Same for the `service_fork` tool.

internal/mcp/service_create.go
```diff
@@ -49,6 +50,11 @@ func (ServiceCreateInput) Schema() *jsonschema.Schema {
 	schema.Properties["replicas"].Default = util.Must(json.Marshal(0))
 	schema.Properties["replicas"].Examples = []any{0, 1, 2}
 
+	schema.Properties["environment"].Description = "Environment tag for the new service. Use 'PROD' only for production workloads — under read-only mode for production services, creating a PROD service is refused."
+	schema.Properties["environment"].Enum = []any{api.EnvironmentTagDEV, api.EnvironmentTagPROD}
+	schema.Properties["environment"].Default = util.Must(json.Marshal(api.EnvironmentTagDEV))
+	schema.Properties["environment"].Examples = []any{api.EnvironmentTagDEV, api.EnvironmentTagPROD}
```
nathanjcochran: I'm not convinced we need the `Examples` field if we also have `Enum`. The enum values already list all of the possible values, so the examples feel pretty redundant (and just take up context, if the MCP client makes them visible to the LLM).

nathanjcochran (approved):
Left two minor comments but LGTM!

nathanjcochran:
Fwiw, I updated the CLAUDE.md in https://github.com/timescale/tiger-cli/pull/229 to clarify both of the points I mentioned above (i.e. that we don't need the Examples field if Enum is set, and that the jsonschema library automatically sets default values), and cleaned up some of the existing MCP tools in the process.

internal/mcp/service_create.go
aprimakina: https://github.com/timescale/tiger-cli/pull/228/changes/106b4157928e42d6929a8014bf264a23622c77d3

internal/mcp/service_create.go
aprimakina: https://github.com/timescale/tiger-cli/pull/228/changes/106b4157928e42d6929a8014bf264a23622c77d3

PR #232, opened 2026-09-15
Add compatibility with Devin

erinmikailstaples:
This pull request adds support for the Devin client to the MCP install command, ensuring that Devin can be installed and configured similarly to other supported clients. The changes update both the implementation and the test suite to recognize and handle Devin appropriately.

**Support for Devin client:**

* Added `Devin` to the `MCPClient` constants and the `supportedClients` list in `mcp_install.go`, including its configuration paths and metadata. [[1]](diffhunk://#diff-34c1efc850a774f8d36ca2b85a75c956b86dae84fb6a0bde87e525e06e783245R103) [[2]](diffhunk://#diff-34c1efc850a774f8d36ca2b85a75c956b86dae84fb6a0bde87e525e06e783245R188-R197)
* Updated the README to document the `tiger mcp install devin` command.

**Test suite updates:**

* Updated the list of supported clients in error messages to include Devin in `mcp_install_test.go`.
* Added test cases to verify installation and recognition of the Devin client, including case-insensitive matching. [[1]](diffhunk://#diff-da419d67cb6afa3a0c20069e3a6ca3041ad5ccc3b00c0fd1a1d4f4a4c980fb63R337-R346) [[2]](diffhunk://#diff-da419d67cb6afa3a0c20069e3a6ca3041ad5ccc3b00c0fd1a1d4f4a4c980fb63R407-R408)


internal/cmd/mcp_install.go
```diff
@@ -184,6 +185,16 @@ var supportedClients = []clientConfig{
...
+	{
+		ClientType:           Devin,
+		Name:                 "Devin",
+		EditorNames:          []string{"devin"},
+		MCPServersPathPrefix: "/mcpServers",
+		ConfigPaths: []string{
+			"~/.devin/mcp_config.json",
+			"~/.codeium/windsurf/mcp_config.json",
+		},
+	},
```
nathanjcochran: Based on [these docs](https://docs.devin.ai/cli/extensibility/mcp/configuration#via-command-line), it looks like the recommended way to install an MCP server into Devin is via the command line. So I think we should add a `buildInstallCommand` section here that calls out to `devin mcp add tiger -- tiger mcp start`, so that we call the Devin CLI to install the MCP server, rather than editing the JSON config file directly. See how we do it for other clients that support adding MCP servers via their CLIs (like Codex below).

Additionally, [these docs](https://docs.devin.ai/cli/extensibility/mcp/configuration#via-config-file) list a few different config file paths than you have here. In particular, it lists the following:
> `~/.config/devin/mcp_config.json` (`%APPDATA%\devin\mcp_config.json` on Windows), `.devin/mcp_config.json`, and `.devin/mcp_config.local.json`

We should make sure all of those files are listed here as well, with the most common default as the first option, so that we correctly find whichever file the user is actually using to make a backup of it (our code checks all config paths [here](https://github.com/timescale/tiger-cli/blob/62bc7eeda64a01cdfc70d02e91ae26f6d8f6d782/internal/cmd/mcp_install.go#L453-L472)).

Also, where did `~/.codeium/windsurf/mcp_config.json` come from? I don't see that listed in the Devin docs 🤔

nathanjcochran: Ahh, I do see `~/.codeium/windsurf/mcp_config.json` listed [here](https://docs.devin.ai/desktop/cascade/mcp#mcp_config-json). But if I'm understanding this warning correctly, that's a legacy setup that we probably don't need to support anymore...?

image

erinmikailstaples: Thank you @nathanjcochran!  Reading the docs its looking like all the windsurf work appears to be no longer maintained, so will just remove that as part of this PR. 

I'll look at how the other MCP servers install via the commandline — should have checked in advance 😓 . 

Incoming soon!

erinmikailstaples:
agh, test failing — converting to draft to return to in a sec.

internal/cmd/mcp_install.go
```diff
@@ -176,12 +176,16 @@ var supportedClients = []clientConfig{
...
+		Name:        "Devin",
+		EditorNames: []string{"devin"},
 		ConfigPaths: []string{
-			"~/.codeium/windsurf/mcp_config.json",
+			"~/.config/devin/mcp_config.json",
+			".devin/mcp_config.json",
+			".devin/mcp_config.local.json",
+		},
+		buildInstallCommand: func(serverName, command string, args []string) ([]string, error) {
+			return append([]string{"devin", "mcp", "add", serverName, "--", command}, args...), nil
```
nathanjcochran: I think we should add `-s user` so it gets installed in user scope, not local scope (the default). That would match what we do for Claude Code (see [here](https://github.com/timescale/tiger-cli/blob/62bc7eeda64a01cdfc70d02e91ae26f6d8f6d782/internal/cmd/mcp_install.go#L166)).

And actually, if we do that, we probably don't need to list the `.devin/mcp_config.json` and `.devin/mcp_config.local.json` files (because I'm realizing now that those are for the project/local scopes, respectively). Sorry for the back-and-forth on that 🙈

nathanjcochran (commented):
Test failure was a flake. I re-ran it and it succeeded 😄

Left one more minor comment, but otherwise this is looking good to me!

nathanjcochran (approved):
LGTM!! Thanks for making those changes!

erinmikailstaples:
> LGTM!! Thanks for making those changes!

TY! not sure why tests are failing in the PR, ran them locally and it's passing 🤔

nathanjcochran:
I pushed a commit to fix the tests: https://github.com/timescale/tiger-cli/pull/232/commits/af74f8a44e8e1e67d271b021e8d6b2e81a6384aa. It was failing due to the `-s user` addition, which wasn't reflected in the tests. Assuming the tests pass in CI now, I think this should be good to merge!

PR #233, opened 2026-09-16
Add 'tiger service rename' command

aprimakina:
Exposes the API's `renameService` operation, which had no CLI surface.

```
tiger service rename svc-12345 analytics-prod
```

Only the display name changes - the service ID, endpoints, and stored credentials are untouched (keyring entries are keyed by project/service/role, pgpass by host), so existing connections and connection strings keep working.

## MCP

Also adds a paired `service_rename` tool, gated by read-only mode like the other write tools. Its output omits the previous name; reporting that would need a fetch before the rename, which neither plane currently makes.

## Follow-up

This is the first piece of broader work on accepting service **names** in place of IDs across the CLI. When that lands, `rename` picks it up by swapping the ref resolution - the `<service-id>` in its usage string becomes `<service-id-or-name>` at that point, not before.


nathanjcochran (approved):
LGTM!

PR #234, opened 2026-09-16
Add `service_delete` MCP tool with elicitation gate for PROD services

nathanjcochran:
Adds a `service_delete` MCP tool, the counterpart to `tiger service delete`, and uses [MCP elicitation](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation) to keep a human in the loop when the target is a production service.

- The tool deletes a service by ID and respects read-only mode the same way the CLI does: `all` blocks every deletion (the tool isn't registered), and `prod` refuses services tagged PROD.
- Deleting a PROD service prompts the user through the MCP client to type the service ID back before it proceeds, matching the CLI's confirmation. A decline, dismissal, or wrong answer cancels cleanly. If the client doesn't support elicitation, the deletion is refused with a pointer to the CLI rather than proceeding unconfirmed.
- The elicitation uses the SDK's multi round-trip pattern (the handler returns an input request and runs again with the answer). Over stdio this works for clients on both the current and older protocol versions. Over the stateless HTTP transport, only clients on the newest protocol version can confirm, since the answer travels inside the request; older ones get the refusal pointing at the CLI, so a PROD service is never deleted unconfirmed either way. The analytics middleware skips the input-request round so each logical call is tracked once.
- Adds a table-driven test harness for MCP tools in `internal/mcp`, mirroring the CLI's `runCmdTests`: each case runs the real server over the mock API client with an in-memory MCP client, so schema validation, middleware, and the elicitation round trip are exercised end to end. I intend to add tests for all other pre-existing MCP tools in a follow-up PR.
- Documents the elicitation pattern and the new test harness in CLAUDE.md, and lists the tool in the README.


internal/mcp/service_delete.go
```diff
@@ -0,0 +1,203 @@
...
+		InputRequests: mcp.InputRequestMap{
+			deleteConfirmationKey: &mcp.ElicitParams{
+				Message: fmt.Sprintf("Delete PRODUCTION service %q (%s)? This permanently destroys the service and all of its data, and cannot be undone.", service.Name, service.ServiceID),
+				RequestedSchema: &jsonschema.Schema{
+					Type: "object",
+					Properties: map[string]*jsonschema.Schema{
+						deleteConfirmationField: {
+							Type:        "string",
+							Title:       "Service ID",
+							Description: fmt.Sprintf("Type the service ID %s to confirm", service.ServiceID),
```
Askir, line 167: okay Its good we give the id here imo :D I think it'd be a bit annoying to search it otherwise!

internal/mcp/service_delete_test.go
```diff
@@ -0,0 +1,197 @@
...
+		"deleted":    true,
+		"message":    `Service "e6ue9697jf" has been deleted.`,
+	}
+	cancelled := map[string]any{
+		"service_id": "e6ue9697jf",
+		"deleted":    false,
+		"message":    `Deletion cancelled: the user did not confirm deleting PROD service "e6ue9697jf" by typing its ID.`,
+	}
+	const noElicitationErr = "deleting service e6ue9697jf requires the user's confirmation because it is tagged PROD, " +
+		"but this MCP client does not support elicitation; run 'tiger service delete e6ue9697jf' from the CLI instead"
```
Askir, line 38: I feel like a lot of users wont know what elicitation actually is unless they built their own MCP server before, maybe its worth changing the error message? But really nitpicky :D

Askir (approved):

internal/mcp/service_delete_test.go
nathanjcochran: Yeah that's fair, but I think I'd rather be clear/accurate about what capability is missing, rather than watering it down and potentially leaving the user confused or without the necessary information to fix the problem (or find a supported MCP client). This message is going back to an LLM anyway, which will probably re-word it, and can likely explain it to the user if they ask.

PR #237, opened 2026-09-17
Add table-driven tests for all MCP tools

nathanjcochran:
Every MCP tool now has a table-driven test in `internal/mcp`, following the pattern set by `service_delete_test.go`: one test file per tool, one inline `runToolTests` table per file ordered by the handler's execution flow, exact matching on errors and structured output, and mock expectations with exact request structs. The tools are exercised through a real in-memory client/server session, so the cases cover the SDK's schema validation and defaults, not just the handler.

- Add tests for all 16 MCP tools.
- Rework the MCP test harness to mirror the `internal/cmd` harness in shape, naming, and file layout: `runTool(t, tool, args, setupMock, opts...)` returns a `toolResult`, and a case configures its run through `runOption`s (`withConfig`, `withConfigAfterStart`, `withNotLoggedIn`, `withExperimental`, `withClientCapabilities`, `withElicitation`, `withKeyringError`) rather than dedicated struct fields. Stubs of package-level vars are `withSetup`-based options, as on the cmd side (`withExecuteQuery`, `withFetchServiceSchema`, `withGenerateServiceName`), and can now return an error so the failure paths are tested too. `wantCallErr` asserts the transport error the SDK reports for an unregistered tool, and `checks` holds post-call assertions.
- Rename the mock-setup field on both `cmdTest` and `toolTest` from `setup` to `setupMock`, matching the `runCommand`/`runTool` parameter it feeds and leaving `setup` to mean the hook slice on `runConfig`.
- Share the database fixtures (`sampleReplica`, `expectGetService`, the readiness messages) in `main_test.go` instead of duplicating them per tool, and collapse the per-file wait-loop poll helpers onto `expectGetService`.
- Reset the mock keyring per case in the MCP harness, matching `internal/cmd`. Without it, a password stored by one tool's test leaked into another's.
- Normalize a JSON `null` array to `[]` in `service_backups` and `service_metrics_available`, as `service_metrics_series` already did, so the three preview tools agree on the same API response.
- Add `toolDBSchema` and `toolFeedback` to the tool-name constants, which the comment there already claimed were the single source of truth.
- Make `common.FetchServiceSchema` a var so tests can stub the database connection, as `common.ExecuteQuery` already is, and give it the same shape: `(ctx, cfg, target, FetchServiceSchemaArgs)`. The args struct absorbs `SchemaOptions` along with the role and pooling flags, so tests assert one struct rather than individual parameters. The conn-level `FetchSchemaFromConn` had no other callers and is now private, and the function moves from `schema_fetch.go` into `schema.go`.
- Make `common.GenerateServiceName` a var so tests can pin the otherwise random name and assert the create request exactly.
- `FetchServiceLogs` now sends its default `until` bound in UTC. Besides being the canonical form for the query parameter, this strips `time.Now()`'s monotonic reading, which was what kept tests from matching the params exactly. Cases that depend on that default run under `synctest`, whose clock starts at a fixed instant; both harnesses document this second use of the field.
- Add success-path tests for the `tiger db schema` CLI command (previously missing), which the new stub makes reachable without a live database.
- Tighten the pre-existing `tiger service create` and `tiger service logs` tests: the generated-name and log-params cases used prefix and field-by-field matchers, and now assert exact requests.
- Remove the helper-level tests now covered by the table-driven MCP tool tests.
- Fix a stale comment in `server.go` about the experimental gate.
- Update CLAUDE.md's testing sections for the harness changes.

Stacked on #234.


Askir (approved):
I did not review in detail, but the spots I checked look good and more coverage in general is always better I would say :D

internal/mcp/service_metrics_available.go
```diff
@@ -70,9 +70,12 @@ func (s *Server) handleServiceMetricsAvailable(ctx context.Context, req *mcp.Cal
 		return nil, ServiceMetricsAvailableOutput{}, common.ExitWithErrorFromStatusCode(resp.StatusCode(), resp.JSON4XX)
 	}
 
-	if resp.JSON200 == nil {
-		return nil, ServiceMetricsAvailableOutput{Series: []string{}}, nil
+	// Default to a non-nil slice so a missing body or a JSON null marshals to
+	// `[]` rather than `null`, keeping the output a valid array.
+	series := []string{}
+	if resp.JSON200 != nil && *resp.JSON200 != nil {
+		series = *resp.JSON200
```
AdrianLC, line 77: Not sure I see how this change was needed. Perhaps this is a case where we're being defensive because of our own test mocks

AdrianLC (approved):

internal/mcp/service_metrics_available.go
nathanjcochran: This was suggested by Claude for the following reasons:

1. If the API ever returns an explicit `null` in the body, `resp.JSON200` (a pointer to a slice) would be non-nil, but `*resp.JSON200` (the slice itself) would be `nil`. As a result, the MCP tool would return a `null` field value instead of `[]`. And I think we'd rather return a consistent `[]` from the MCP tool. So this is basically just being defensive against the API returning bad/inconsistent responses (but I admittedly didn't look into the API code to see whether it's actually possible for it to return an explicit `null` today). 
2. It's already how the `service_metrics_series` MCP tool worked, which you originally wrote (see [here](https://github.com/timescale/tiger-cli/pull/175/changes#diff-c794b5603c32fc9ada2f9eb11c9fa45dd1a90a51ad3779fe41fc3229db055274R1439-R1444)). So this change is just making all of the tools you added consistent in how they handle `[]` vs `null` coming back from the API. I think this is the real reason Claude noticed it - because there was a difference in how these related MCP tools behaved, and it showed up in the tests.

